PR #4, opened 2015-11-10
Don't fail on empty facts[:ansible_lsb]

agx:
This doesn't seem to be set on all Linux distros (e.g. OpenSUSE 13.2)


dLobatog:
Cool, thank you! I'll put this on the first release.

dLobatog:
Merged as db96862e8665cca2de142465aad1103462b70add, thanks @agx!

PR #5, opened 2015-11-10
Move ansible's default interface to front of interface list

agx:
Foreman's FactParser.suggested_primary_interface picks the first physical or
virtual interface. In order to make sure we favour the one preferred by ansible
put it in the front of the list.

If there's no preferred interface at leats sort the list so it's at least
deterministic.

Mabe not that important but it looks nicer if we don't make an arbitrary interface the primary one.


dLobatog:
Merged as 1dae193c3481262b05ee210903e361dd6176af10, thanks @agx!

PR #6, opened 2015-11-10
Provide additional report details

agx:
These are four patches for the ansible callback:

* the first one just moves the docs to the proper location
* the second one makes sure we don't hammer on the report API like crazy for one task at a time
* the third one tries to fill the log entries with some sense
* the 4th one uses ansibles v2 plugin API to use the ansible task name as resource name

This might need more work so just let me know if this is heading into the right direction. With this applied the logs are a bit closer as what one gets with puppet.

### DIFF
--- a/extras/foreman_callback.py
+++ b/extras/foreman_callback.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 import json
 import uuid
 import requests
+import time
 
 try:
     from ansible.plugins.callback import CallbackBase
@@ -43,6 +44,7 @@ class CallbackModule(parent_class):
     """
     def __init__(self):
         self.items = defaultdict(list)
+        self.start_time = int(time.time())
 
     def log(self, host, category, data):
         if type(data) != dict:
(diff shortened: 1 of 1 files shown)


agx:
Hi!
Does this make sense?

dLobatog:
@agx Thanks! I'll try it out in a moment. Sorry it took me so long to get to this, I've been busy with other stuff / travels / etc. Hopefully in an hour or so I'll get back to you with more info (or this merged :smile:  thanks again!)

dLobatog:
Awesome job @agx , merging... 
I'll add some tests for the parser, SSL support in the callback and I'll try to move this to theforeman organization and start releasing packages to make it official.

dLobatog:
Merged as c84c6ed22b6b54113271e8f7c4ee6a6c256bb760, thanks @agx!

agx:
Thanks for merging! I'd be happy to help moving this to theforeman if there are further things needed for that.

PR #14, opened 2016-02-10
Add python setup script

agx:
This allows us to easily build and distribute the ansible hook like:

     python setup.py bdist_rpm

and it is python after all. This will also allow us to use Python's
unittest machinery in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,55 @@
+#!/usr/bin/python
+# vim: set fileencoding=utf-8 :
+# Copyright (C) 2016 Guido Günther <agx@sigxcpu.org>
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 3 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program; if not, please see
+#    <http://www.gnu.org/licenses/>
+# END OF COPYRIGHT #
+
+import re
+from setuptools import setup
+
+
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -0,0 +1,55 @@
...
+                break
+        else:
+          raise Exception("Could not determin version")
+
+    return version
+
+
+setup(name = "foreman_ansible",
+      version = fetch_version(),
+      author = u'Daniel Lobato Garcia',
```
dLobatog: Feel free to add yourself here if you can pass an array

setup.py
```diff
@@ -0,0 +1,55 @@
...
+    version_re = re.compile(".*VERSION\s*=\s*'([0-9\.]+)'\s*$")
+
+    with open("lib/foreman_ansible/version.rb") as f:
+        for line in f:
+            m = version_re.match(line)
+            if m:
+                version = m.group(1)
+                break
+        else:
+          raise Exception("Could not determin version")
```
dLobatog: Typo, 'determin' -> 'determine'

dLobatog:
@agx Fine with me :+1: , however I'm not sure I understand what this is for right now - it doesn't put the callback script anywhere does it?

agx:
Hi,
On Wed, Feb 10, 2016 at 10:36:56AM -0800, Daniel Lobato García wrote:
> @agx Fine with me :+1: , however I'm not sure I understand what this is for right now - it doesn't put the callback script anywhere does it?

It puts it to /usr/bin/ since it's listed in scripts. That way

    python setup.py bdist_rpm

makes it easily available for installation on a separate host than the
foreman server (which might often be the case). It's then just a matter
of:

    ln -s /usr/bin/foreman_callback.py /path/to/ansible/callbacks.py

to activate it. I'm happy to to put it to somewhere else
(e.g. /usr/share/ansible/callbacks/whatnot/....) but since there doesn't
seem to be a default from the ansible site I went for setup.py's default
for the moment.

Another bonus is that we an start using

    python setup.py test
    
and friends to test the script in the future.
Cheers,
 -- Guido

dLobatog:
Merged as 55b4491fc151f09206e6b213273b4a68c46c5140, thanks @agx!

PR #17, opened 2016-02-16
Add a check if 'invocation' key is present in and entry

kireevco:
It's quite common that there are tasks with `no_log: True`. For instance tasks that load sensitive date are mostly with `no_log` enabled. When there is no log available `invocation` key is not present in `res` and `entry` objects.

This PR adds a safe check for this key.


dLobatog:
Merged as 45170b2c6a5cd57b97223c4ee206d493c72d0430, thanks @kireevco!

PR #18, opened 2016-02-19
Fixes #13812 - Remote execution provider

dLobatog:
Ansible can be used as a provider for remote execution. The job
templates can be Ansible playbooks that use host parameters, properties
and ERB.
Job templates, however, should have a 'hosts' section that just contains
<%= @host.name %>, because a new inventory is generated per host with
all the required variables.
In any case, the way it should work is:

Foreman sends request to proxy - including the 'hosts' it's supposed to
run on.
Proxy generates an inventory with the hosts and variables required
Proxy runs ansible and reports to Foreman

The reason why ansible_foreman_inventory cannot be used in this case is
because it's less flexible than the 'search' field of REX, where one can
use the 'scoped_search' syntax to figure out what hosts to run the
playbook on. If we used ansible_foreman_inventory for that, we would be
forced to run our playbooks on a set of hosts, hostgroups, organiztions
or locations.

----

Not to be merged without tests / a plan for installer support / documentation

### DIFF
--- a/app/services/foreman_ansible/ansible_template_renderer.rb
+++ b/app/services/foreman_ansible/ansible_template_renderer.rb
@@ -1,0 +1,86 @@
+if defined? ForemanRemoteExecution
+  module ForemanAnsible
+    class AnsibleTemplateRenderer < InputTemplateRenderer
+      alias_method :playbook, :render
+
+      def initialize(job_template,
+                     host = nil,
+                     invocation = nil,
+                     input_values = nil,
+                     preview = false,
+                     templates_stack = [])
+        super(job_template, host, invocation, input_values, preview,
+              templates_stack)
+      end
+
+      # Takes the variables from host and host group parameters and
+      # sets them in the 'Host variables' and 'Group variables' format
+      # in the inventory. For instance:
+      #
+      # host.example.com
+      #
+      # [hostgroup:vars]
+      # parameter=value
+      #
+      # It does not nest group variables, but instead it creates a new group per
+      # host group full title
+      #
+      def inventory
+        inventory = ''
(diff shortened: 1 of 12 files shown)


app/services/foreman_ansible/ansible_template_renderer.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+          variables += "[#{group.title}:vars]\n"
+          parameters.each do |parameter|
+            variables += "#{parameter.first}=#{parameter.last} \n"
+          end
+          variables += "\n"
+        end
+
+        variables
+      end
```
agx: In foreman_ansible_inventory we currently have all the the params in a foreman_params hash and the hostvars in a foreman hash - this  makes sure we don't collide with any ansible variables like e.g. groups. I wonder if we should do the same to be compatible in the format? This way ansible playbooks could look the same whether we use remote execution or the external inventory script.

We're also prefixing the host group with "foreman_hostgroup_" to have a separate namespace.

agx:
> The reason why ansible_foreman_inventory cannot be used in this case is
because it's less flexible than the 'search' field of REX, where one can
use the 'scoped_search' syntax to figure out what hosts to run the
playbook on. If we used ansible_foreman_inventory for that, we would be
forced to run our playbooks on a set of hosts, hostgroups, organiztions
or locations.

It would also introduce a python dependency on the smart proxy for now good reason.

app/services/foreman_ansible/ansible_template_renderer.rb
dLobatog: That's actually a great idea - thanks! I'll make sure to do that

PR #21, opened 2016-05-04
Allow to exclude hosts in callback plugin

stintel:


agx:
The current patch can only exclude a single host since the env var is a string not a Python list. You need to split at a character (':' like in paths?) to get a list.

Additionally the commit message should have some motivation _why_ this is needed. 

Apart from that I'm o.k. with a exclude list but maybe @dblatog has further suggestions?

dLobatog:
@agx Nothing else, should be ready to merge after @stintel does what you said. Thanks for the contribution!

dLobatog:
@stintel ping?

stintel:
@dLobatog @agx is it acceptable now?

agx:
On Wed, May 18, 2016 at 01:38:16AM -0700, Stijn Tintel wrote:
> @dLobatog @agx is it acceptable now?

Thanks for updating the patch I'm still missing some motivation in the
commit message why we want this? If it's just for excluding localhost I
wonder if we shouldn't rather exclude all hosts that don't have a dot in
it?

PR #23, opened 2016-07-21
Fix missing logger method name

domcleal:
Causes `ArgumentError: wrong number of arguments (1 for 0)` otherwise.


dLobatog:
Merged as af4eb2a9dd14757cc7be7b1e58e0d6955f53dd6f, thanks @domcleal!

PR #24, opened 2016-08-02
Fixes #15948 - Add in the engine initializer to extract strings

bkearney:


lib/foreman_ansible/engine.rb
```diff
@@ -10,6 +13,13 @@ class Engine < ::Rails::Engine
     config.autoload_paths += Dir["#{config.root}/app/services"]
     config.autoload_paths += Dir["#{config.root}/app/views"]
 
+    initializer 'foreman_docker.register_gettext', :after => :load_config_initializers do
```
agx: Should this bee a foreman_ansible instead of foreman_docker ?

dLobatog: Yeah, it doesn't really matter unless you have both foremandocker & foremanansible installed

bkearney:
@agx sorry, bad copy/paste. I pushed a corrected initializer.

dLobatog:
Merged as 537415c14d364c72518141b2727e7e24f50d9e1d, thanks @bkearney!

bkearney:
@dLobatog thanks. Can you push to Transifex, or is that something @domcleal does?

dLobatog:
@bkearney I don't think I have push permissions there, sorry :disappointed_relieved: !

PR #25, opened 2016-08-03
Fixes #15836 - Import roles from smart proxy

xprazak2:


dLobatog:
@xprazak2 Thanks for the PR. I have to add testing to foreman-infra too to make sure this stays stable & fulfills rubocop.

A few comments (haven't tried this yet nor read it fully):
* Please split this up into PRs for API/UI, and other functionality, large changesets are difficult to review effectively
* What's the API necessary for?
* Why is the UI for roles necessary? Does it show the roles that are automatically imported from the Foreman host (w/o proxy)? 
* What version of the proxy supports this? What version of the proxy plugin? Is it packaged? 

thanks!

xprazak2:
* I will break this up into smaller parts
* The API may not be necessary, but shouldn't the functionality of API/UI be consistent? 
* The UI should show all roles regardless of how they are imported. The aim here was to achieve look & feel similar to puppet classes import.
* There are no restrictions on proxy apart from having smart_proxy_ansible plugin. Proxy plugin has not been packaged yet.

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  def destroy
+    if @role.destroy
+      process_success
+    else
+      process_error :object => @role
+    end
+  end
+
+  def import
+    @changed = @importer.import
```
dLobatog: Please render local variables ( :locals => { :changed => @importer.import} ) in templates instead of relying in instance variables, it makes more difficult to change logic in templates

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,44 @@
+class AnsibleRolesController < ::ApplicationController
+  include Foreman::Controller::AutoCompleteSearch
+
+  before_action :find_role, :only => [:destroy]
```
dLobatog: Use `:find_resource`, provided by core, it should able to find it

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  private
+
+  def find_role
+    @role = resource_base.find_by_id(params[:id])
+  end
+
+  def find_proxy
+    @proxy = SmartProxy.find(params[:proxy])
```
dLobatog: `SmartProxy.authorized(:view_smart_proxies).find(params[:proxy])` rather,

app/lib/proxy_api/ansible.rb
```diff
@@ -0,0 +1,14 @@
...
+  class Ansible < ::ProxyAPI::Resource
+    def initialize(args)
+      @url = args[:url] + '/ansible/'
+      super args
+    end
+
+    def roles
+      parse(get 'roles')
+    rescue => e
+      raise ProxyException.new(url, e, N_('Unable to get roles from Ansible'))
```
dLobatog: Can you write a test for this?

app/lib/proxy_api/ansible.rb
```diff
@@ -0,0 +1,14 @@
+module ProxyAPI
+  class Ansible < ::ProxyAPI::Resource
+    def initialize(args)
+      @url = args[:url] + '/ansible/'
+      super args
+    end
+
+    def roles
+      parse(get 'roles')
+    rescue => e
```
dLobatog: Rescue *all* exceptions? Which ones have you seen?

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -1,24 +1,54 @@
...
+    def initialize(proxy = nil)
+      @ansible_proxy = proxy
+    end
+
+    def proxy_api
+      return @proxy_api if @proxy_api
+      find_proxy_api
+    end
+
+    def import
```
dLobatog: nitpick, `import!` to make the dev notice it'll trigger an important change

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -1,24 +1,54 @@
...
-          Rails.logger.debug("Failed to save role #{role_name}: "\
-                            "#{role.errors.full_messages.join(', ')}")
-        end
+    attr_reader :ansible_proxy
+
+    def initialize(proxy = nil)
+      @ansible_proxy = proxy
+    end
+
+    def proxy_api
```
dLobatog: Can't you do this with just `@proxy_api = find_proxy_api`  in the initializer?

app/views/foreman_ansible/hosts/_tab_content.html.erb
```diff
@@ -1,4 +1,4 @@
 <div class='tab-pane' id='ansible_roles'>
-  <%= multiple_checkboxes f, :ansible_roles, @host, ansible_roles,
+  <%= multiple_checkboxes f, :ansible_roles, @host, AnsibleRole.all,
```
dLobatog: If you want to do this, you'll have to add permissions to AnsibleRole then call .authorized here

app/views/ansible_roles/import.html.erb
```diff
@@ -0,0 +1,49 @@
+<% title _("Changed roles") %>
+<%= form_tag send("confirm_import_ansible_roles_path") do %>
```
dLobatog: Why send?

app/views/ansible_roles/import.html.erb
```diff
@@ -0,0 +1,49 @@
...
+                <%= { "new" => _("Add:"), "obsolete" => _("Remove:") }[kind] %>
+              </td>
+            </tr>
+          <% end %>
+        <% end %>
+      <% end %>
+    </tbody>
+  </table>
+  <div>
+    <%= link_to _("Cancel"), send("ansible_roles_path"), :class => "btn btn-default" %>
```
dLobatog: Ditto, why send?

app/views/ansible_roles/import.html.erb
```diff
@@ -0,0 +1,49 @@
...
+                               "toggleCheckboxesBySelector('.role_select_boxes')",
+                               :title => _("Check/Uncheck all")) %>
+        </th>
+        <th><%= _("Name") %></th>
+        <th><%= _("Operation") %></th>
+      </tr>
+    </thead>
+    <tbody>
+      <% ["new", "obsolete"].each do |kind| %>
+        <% unless (roles = @changed[kind]).empty? %>
```
dLobatog: nitpick, `if (roles = @changed[kind]).present?` is more readable imo

dLobatog:
@xprazak2 I've left a few comments, didn't fully finish with the review - but some things are worth looking into already, please take a look. Also I noticed it doesn't pass rubocop (`bundle exec rubocop app`) and a test seems to fail. What version of Foreman is this supposed to work with?

https://github.com/theforeman/foreman-infra/pull/231 is tracking test/rubocop integration in this repo, sorry for not having it automatically yet :disappointed: 

Thanks again!

app/services/foreman_ansible/roles_importer.rb
xprazak2: I could, but import is two step process. In the first step (`import` action), we get the roles from proxy, display the changes to user and let him decide what to import. In the second step (`confirm_import` action), we simply persist the changes user selected. We do not actually need proxy for the second step. We could also move `@proxy_api = find_proxy_api` into initializer and extract the `finish_import` method into a separate class.

xprazak2:
@dLobatog, I made changes according to comments, rubocop and tests are now locally passing for me.

dLobatog:
@xprazak2 Mind to rebase this? I recently merged a PR with fixes to tests, it'd run that automatically now :smile:

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,42 @@
+#:nodoc:
```
dLobatog: Please write some doc, even if they're basic, they allows us to define boundaries for our classes and make newcomers more familiar with this

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+class AnsibleRolesController < ::ApplicationController
```
dLobatog: Nitpick - put this in the `ForemanAnsible` namespace, then just call it RolesController. It'll allow us to not override stuff in ApplicationController like on line 19 (`process_error :object => @role`) would just become `process_error`

app/lib/proxy_api/ansible.rb
```diff
@@ -0,0 +1,15 @@
...
+  # ProxyAPI for Ansible
+  class Ansible < ::ProxyAPI::Resource
+    def initialize(args)
+      @url = args[:url] + '/ansible/'
+      super args
+    end
+
+    def roles
+      parse(get('roles'))
+    rescue Errno::ECONNREFUSED => e
```
dLobatog: A few more... 

`SocketError, Timeout::Error, Errno::EINVAL, Errno::ECONNRESET,
 EOFError, Net::HTTPBadResponse, Net::HTTPHeaderSyntaxError,
 Net::ProtocolError, RestClient::ResourceNotFound`

test/functional/ansible_roles_controller_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_plugin_helper'
+# functional tests for AnsibleRolesController
+class AnsibleRolesControllerTest < ActionController::TestCase
+  setup do
+    @role = FactoryGirl.create(:ansible_role)
+  end
+
+  test 'should get index' do
```
dLobatog: Substitute by `basic_index_test`

test/functional/ansible_roles_controller_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_plugin_helper'
+# functional tests for AnsibleRolesController
+class AnsibleRolesControllerTest < ActionController::TestCase
+  setup do
+    @role = FactoryGirl.create(:ansible_role)
```
dLobatog: No need to create one role in the index test?

test/test_plugin_helper.rb
```diff
@@ -1,9 +1,11 @@
 require 'test_helper'
 
 def sample_facts_file
-  File.read(
-    File.join(
-      ForemanAnsible::Engine.root, 'test', 'fixtures', 'sample_facts.json'))
+  File.read(File.join(
```
dLobatog: Nitpick, 

```ruby
File.read(
    File.join(
      ForemanAnsible::Engine.root,
      'test',
      'fixtures',
      'sample_facts.json'
    )
  )

```

reads a bit better IMO - also it's what's suggested in the newest Rubocop versions (0.42)

lib/foreman_ansible/version.rb
```diff
@@ -2,5 +2,5 @@
 # This way other parts of Foreman can just call ForemanAnsible::VERSION
 # and detect what version the plugin is running.
 module ForemanAnsible
-  VERSION = '1.0'
+  VERSION = '1.0'.freeze
```
dLobatog: :sparkler: :smile:

lib/foreman_ansible/engine.rb
```diff
@@ -30,9 +31,20 @@ class Engine < ::Rails::Engine
           permission :play_roles,
                      { :hosts => [:play_roles, :multiple_play_roles] },
                      :resource_type => 'Host::Managed'
+          permission :view_ansible_roles, :ansible_roles => [:index],
+                                          :hosts => [:new, :edit]
+          permission :destroy_nasible_roles, :ansible_roles => [:destroy]
+          permission :edit_ansible_roles,
+                     :ansible_roles => [:import, :confirm_import]
```
dLobatog: Mind to call this permission `:import_ansible_roles`? It's just for consistency so it matches with `import_puppetclasses` and `import_environments` permission

app/views/ansible_roles/index.html.erb
```diff
@@ -0,0 +1,30 @@
+<% title _("Ansible Roles") %>
```
dLobatog: This is missing a 'Welcome' page following Patternfly's empty state pattern:
https://www.patternfly.org/pattern-library/communication/empty-state/#_

You can pretty much copy the environments page from core, changing the icon to fa-play:
https://github.com/theforeman/foreman/blob/develop/app/views/environments/welcome.html.erb

app/views/ansible_roles/index.html.erb
```diff
@@ -0,0 +1,30 @@
...
+<table class="<%= table_css_classes 'table-fixed' %>">
+  <thead>
+    <tr>
+      <th class="col-md-3"><%= sort :name, :as => s_("Role|Name") %></th>
+      <th class="col-md-3"><%= _("Imported at") %></th>
+      <th class="col-md-1"><%= _("Actions") %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% for role in @roles %>
```
dLobatog: `.each` please, to properly scope the content of the loop (+ performance, let's save the planet :smile:)

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -1,24 +1,55 @@
...
+
+    def import!
+      imported = proxy_api.roles.map do |role_name|
+        ::AnsibleRole.find_or_initialize_by(:name => role_name)
+      end
+      detect_changes imported
+    end
+
+    def finish_import(changes)
+      if changes
```
dLobatog: Use a guard clause, e.g `return unless changes.present?` to avoid making "if" methods.

app/models/ansible_role.rb
```diff
@@ -1,6 +1,10 @@
 # Simple model to store basic info about the Ansible role
 class AnsibleRole < ActiveRecord::Base
+  include Authorizable
+
+  self.include_root_in_json = false
```
dLobatog: Why?

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -1,24 +1,55 @@
 module ForemanAnsible
-  # Service to import roles from the filesystem.
-  # Should be extracted to a gem that both the proxy and Foreman can use.
+  # imports roles from smart proxy
```
dLobatog: Read the previous comment, there are users already using this - the importer should handle both the local fs & the smart proxy. Maybe it's best to split that functionality into two classes

db/migrate/20160729094457_add_columns_to_ansible_role.rb
```diff
@@ -0,0 +1,12 @@
+#:nodoc:
+class AddColumnsToAnsibleRole < ActiveRecord::Migration
+  def up
+    add_column :ansible_roles, :created_at, :datetime
+    add_column :ansible_roles, :updated_at, :datetime
```
dLobatog: Add a default value for this - equal to the Time.now or something - otherwise this breaks https://github.com/theforeman/foreman_ansible/pull/25/files#diff-97a0da96e3dffb0eb149878984f699edR17 when calling .updated_at on pre-existing roles (remember there are roles already in the system prior to this PR)

db/migrate/20160729094457_add_columns_to_ansible_role.rb
```diff
@@ -0,0 +1,12 @@
+#:nodoc:
```
dLobatog: The rubocop doc requirement is there to explain the 'why's of this migration - the 'how' is self-explanatory

app/views/ansible_roles/import.html.erb
```diff
@@ -0,0 +1,49 @@
...
+          <%= link_to_function(icon_text("check"),
+                               "toggleCheckboxesBySelector('.role_select_boxes')",
+                               :title => _("Check/Uncheck all")) %>
+        </th>
+        <th><%= _("Name") %></th>
+        <th><%= _("Operation") %></th>
+      </tr>
+    </thead>
+    <tbody>
+      <% ["new", "obsolete"].each do |kind| %>
```
dLobatog: This should show something when no new or obsolete roles were found, right now it just displays an empty table which is kind of confusing. I think at that point it'd make sense to take the tbody and put it in a partial

dLobatog:
@xprazak2 Thanks for the updates, it looks much more mature & nearing completion. I appreciate the work.

I see just two problems with the current approach:

1.  **It can't handle local roles** (imported from the Foreman host itself). 
    * I realize the proxy does this already but I've had good feedback from people who didn't want to install more stuff in their proxies & on IRC/mailing list about liking that approach. IMO, the way to do this properly is going to be to write a gem that reads the roles, and making both Foreman and the proxy use that gem. That way we don't have to write the same code twice. 
   * For the moment I think it's fine if you just keep the code that's already there and try to adapt the RolesImporter & the UI to handle both cases. 

2.  **It's unclear whether users should or not have proxies with different sets of roles** (not critical..yet)
     * The most user friendly way of doing this is to associate roles with N proxies (or the foreman host itself) , where we know they're available. Then at runtime we trigger the run based off this information, e.g:
     * Proxy A has roles 'redmine' and 'jenkins'
     * Proxy B has roles 'ruby' and 'python'
     * We want to apply 'redmine' and 'ruby' to Host C in Foreman. 
         * Foreman should be able to figure out that it'd trigger the run from Proxy A and Proxy B, and the user doesn't need to know about it. The current approach of associating hosts with Puppet servers makes a bit more sense in Puppet but it's just more work for the user we could save them.

If you can fix the problem with 1 & the comments inline I think it'd be very near to merge if not 100% complete. Thanks again & let me know if you need help on IRC!

app/services/foreman_ansible/roles_importer.rb
ares: Just curious, why would we encourage people to handle local fs? We support it only through proxy for puppet modules. Having two paths that we have to support seems as extra burden to me.

app/models/ansible_role.rb
xprazak2: The importing form posts individual roles as json, this is for the sake of having a simple array of js objects where each object is a role.

xprazak2:
I did not see why we would want to handle the import directly from local fs, but not adding another stuff to a proxy seems like a valid point. I will be back with update.

xprazak2:
We should be able to import local roles now.

app/services/foreman_ansible/roles_importer.rb
dLobatog: It is extra burden on the developer. It's so much easier for the user though, who doesn't have to care about proxy, proxy config files, or proxy plugins. 

I aim to alleviate the burden (duplication, really) between doing the same operation in proxy & Foreman via extracting this to a gem `ansible_roles_importer` or similar.

dLobatog:
@xprazak2 @tstrachota 's PR to include Roles in Host groups was merged just yesterday, mind to rebase?

dLobatog:
Looks like it's still out of date?

dLobatog:
Feel free to use https://github.com/dlobatog/foreman_ansible/tree/roles-import-oprazak - it went out of date again, I'm trying it out

app/helpers/foreman_ansible/ansible_roles_helper.rb
```diff
@@ -1,9 +1,21 @@
...
-    def ansible_roles
-      ForemanAnsible::RolesImporter.import!
-      AnsibleRole.all
+    def ansible_proxy_links(hash, classes = nil)
+      links = SmartProxy.with_features('Ansible').map do |proxy|
+        display_link_if_authorized(_('Import from %s') % proxy.name,
+                                   hash.merge(:proxy => proxy),
+                                   :class => classes)
+      end.flatten
+      links.unshift display_link_if_authorized(_('Import from localhost'), hash, :class => classes)
```
dLobatog: `Foreman host` instead.. otherwise it's kind of confusing especially if you have a localhost proxy

app/controllers/ansible_roles_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      process_success
+    else
+      process_error
+    end
+  end
+
+  def import
+    changed = @importer.import!
+    if changed.values.all? { |roles| roles.empty? }
+      notice _('No changes in roles detected on %s.') % @proxy.name
```
dLobatog: This doesn't work if I import from the host ifself, `@proxy` doesn't exist

app/views/ansible_roles/import.html.erb
```diff
@@ -0,0 +1,47 @@
+<% title _("Changed roles") %>
```
dLobatog: Changed Ansible roles, just like the index page - otherwise they could be regular Foreman RBAC roles

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -1,24 +1,69 @@
...
+      imported = roles.map do |role_name|
+        ::AnsibleRole.find_or_initialize_by(:name => role_name)
+      end
+      detect_changes imported
+    end
+
+    def finish_import(changes)
+      return unless  changes.present?
+      create_new_roles changes['new'] if changes['new']
+      delete_old_roles changes['obsolete'] if changes['obsolete']
```
dLobatog: It'd be nice to show in this table if there are hosts/hostgroups associated with these hosts, maybe we can do it later

app/lib/proxy_api/ansible.rb
```diff
@@ -0,0 +1,28 @@
...
+      EOFError,
+      Net::HTTPBadResponse,
+      Net::HTTPHeaderSyntaxError,
+      Net::ProtocolError,
+      RestClient::ResourceNotFound
+    ].freeze
+
+    def roles
+      parse(get('roles'))
+    rescue *PROXY_ERRORS => e
```
dLobatog: :clap:

app/views/ansible_roles/index.html.erb
dLobatog: fixed

dLobatog:
All good, thanks @xprazak2 - notice you're missing the permissions for this model hence tests are failing - http://ci.theforeman.org/job/test_plugin_pull_request/2861/database=mysql,label=fast,ruby=2.0.0/testReport/junit/(root)/HostgroupsControllerTest/test_new/

You'd need at least view, create and destroy permissions if I'm not wrong. Check https://github.com/theforeman/foreman_plugin_template/blob/master/lib/foreman_plugin_template/engine.rb#L24 or other plugins engine.rb for inspiration.

I saw the `AnsibleRolesHelperTest` is still removed, I guess that's by mistake, please recover it.
I'll merge after that, provided tests pass, which I think they'll do. Thanks again @xprazak2 !

https://github.com/theforeman/foreman_ansible/pull/29 completes this

xprazak2:
@dLobatog thanks for hints, tests are failing even though I have permissions:
https://github.com/theforeman/foreman_ansible/pull/25/files#diff-966ee0d35def95d426c9f4d348ffba11R33

I removed `AnsibleRolesHelperTest` since we no longer import roles through `ansible_roles` helper and it was the only method being tested there.

lib/foreman_ansible/engine.rb
```diff
@@ -25,17 +26,33 @@ class Engine < ::Rails::Engine
...
         requires_foreman '>= 1.12'
-
-        security_block :ansible do
+        security_block :foreman_ansible do
           permission :play_roles,
                      { :hosts => [:play_roles, :multiple_play_roles] },
                      :resource_type => 'Host::Managed'
+          permission :view_ansible_roles,
+                     { :ansible_roles => [:index],
+                       :hosts => [:new, :edit],
```
dLobatog: These two permissions (host new/edit and hostgroups new/edit) are overzealous. There is no reason why "view_ansible_roles" should allow users to modify hosts, it's unexpected

dLobatog:
@xprazak2 Thanks for the clarification. I looked at it again and it seems like you have the permissions but something's iffy with the way they're loaded. Also I commented inline because the 'view_ansible_roles' permission grants too much access. I do realize there might be gaps on restricting what roles are shown on host/hostgroup edit/new actions but we can fix that on another Pr.

For some reason at the time the hostgroup test runs, the permissions for AnsibleRole are not loaded, however if I put a breakpoint on `app/services/foreman/plugin.rb`#permission, it looks like it's creating them just fine. `Permission.select { |e| e.resource_type == 'AnsibleRole' }` returns nothing during hte test.

Please fix these tests and we can merge this, also comment on how you fixed them as I'm curious why it's not working :smile:

xprazak2:
@dLobatog I think I finally figured it out. Foreman core uses permission fixtures in tests instead of those defined in `services/foreman/access_permissions.rb` and  `security_block` does not load the plugin permissions in test env. When I do `Permission.all` during test, I get 143 permissions, all from core. I believe the reason we haven't encountered this before is because `AssociationAuthorizer` is called from `multiple_selects` helper and looks up the permissions, but no other plugin actually uses this helper AFAIK.

I moved `permission_block` into separate file so I could require it in our `test_helper` and now I am anxiously waiting if Jenkins turns green.

test/unit/ansible_role_test.rb
```diff
@@ -2,6 +2,10 @@
 
 # Tests for the behavior of Ansible Role, currently only validations
 class AnsibleRoleTest < ActiveSupport::TestCase
+  setup do
```
dLobatog: What is this needed for? It's creating a role on every should of this test with no purpose

dLobatog:
@xprazak2 Didn't seem to work - anyway you can test it locally with `ruby -I'test' test/unit/hostgroup_test.rb' for instance, no need to wait for Jenkins to finish. Rather than adding a require to every single test, I think a way to accomplish this like core does is to add a fixture https://github.com/theforeman/foreman/blob/develop/test/fixtures/permissions.yml , recently we had to add a few for view_params - thanks for the effort, it's an annoying thing to fix

xprazak2:
@dLobatog it worked when I tested locally but it does not matter now. I changed it to use fixtures as you suggested.

dLobatog:
@xprazak2 Thanks, I'm merging this :smile: 

If you can figure out a way to add these fixtures to `fixtures_path` without moving files around that'd be the best!

dLobatog:
@xprazak2 Merged, 8baaaed8114945177652e2d971d9605342f34894 , thanks for your patience making fixes for this, again I don't mean to be officious & add red tape, just keeping this repo neat & useful. :smile: 

This will be important for the next release.

PR #30, opened 2016-08-11
Fixes #15837 - Allow roles assignment to hostgroups

tstrachota:


tstrachota:
I found out that assigned roles aren't saving for both hosts and hostgroups anymore. Master is affected too. I'm investigating what happened.

tstrachota:
Solved, `parameter_filter` needs to be array for 1:N relations:
```ruby
parameter_filter Hostgroup, :ansible_role_ids => []
```

dLobatog:
Tests fixed here too https://github.com/theforeman/foreman_ansible/pull/31, on to the review..

app/views/foreman_ansible/ansible_roles/_included_roles.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="form-group">
```
dLobatog: I see what you tried to do here, but we follow a more intuitive model (e.g: user would wonder why do these roles not show on 'Selected items' on other inherited stuff.

screenshot from 2016-08-12 13-08-15

to look like this (example from nested taxonomies):

screenshot from 2016-08-12 13-12-28

app/models/concerns/foreman_ansible/hostgroup_extensions.rb
```diff
@@ -0,0 +1,20 @@
+module ForemanAnsible
+  # Relations to make Host::Managed 'have' ansible roles
```
dLobatog: Hostgroup, not Host::Managed

tstrachota:
Hmm, you're right, the selected items with tooltip look more intuitive. I'll re-visit the UI.

app/helpers/foreman_ansible/ansible_roles_helper.rb
```diff
@@ -5,5 +5,9 @@ def ansible_roles
       ForemanAnsible::RolesImporter.import!
       AnsibleRole.all
     end
+
+    def available_ansible_roles(target)
```
dLobatog: Looks like a good fit for a instance method on Host/Hostgroup `.available_roles` , through a concern they both can share it or something

app/models/concerns/foreman_ansible/hostgroup_extensions.rb
```diff
@@ -0,0 +1,20 @@
...
+  module HostgroupExtensions
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :hostgroup_ansible_roles, :foreign_key => :hostgroup_id
+      has_many :ansible_roles, :through => :hostgroup_ansible_roles,
+                               :dependent => :destroy
+
+      def all_ansible_roles
```
dLobatog: If you end up doing  https://github.com/theforeman/foreman_ansible/pull/30/files#r74576083 , this method could also go in the concern

app/views/foreman_ansible/hostgroups/_tab_content.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class='tab-pane' id='ansible_roles'>
+
```
dLobatog: Why these empty lines?

dLobatog:
Thanks - nearly everything is good I think, except for the duplication between Host/Hostgroup that I think we can solve easily, 'included roles' thing, and that method in the helper.  It doesn't pass rubocop - but I'll fix that.

I still have to look at tests but overall good :smile: - I'll test a few odd cases and see if it fails, comment on the current tests and hopefully we can merge this.

app/overrides/hostgroup_ansible_roles_tab.rb
```diff
@@ -0,0 +1,14 @@
+# Displays Ansible roles tab on Host form
```
dLobatog: Hostgroup

test/unit/concerns/host_managed_extensions_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_plugin_helper'
+
+class HostManagedExtensionsTest < ActiveSupport::TestCase
+  before do
+    @role1 = FactoryGirl.create(:ansible_role)
+    @role2 = FactoryGirl.create(:ansible_role)
+    @role3 = FactoryGirl.create(:ansible_role)
+
+    @hostgroup_parent = FactoryGirl.create(:hostgroup, :ansible_roles => [@role2])
+    @hostgroup = FactoryGirl.create(:hostgroup, :parent => @hostgroup_parent)
```
dLobatog: This is creating a lot of objects every time for no reason (they're not used everywhere)

dLobatog: Plus can you use `.build` instead ?

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -7,6 +7,15 @@ module HostManagedExtensions
       has_many :host_ansible_roles, :foreign_key => :host_id
       has_many :ansible_roles, :through => :host_ansible_roles,
                                :dependent => :destroy
+
+      def all_ansible_roles
+        (ansible_roles + parent_ansible_roles).uniq
+      end
+
+      def parent_ansible_roles
```
dLobatog: `inherited_ansible_roles` is a bit more descriptive IMO, hosts don't have "parent" (and this method returns more than just the parent hostgroup roles)

app/models/concerns/foreman_ansible/hostgroup_extensions.rb
```diff
@@ -0,0 +1,20 @@
...
+        (ansible_roles + parent_ansible_roles).uniq
+      end
+
+      def parent_ansible_roles
```
dLobatog: `inherited_ansible_roles`

test/unit/concerns/hostgroup_extensions_test.rb
```diff
@@ -0,0 +1,34 @@
+require 'test_plugin_helper'
+
+class HostgroupExtensionsTest < ActiveSupport::TestCase
+  before do
+    @role1 = FactoryGirl.create(:ansible_role)
+    @role2 = FactoryGirl.create(:ansible_role)
+    @role3 = FactoryGirl.create(:ansible_role)
+
+    @hostgroup_parent = FactoryGirl.create(:hostgroup, :ansible_roles => [@role2])
+    @hostgroup = FactoryGirl.create(:hostgroup, :ansible_roles => [@role1])
```
dLobatog: This is literally the same behavior tested twice. Even more reasons to move it to a concern and test it on a dummy object  - https://github.com/theforeman/foreman_ansible/pull/30/files#r74576165 -

dLobatog: I realize `parent_ansible_roles` is the only difference though

test/unit/concerns/host_managed_extensions_test.rb
```diff
@@ -0,0 +1,35 @@
+require 'test_plugin_helper'
+
+class HostManagedExtensionsTest < ActiveSupport::TestCase
+  before do
+    @role1 = FactoryGirl.create(:ansible_role)
+    @role2 = FactoryGirl.create(:ansible_role)
+    @role3 = FactoryGirl.create(:ansible_role)
```
dLobatog: Not used?

test/unit/concerns/hostgroup_extensions_test.rb
```diff
@@ -0,0 +1,34 @@
+require 'test_plugin_helper'
+
+class HostgroupExtensionsTest < ActiveSupport::TestCase
+  before do
+    @role1 = FactoryGirl.create(:ansible_role)
+    @role2 = FactoryGirl.create(:ansible_role)
+    @role3 = FactoryGirl.create(:ansible_role)
```
dLobatog: Not used?

test/unit/hostgroup_ansible_role_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_plugin_helper'
+
+class HostgroupAnsibleRoleTest < ActiveSupport::TestCase
+  should belong_to(:hostgroup)
+  should belong_to(:ansible_role)
+  should validate_presence_of(:ansible_role_id)
```
dLobatog: uniqueness & scope missing

dLobatog:
All set, thanks for doing this @tstrachota . I left a lot of comments but most of them are resolved by removing the duplication via a concern and other minor fixes. The only 'complicated' one is the UI thing but you have an example in taxonomies (maybe there are more). Thanks again!

test/unit/concerns/host_managed_extensions_test.rb
tstrachota: Not directly but I wanted to have one more role there to test that I'm actually not displaying what I shouldn't.

test/unit/concerns/host_managed_extensions_test.rb
tstrachota: I wasn't able to test the inheritance without saving the objects in the db, thus I didn't use build.

test/unit/concerns/hostgroup_extensions_test.rb
tstrachota: I moved the model code to a concern but the tests differ in setup logic so I left them as they were.

tstrachota:
@dLobatog updated. UI uses `multiple_selects` now.

app/views/foreman_ansible/ansible_roles/_select.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= multiple_selects(f, :ansible_roles, ansible_roles, object.all_ansible_roles.map(&:id),
```
dLobatog: I think this is where tests are failing. On a new host, `object` is nil. Call `f.object` instead of `object` and it'd be fixed

app/views/foreman_ansible/hosts/_tab_content.html.erb
```diff
@@ -1,4 +1,3 @@
 <div class='tab-pane' id='ansible_roles'>
-  <%= multiple_checkboxes f, :ansible_roles, @host, ansible_roles,
-    :label => _('Roles') %>
+  <%= render 'foreman_ansible/ansible_roles/select', :object => @hostgroup, :f => f %>
```
dLobatog: No need to pass any `object` if you just pass `f.` and call `f.object` on the partial

app/views/foreman_ansible/hostgroups/_tab_content.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class='tab-pane' id='ansible_roles'>
+  <%= render 'foreman_ansible/ansible_roles/select', :object => @hostgroup, :f => f %>
```
dLobatog: No need to pass any `object` given https://github.com/theforeman/foreman_ansible/pull/30/files#r74913763

app/views/foreman_ansible/hosts/_tab_content.html.erb
dLobatog: At that point, since you don't pass anything, you can reuse the _tab_content partial for both host and hostgroup btw

test/unit/hostgroup_ansible_role_test.rb
dLobatog: fixed below

dLobatog:
@tstrachota Thanks - notice the problem that's making tests fail. Other than that it should be ready :smile: 

I'll be fixing rubocop cops after this is merged btw

dLobatog:
All good now , will merge if tests are green :smile: Not sure what happened with CodeClimate but I don't think this PR caused it to fail. Thanks @tstrachota !

dLobatog:
I'll rebase this into 1 commit, remember to do it next time :smile:

dLobatog:
Merged as 78e6be98672d8b78416c67ece24caa8d329ab3ee, thanks @tstrachota !

PR #35, opened 2016-08-17
Fixes #16147 - set index names explicitly

iNecas:
To avoid limit on label size in Postgres

As part of the change, increasing the limit on lines size to 120,
80 it just too little.

### DIFF
--- a/db/migrate/20160802153302_create_join_table_hostgroup_ansible_roles.rb
+++ b/db/migrate/20160802153302_create_join_table_hostgroup_ansible_roles.rb
@@ -1,10 +1,10 @@
 # Defines the relation between Hostgroup and AnsibleRole
+# rubocop:disable Metrics/LineLength
 class CreateJoinTableHostgroupAnsibleRoles < ActiveRecord::Migration
   def change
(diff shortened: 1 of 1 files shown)


daviddavis:
Regarding the line size, I would prefer 80. I can only view about that many columns when I have split vim session on my laptop. However, I'd be fine with 100 since it's what we use on Katello.

iNecas:
80 lines is so 80's :D

daviddavis:
vim is so 80's (or 90's I guess) but I still enjoy using it 😄

dLobatog:
I'm of the same opinion as @daviddavis , it's annoying to not be able to see everything with vim split (even worse if you have split editor in a graphic IDE), can we do that in another PR?

iNecas:
oh my… whatever, updated to disable it just at the particular file. /me goes to :cry:  in a corner for a while but will be ok eventually.

iNecas:
Tests running here http://ci.theforeman.org/view/PR%20Tests/job/test_plugin_pull_request/2837/

dLobatog:
Merged as 66313ecff452a95c7825659289789a91a195fac3, thanks @iNecas!

PR #29, opened 2016-08-10
[WIP] Fixes #15856 - Play ansible roles through smart proxy

adamruzicka:

PR #40, opened 2016-08-25
Refs #15856 - Play ansible roles through smart proxy (rebase)

iNecas:
I've rebased #29 + did some review and small fixes to make it working.
There is quite a lot of duplicities with remote execution, as observed
in http://projects.theforeman.org/issues/16058.

I would like to give it a try and extract the commond code now, which
gives us also opportunity to clean the code a bit.

Also, after watching the community demo today, I've realized the plugin
want's to keep an ability to use the functionality without running a proxy.
Given we need the dynflow code to be present in different gem than the proxy
itself due to ability to run on RHEL6, I'm willing to try the following.

- [x] - teach foreman-tasks how to run the actions directly, without foreman proxy
- [x] - extract the common code it has with remote execution into a common place
- [x] - finish https://github.com/theforeman/foreman-tasks/pull/198
- [x] - update smart_proxy_ansible PR
 
If successful, it would give both this plugin and remote execution ability
to be used both with and without foreman proxy.


app/services/foreman_ansible/proxy_load_balancer.rb
```diff
@@ -0,0 +1,36 @@
+module ForemanAnsible
+  class ProxyLoadBalancer
```
dLobatog: This class is literally taken from RemoteExecution, move it into a gem then we can use it here?

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -14,5 +14,24 @@ def inherited_ansible_roles
         hostgroup.all_ansible_roles
       end
     end
+
+    def ansible_proxies(authorized = true)
```
dLobatog: This doesn't belong in Host::Managed, move it to a service or similar. It just needs 'location/organization' to work.

app/services/foreman_ansible/proxy_load_balancer.rb
iNecas: That's what I meant in the initial comment https://github.com/theforeman/foreman_ansible/pull/40#issue-173284489

iNecas:
I have extracted and simplified the code shared with remote execution and made it possible to run the actions directly, or via smart proxy, with sharing as much code as possible between them.

I also update the inventory generator to have similar output as the foreman_ansible_inventory.py. 

Depends on https://github.com/theforeman/foreman-tasks/pull/198

app/models/concerns/foreman_ansible/host_managed_extensions.rb
iNecas: Updated

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -8,11 +8,42 @@ def initialize(hosts)
       @hosts = hosts
     end
 
-    def tempfile
-      tempfile = Tempfile.new("foreman-#{SecureRandom.uuid}-inventory")
-      tempfile.write(hosts.map(&:fqdn).join('\n'))
-      tempfile.close
-      tempfile
+    def json
```
adamruzicka: Maybe `hash` or `to_hash` would be better. It seems counter-intuitive to have to call `.json.to_json` to actually get the json

app/lib/actions/foreman_ansible/play_host_roles.rb
```diff
@@ -0,0 +1,46 @@
+module Actions
+  module ForemanAnsible
+    class PlayHostRoles < Actions::EntryAction
+      include ::Actions::Helpers::WithContinuousOutput
+      include ::Actions::Helpers::WithDelegatedAction
+
+      def plan(host, proxy_selector = ::ForemanAnsible::ProxySelector.new)
```
adamruzicka: The `ProxySelector` was moved under `::ForemanTasks`

app/lib/actions/foreman_ansible/play_hosts_roles.rb
```diff
@@ -0,0 +1,26 @@
+module Actions
+  module ForemanAnsible
+    class PlayHostsRoles < Actions::ActionWithSubPlans
+
+      def plan(hosts)
+        plan_self(:host_ids => hosts.map(&:id))
+      end
+
+      def create_sub_plans
+        proxy_selector = ::ForemanAnsible::ProxySelector.new
```
adamruzicka: The `ProxySelector` was moved under `::ForemanTasks`

app/lib/actions/foreman_ansible/play_host_roles.rb
iNecas: Update the foreman-tasks code from PR: I've updated them altogether  https://github.com/theforeman/foreman-tasks/pull/198/commits/a6cf005703e59d6d08edfd3343c1b91a5e000d2f

app/lib/actions/foreman_ansible/play_hosts_roles.rb
iNecas: responded at https://github.com/theforeman/foreman_ansible/pull/40#discussion_r77330580

Gemfile
```diff
@@ -1,3 +1,4 @@
 source 'http://rubygems.org'
 
 gemspec
+gemspec :name => 'foreman_ansible_core'
```
dLobatog: Does this exist already?

dLobatog: nevermind I see it at the end of the PR

app/lib/actions/foreman_ansible/play_host_roles.rb
iNecas: I've realized I've forgot to push thte proxy selector class to the git :(, it's there now

iNecas:
I've added simple configuration mechanism to foreman-tasks-core in order to be able to set the settings from all of this variants:

1. in foreman directly, just by setting `SETTINGS[:ansible]` key, easiliy doable Foreman's in `settings.d` 
2. in smart-proxy by passing the smart-proxy settings (see https://github.com/theforeman/smart_proxy_ansible/pull/2 on how it's done0
3. in smart-proxy-dynflow-core, by placing the ansible.yml to `config/settings.d/ansible.yml` https://github.com/theforeman/smart_proxy_dynflow/pull/26 handles the configuration loading

All three variants should work now

iNecas:
Here are the PRs to refactor remote execution to use the same code base https://github.com/theforeman/foreman_remote_execution/pull/203 and https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/31

iNecas:
@dLobatog given we are running out of time pretty quicky to get this to Foreman 1.13, I would suggest a plan B:

1. polish the code (rubocop), making current test suit passing for the changes
2. merge
3. release
4. installer updates
5. finish unit tests

I know it's not ideal but I can't think of any other realistic option at the moment.

iNecas:
I've rebased and tried to make rubocop happy: it was still yelling at me locally about some code I haven't even touched

dLobatog:
@iNecas Here are my results from some testing (using https://github.com/theforeman/smart_proxy_dynflow/pull/26, without it it wouldn't work):

* Reports failure or success just fine when there's a proxy with the Ansible feature, and runs the specified roles :white_check_mark: 
* Changes the configuration status correctly depending on the Task result :white_check_mark: 
* If there are no Ansible proxies, it fails to run :negative_squared_cross_mark: 
* Host and hostgroup parameters are ignored. I have a role that installs a package coming from Foreman, 'surefail_package', I set a host parameter with that name, value 'tig', but it still installs the default package in the role. :negative_squared_cross_mark: 

I'll check what's wrong with the parameters, if you want to do installer changes already to support the Ansible proxy and so forth, we can do this in parallel (like a job invocation :grin:)

app/lib/actions/foreman_ansible/play_host_roles.rb
```diff
@@ -0,0 +1,56 @@
...
+  module ForemanAnsible
+    # Actions that initiaztes the playbook run for roles assigned to
+    # the host. It doest that either locally or via a proxy when available.
+    class PlayHostRoles < Actions::EntryAction
+      include ::Actions::Helpers::WithContinuousOutput
+      include ::Actions::Helpers::WithDelegatedAction
+
+      def plan(host, proxy_selector = ::ForemanAnsible::ProxySelector.new)
+        input[:host] = { :id => host.id, :name => host.fqdn }
+        proxy = proxy_selector.determine_proxy(host)
```
dLobatog: Currently the action fails at this point when no proxy is available

adamruzicka:
> If there are no Ansible proxies, it fails to run

"no Ansible proxies" as in "no proxy has the Ansible feature" or "all proxies with the Ansible feature are down"?

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -8,11 +8,42 @@ def initialize(hosts)
...
+    end
+
+    def host_params(host)
+      render_rabl(host.host_parameters.authorized, 'api/v2/parameters/base')
+    end
+
+    def host_facts(host)
+      values = FactValue.authorized(:view_facts).my_facts.
+               no_timestamp_facts.where(:host_id => host.id)
+      FactValue.build_facts_hash(values.all)
```
dLobatog: This shouldn't contain all facts, (more updated) Ansible facts are already present at playbook runtime - I don't see a lot of value on providing Puppet facts too but if you insist :)

dLobatog: It also takes quite a while to process this, I measured it and it's the longest task when creating the inventory

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -8,11 +8,42 @@ def initialize(hosts)
...
+    def host_roles(host)
+      host.all_ansible_roles.map(&:name)
+    end
+
+    def host_attributes(host)
+      render_rabl(host, 'api/v2/hosts/main')
+    end
+
+    def host_params(host)
+      render_rabl(host.host_parameters.authorized, 'api/v2/parameters/base')
```
dLobatog: Call `host.host_params` instead. The reasons are two:

* `host.host_parameters` does NOT contain hostgroup parameters (or any inherited parameters)
* `host.host_params` provides params in JSON already, so you don't have to render anything

dLobatog:
@adamruzicka I tried with the second one, shutting down all proxies with Ansible, but I assume the former option will fail too because of what I saw in the action

adamruzicka:
@dLobatog removing the proxy completely works (at least on my machine). I can think of reasons for both why it should and why it shouldn't fail at that moment. How about making this configurable in settings?

app/lib/actions/foreman_ansible/play_host_roles.rb
adamruzicka: In any case it should fail in `plan_delegated_action`, not while determining the proxy

dLobatog:
@adamruzicka Making what configurable? I'll try removing the proxy then

Gemfile
```diff
@@ -1,3 +1,7 @@
 source 'http://rubygems.org'
 
-gemspec
+gemspec :name => 'foreman_ansible'
+
+gem 'foreman_ansible_core', :path => '.'
```
dLobatog: Mistake?

foreman_ansible_core.gemspec
```diff
@@ -0,0 +1,22 @@
...
+Ansible integration with Foreman - core parts for dealing with Ansible concepts,
+usable by foreman_ansible or smart_proxy_ansible to delegate the execution.
+DESC
+  s.licenses    = ['GPL-3']
+
+  s.files = Dir['lib/foreman_ansible_core/**/*'] +
+            ['lib/foreman_ansible_core.rb', 'bin/json_inventory.sh']
+
+  s.add_development_dependency 'rubocop', '~> 0.42'
+  s.add_dependency 'foreman-tasks-core', '~> 0.8.0'
```
dLobatog: Also on `foreman_ansible_core`

app/services/foreman_ansible/inventory_creator.rb
iNecas: The only reason was to mimic the foreman ansible invnetory script, where they do that. I don't insist on that being there. Will remove

app/lib/actions/foreman_ansible/play_host_roles.rb
iNecas: The behaviour right now is:

1. if there is a proxy available for the host with ansible feature and it's running, it tries use it
2. if there a proxy available for the host with ansible feature, but it's not running, fail, because the user probably expected to use the proxy
3. if there is no proxy available with ansible feature, use the build in

It's not this way to address the simple-to start case, while respecting requirements on more complex network topologies and production system, where it's more preferred to use the proxy. For example, if I have some remote location with a proxy and hosts are not reachable from the main server, I would not expect the main server to try running directly, even if the proxy is down at the moment.

foreman_ansible_core.gemspec
iNecas: this is foreman_ansible_core gemspec, the recursive dependency is not what we want probably…

dLobatog: My bad, I didn't see it in the other one

app/services/foreman_ansible/inventory_creator.rb
dLobatog: :+1: , also explain the all/_meta thing in a comment

app/services/foreman_ansible/inventory_creator.rb
dLobatog: The inventory script shouldn't do it too either, the issue is that there was no endpoint to get just  non-Ansible facts (IIRC)

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -8,11 +8,42 @@ def initialize(hosts)
...
     end
 
-    def tempfile
-      tempfile = Tempfile.new("foreman-#{SecureRandom.uuid}-inventory")
-      tempfile.write(hosts.map(&:fqdn).join('\n'))
-      tempfile.close
-      tempfile
+    def json
+      { 'all' => { 'hosts' => hosts.map(&:fqdn) },
+        '_meta' => { 'hostvars' => host_vars } }
```
dLobatog: How does one address params in Ansible with this? host_vars looks like:

```ruby
{ 'toshiba.lobatolan.home' =>  
   { 'foreman_params' =>
     [ {:id=>10, :name=>"surefail_package", :value=>"tig"}]
   }
}
```

iNecas: Right, again, I've used the inventory script as basis, but probably not interpreted that well… anyway, sould be in better format now, using the host_params, as suggested

adamruzicka: The contents of the `json['_meta']['hostvars'][fqdn]` gets assigned to the top level, so `json['_meta']['hostvars']['localhost']['foreman']` can be accessed just as `foreman` in the playbook

See [ansible docs](http://docs.ansible.com/ansible/developing_inventory.html#tuning-the-external-inventory-script) for details on the `_meta` magic

app/services/foreman_ansible/inventory_creator.rb
iNecas: remove the facts

iNecas: It was also probably the reason why you were hitting the dynflow data attribute limit issue

app/services/foreman_ansible/inventory_creator.rb
dLobatog: I see, thanks. I've set my variable in the playbook to `foreman['surefail_package']` and it did not work either,  'dict object' has no attribute 'surefail_package'`. 

The `foreman` dict contains this:
```json
    "msg": "From Foreman host {u'comment': u'', u'environment_id': 1, u'managed': False, u'model_id': 2, u'ip': u'192.168.0.106', u'provision_method': u'build', u'last_compile': u'2016-09-07T09:04:05.000Z', u'updated_at': u'2016-09-07T09:04:28.000Z', u'realm_name': None, u'puppet_status': 16781312, u'sp_mac': None, u'operatingsystem_name': u'Fedora release 24 (Twenty Four)', u'disk': None, u'location_id': 1, u'domain_id': 1, u'location_name': u'Spain', u'compute_resource_id': None, u'subnet_id': None, u'sp_name': None, u'environment_name': u'production', u'architecture_name': u'x86_64', u'domain_name': u'lobatolan.home', u'capabilities': [u'build'], u'id': 2, u'organization_name': u'Lobacorp', u'global_status': 2, u'architecture_id': 1, u'build': False, u'hostgroup_title': None, u'subnet_name': None, u'sp_ip': None, u'model_name': u'Satellite Radius P55W-B', u'global_status_label': u'Error', u'owner_id': 4, u'operatingsystem_id': 2, u'sp_subnet_id': None, u'configuration_status_label': u'Error', u'name': u'toshiba.lobatolan.home', u'certname': u'toshiba.lobatolan.home', u'image_name': None, u'organization_id': 2, u'image_id': None, u'mac': u'60:57:18:52:89:97', u'ip6': None, u'puppet_ca_proxy_id': 2, u'last_report': u'2016-09-07T07:03:52.000Z', u'subnet6_id': None, u'puppet_proxy_id': 2, u'compute_profile_name': None, u'uuid': None, u'ptable_name': None, u'image_file': u'', u'compute_resource_name': None, u'created_at': u'2016-08-10T18:30:01.000Z', u'enabled': True, u'hostgroup_id': None, u'configuration_status': 16781312, u'hostgroup_name': None, u'use_image': None, u'compute_profile_id': None, u'medium_id': None, u'subnet6_name': None, u'ptable_id': None, u'owner_type': u'User', u'realm_id': None, u'installed_at': None, u'medium_name': None}"
```

and `foreman_params` contains this:

```json
    "msg": "From Foreman host [{u'id': 10, u'value': u'tig', u'name': u'surefail_package'}]"
```

The latter is fine, it's *almost*how the inventory does it. It should be:

```json
 { 'host_param_name' : 'host_param_value', 'host_param2_name' : 'host_param2_value' }
```

It's a small change but with that the inventory and foreman_params have the same format and playbooks can be reused.

See https://github.com/theforeman/foreman_ansible_inventory#variables-and-parameters

iNecas: updated

Gemfile
iNecas: I though I needed that, but it seems it's not needed. Removed

Gemfile
```diff
@@ -2,6 +2,5 @@ source 'http://rubygems.org'
 
 gemspec :name => 'foreman_ansible'
 
-gem 'foreman_ansible_core', :path => '.'
 gem 'foreman-tasks', :path => '../foreman-tasks'
```
dLobatog: These 2 can also be removed

dLobatog:
Success, now `host_params` do inherit from hostgroup/global/taxonomies and are presented in the same format as the inventory :smile: 

```
    "msg": "From Foreman host {u'remote_execution_ssh_user': u'root', u'enable-puppetlabs-pc1-repo': u'true', u'enable-puppetlabs-repo': u'true', u'surefail_package': u'tig'}"
```

app/lib/actions/foreman_ansible/play_host_roles.rb
dLobatog: Fair enough, maybe this should be configurable in -tasks for all stuff that use determine_proxy, not just here. In any case it's fine.

dLobatog:
Last think I'll test is whether it uses Foreman when the proxy with 'Ansible' is in taxonomies different to the ones of the host.

Gemfile
iNecas: updated

app/services/foreman_ansible/proxy_selector.rb
```diff
@@ -0,0 +1,21 @@
+module ForemanAnsible
+  # Contains proxy selection rules for a host playbook run
+  class ProxySelector < ::ForemanTasks::ProxySelector
+    # rubocop:disable Metrics/MethodLength
+    def available_proxies(host)
+      proxies = {}
+
+      proxies[:fallback] = host.smart_proxies.with_features('Ansible')
+      proxy_scope = if Taxonomy.enabled_taxonomies.any?
+                      ::SmartProxy.with_taxonomy_scope_override(host.location,
```
dLobatog: This doesn't work. I have a host with:

Location: 'Czech Republic' (no proxies)
Organization: 'E-Corp' (no proxies)

But this method still returns proxies in other organizations/locations? I'll investigate as I think it's a problem in core.

dLobatog: Nevermind, it does work after reverting https://github.com/theforeman/foreman/commit/35ed04f95f14c934ffebc874e90a3a4a6684fe71

iNecas:
I have released the foreman-tasks bits, let's see what Mr. Jenkins sais

PR #41, opened 2016-08-26
Fixes #16335 - Improve the strings for translators

bkearney:
* Prefer using interpolation of concats
* Strings should be complete

Made the changes above, and extracted latest strings 
to ensure that all the current strings were reviewed.


dLobatog:
Merged as f6f1a78b8e9e135c8904435f5ab605a9410d649f, thanks @bkearney!

PR #45, opened 2016-09-06
Fixes #16442 - Importing from proxy cannot show 'no changes'

dLobatog:
It tries to use a proxy variable that's not defined, it should use
'@proxy'


iNecas:
ACK

PR #47, opened 2016-09-13
Fixes #16537 - Tests to ensure play_roles work

dLobatog:
Includes:
  - Fixes for rubocop errors
  - Add foreman_ansible_core tests
  - Fixes and tests for proxy selector
  - Controller tests

### DIFF
--- a/lib/foreman_ansible_core/playbook_runner.rb
+++ b/lib/foreman_ansible_core/playbook_runner.rb
@@ -1,10 +1,9 @@
 require 'foreman_tasks_core/runner/command_runner'
 require 'tmpdir'
-require 'securerandom'
 
 module ForemanAnsibleCore
   # Implements ForemanTasksCore::Runner::Base interface for running
-  # Ansible playbooks
+  # Ansible playbooks, used by the Foreman Ansible plugin and Ansible proxy
   class PlaybookRunner < ForemanTasksCore::Runner::CommandRunner
     def initialize(inventory, playbook)
       super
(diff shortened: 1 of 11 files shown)


lib/foreman_ansible_core/playbook_runner.rb
```diff
@@ -72,19 +71,27 @@ def ensure_directory(path)
...
         @working_dir = Dir.mktmpdir
         @tmp_working_dir = true
       end
     end
+
+    def initialize_ansible_dir(ansible_dir)
+      if File.exist?(ansible_dir)
+        @ansible_dir = ansible_dir
+      else
+        raise "Ansible dir #{@ansible_dir} does not exist"
```
iNecas: `s/@ansible_dir/ansible_dir`

test/support/foreman_tasks/task.rb
```diff
@@ -0,0 +1,47 @@
+module Support
+  module ForemanTasks
+    module Task
+      def stub_tasks!
```
iNecas: Seems like a copy-paste from https://github.com/Katello/katello/blob/master/test/support/foreman_tasks/task.rb, probably right time to extract this to foreman-tasks, to be reusable across projects.

iNecas:
You mentioned "Fixes and tests for proxy selector", what was broken there?

test/support/foreman_tasks/task.rb
dLobatog: Indeed

dLobatog:
@iNecas Nothing, I think I wrote the comment when https://github.com/theforeman/foreman/pull/3824 was still a problem. The only functional change of this PR is adding 'error' messages to the controllers. I'm still not sure if we should cast `:not_defined` to `:not_available`.. it seems rather annoying if someone has a proxy in Organization A with Ansible, and users in Organization B (which doesn't have a proxy nor want it) are forced to install one now

iNecas:
ACK

dLobatog:
Last commit before releasing this. Thanks @iNecas !

PR #49, opened 2016-09-19
Fixes #15976 - Add more info about Ansible roles

xprazak2:
This extracts the common code for retrieving roles into a separate gem (foreman_ansible_exporter)
that can be used by both Foreman and Smart Proxy. It also exposes more details about the imported roles.

foreman_ansible_exporter will be listed as a dependency in gemspec once released, for now I just run from [source](https://github.com/xprazak2/foreman_ansible_exporter) and drop `gemspec :path => path/to/foreman_ansible_exporter` into my Gemfile.local.rb in core's bundler.d.

I realize this may not be entirely the way we want to go for a couple of reasons, but anyway. Let me know what you think. If we decide this is a way to go, I will gladly add tests and corresponding changes to `smart_proxy_ansible`

### DIFF
--- a/app/controllers/ansible_files_controller.rb
+++ b/app/controllers/ansible_files_controller.rb
@@ -1,0 +1,95 @@
+# UI controller for ansible files
+class AnsibleFilesController < ::ApplicationController
+  include Foreman::Controller::AutoCompleteSearch
+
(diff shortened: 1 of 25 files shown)


xprazak2:
I think we can close this since it has more problems than benefits.

PR #51, opened 2016-09-28
Fixes #16714 - Add apipie initializer

dLobatog:
This initializer was missing and it was required to generate the apipie
cache and docs.


dLobatog:
Merged to fix tests

PR #54, opened 2016-11-18
Fixes #17397 - Use StructuredFactImporter on 1.13+

dLobatog:
foreman-ansible provided its own parser for parsing structured facts.
However this is not needed on 1.13+ versions, as the Facter 2+ parser in
core is able to parse Ansible facts just fine, even better in some
cases.


agx:
Looks good to me!

dLobatog:
@agx thanks 😄

PR #60, opened 2016-11-30
Fixes #17150 - Allow setting some ansible connection parameters

tbrisker:
This currently allows setting the ansible_user, ansible_port and
ansible_ssh_pass for all hosts. There is a default setting which can
also be overriden by using host or hostgroup parameters with the same
name.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -1,0 +1,18 @@
+class Setting::Ansible < ::Setting
+  def self.load_defaults
+    return unless super
+    self.transaction do
+      [
+        self.set('ansible_port', N_('Foreman will use this port to ssh into hosts for running playbooks'), 22, N_('Default port')),
+        self.set('ansible_user', N_('Foreman will try to connect as this user to hosts when running Ansible playbooks.'), 'root', N_('Default user')),
+        self.set('ansible_ssh_pass', N_('Foreman will use this password when running Ansible playbooks.'), 'ansible', N_('Default password'))
+      ].compact.each { |s| self.create s.update(:category => 'Setting::Ansible') }
+    end
+
+    true
+  end
+
+  def self.humanized_category
+    N_('Ansible')
+  end
+end
--- a/app/services/foreman_ansible/inventory_creator.rb
+++ b/app/services/foreman_ansible/inventory_creator.rb
@@ -17,16 +17,34 @@ module ForemanAnsible
     # with all hosts.
     def to_hash
       { 'all' => { 'hosts' => hosts.map(&:fqdn) },
-        '_meta' => { 'hostvars' => host_vars } }
+        '_meta' => { 'hostvars' => hosts_vars } }
     end
 
-    def host_vars
+    def hosts_vars
       hosts.reduce({}) do |hash, host|
-        hash.update(host.fqdn =>
-                    { 'foreman' => host_attributes(host),
-                      'foreman_params' => host_params(host),
-                      'foreman_ansible_roles' => host_roles(host) })
+        hash.update(host.fqdn => host_vars(host))
       end
+    end
+
+    def host_vars(host)
+      {
+        'foreman' => host_attributes(host),
+        'foreman_params' => host_params(host),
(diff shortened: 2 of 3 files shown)


app/models/setting/ansible.rb
```diff
@@ -0,0 +1,19 @@
+class Setting::Ansible < ::Setting
+  def self.load_defaults
+    return unless ActiveRecord::Base.connection.table_exists?('settings')
```
dLobatog: When does this early return condition happen?

app/models/setting/ansible.rb
```diff
@@ -0,0 +1,19 @@
...
+    return unless super
+    self.transaction do
+      [
+        self.set('ansible_port', N_('Foreman will use this port to ssh into hosts for running playbooks'), 22, N_('Default port')),
+        self.set('ansible_user', N_('Foreman will try to connect as this user to hosts when running Ansible playbooks.'), 'ansible', N_('Default user')),
```
dLobatog: `root` seems like a better default

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -22,10 +22,14 @@ def to_hash
...
     def host_vars
       hosts.reduce({}) do |hash, host|
-        hash.update(host.fqdn =>
-                    { 'foreman' => host_attributes(host),
+        hash.update(host.fqdn => {
+                      'foreman' => host_attributes(host),
                       'foreman_params' => host_params(host),
-                      'foreman_ansible_roles' => host_roles(host) })
+                      'foreman_ansible_roles' => host_roles(host),
+                      'ansible_port' => host_port(host),
```
dLobatog: Please add the old version of these params to ensure compatibility with Ansible 1.x (`ansible_ssh_user', 'ansible_ssh_port') - no harm on sending a few more bytes for this

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -41,6 +45,19 @@ def host_params(host)
       host.host_params
     end
 
+    # TODO - make these overridable using parameters
```
dLobatog: No TODO comments please

dLobatog:
Works well btw - I'm able to bork my installation by changing the settings 😄  - "Failed to connect to the host via ssh: ssh: connect to host localhost.lobatolan.home port 3201: `

app/models/setting/ansible.rb
tbrisker: Not sure, this is from the instructions in http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin

tbrisker:
@dLobatog Thanks, addressed your comments and updated to also allow overriding using parameters.

app/services/foreman_ansible/inventory_creator.rb
dLobatog: fixed

dLobatog:
@tbrisker Looks good to me, thanks for adding the host params part. Can you fix the rubocop errors that CodeClimate is complaining about and I'll merge after that? They're all fairly simple, the 'trickiest' one is reducing the ABC for load_defaults should be as simple as moving the ansible settings to a method like 

```ruby

  def self.ansible_settings
    [
      self.set('ansible_port', N_('Foreman will use this port to ssh into hosts for running playbooks'), 22, N_('Default port')),
      self.set('ansible_user', N_('Foreman will try to connect as this user to hosts when running Ansible playbooks.'), 'root', N_('Default user')),
      self.set('ansible_ssh_pass', N_('Foreman will use this password when running Ansible playbooks.'), 'ansible', N_('Default password')),
    ]
  end
```

tbrisker:
@dLobatog I fixed all of the rubocop issues except int the settings which I left the same to be consistent with how we declare settings in all plugins.

dLobatog:
Ok fair enough, I'll change it on a different PR to see how it is or disable cops in that file perhaps.

dLobatog:
Thanks @tbrisker ! I'll make a new minor release with this tomorrow.

PR #64, opened 2016-12-15
Fixes #17702 - Make ansible run cancelable

chilicat:
Make ansible task cancelable by killing the ansible process.

### DIFF
--- a/lib/foreman_ansible_core/playbook_runner.rb
+++ b/lib/foreman_ansible_core/playbook_runner.rb
@@ -29,6 +29,13 @@ module ForemanAnsibleCore
       command
     end
 
+    def kill
+      publish_data('== TASK ABORTED BY USER ==', 'stdout')
+      publish_exit_status(1)
+      ::Process.kill('SIGTERM', @command_pid)
+      close
+    end
+
     def close
       super
       FileUtils.remove_entry(@working_dir) if @tmp_working_dir


dLobatog:
[test]

dLobatog:
@chilicat This doesn't do much by itself. I don't have anything against this code but it'll be dead code if nothing uses it. You'd have to add functionality to call this from either the proxy or Foreman for it to work. Were you missing some extra files in the commit or something similar?

Thanks for the contribution!

chilicat:
In my testing the cancel button in the task view was not able cancel the running ansible task. By adding this simple lines everything works as expected. 

Why do you think it is dead code?

chilicat:
Kill can be overwritten:
https://github.com/theforeman/foreman-tasks/blob/5b9507dd42bb98976344e2adaf5f2c1c87c2df71/lib/foreman_tasks_core/runner/base.rb#L36

chilicat:
I have improved the implementation. The task has behaved different in a development environment, it was not enough to just kill the process.

dLobatog:
[test] - @chilicat I was dumb enough not to look into foreman-tasks-core. Will merge if tests pass 😄 thanks again

dLobatog:
Thanks @chilicat !

chilicat:
Thank you for the effort reviewing and commenting.
Happy holidays.

PR #65, opened 2016-12-15
fixes #17701 - Ansible Button on Host page emits two ansible tasks

chilicat:
Steps to reproduce:
1. Apply role on a single host "myhost"
2. got to host page: https://katello/hosts/myhost.mydomain
3. Click "|> Ansible roles"
4. Goto Tasks view (https://katello/foreman_tasks/tasks)
5. Notice that two task are emitted (Name: Play Ansible roles on host /myhost.mydomain)

I used a chrome browser

This issue gave has particular headache since the visible ansible run was failing with all kind of weird issues caused by the second (hidden) ansible run.

### DIFF
--- a/app/helpers/foreman_ansible/hosts_helper_extensions.rb
+++ b/app/helpers/foreman_ansible/hosts_helper_extensions.rb
@@ -13,7 +13,8 @@ module ForemanAnsible
         icon_text('play', ' ' + _('Ansible roles'), :kind => 'fa'),
         play_roles_host_path(:id => args.first.id),
         :id => :ansible_roles_button,
-        :class => 'btn btn-default'
+        :class => 'btn btn-default',
+        :method => :post
       )
       title_actions(button_group(button)) if args.first.ansible_roles.present?
       host_title_actions_without_run_ansible_roles(*args)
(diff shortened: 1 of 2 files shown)


dLobatog:
[test]

dLobatog:
@chilicat This is good, but I want to keep a way to run ansible tasks on demand through the API. I'll make a release with your other PRs except this one - I'll add an endpoint for Ansible tasks to the API, merge this and make another release with both patches, just to not break the endpoint /hosts/ansible_roles  for current users.

dLobatog:
@chilicat Also mind to add a functional test to ensure everything keeps on working correctly?

chilicat:
@dLobatog thanks for looking into my PRs. 

How do I run the unit tests locally? 

"I'll add an endpoint for Ansible tasks to the API" 
Do you mean you add a POST endpoint in addition to the GET to keep the backwards compatibility?

"but I want to keep a way to run ansible tasks on demand through the API"
That should be still possible but via POST, which makes more sense in my opinion since running ansible is applying a change where GET is for readonly resources.

dLobatog:
@chilicat 

To run the tests locally, from the foreman directory you can run `rake test:foreman_ansible`. If you look at `lib/tasks/foreman_ansible_tasks.rake` in this repository, that's where we've put the task.

POST is fine, I tried to express what I think would make the most sense in https://github.com/theforeman/foreman_ansible/pull/67#pullrequestreview-14113209

chilicat:
Issue is fixed by #67

PR #67, opened 2016-12-21
fixes #17802 - Playbook execution on hostgroup-level

ghost:
Playbook execution on hostgroup-level and API extensions for triggering
playbooks

In some situations, it is necessary to run to run Ansible roles not only
for each host separately, but also for all members of a hostgroup in
parallel. Therefore, there should be a way to trigger Ansible plays for
hostgroups without splitting the playbook execution into subtasks for
each host. This is of special interest if a configuration process spans
multiple systems at the same time (for example cluster provisions).
In addition to that, playbooks should also be able to be triggered using
the v2 API, such that the user does not need to rely on the Foreman UI
to trigger Ansible roles.

Proposed new endpoints for the extending the API are:

POST http://<foreman_host>/ansible/api/v2/play_roles_on_host
POST http://<foreman_host>/ansible/api/v2/play_roles_on_hostgroup


app/controllers/api/v2/play_roles_controller.rb
```diff
@@ -0,0 +1,93 @@
+module Api
+  module V2
+    # API controller for Ansible Plays
+    class PlayRolesController < ::Api::V2::BaseController
+      include ::Api::Version2
+      include ForemanTasks::Triggers
+
+      rescue_from ActiveRecord::RecordNotFound, :with => :resource_not_found
```
dLobatog: This is already covered by application_controller.rb in core

app/controllers/api/v2/play_roles_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      before_action :find_hosts, :only => [:play_roles_on_host]
```
dLobatog: If you extend hosts/hostgroups separately, you can rely on `find_resource` from core to do this (https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/architectures_controller.rb#L7)

app/controllers/api/v2/play_roles_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+
+        @hostgroups.uniq.each do |hostgroup|
+          @result.append(
+            :hostgroup => hostgroup, :foreman_tasks => async_task(
+              ::Actions::ForemanAnsible::PlayHostgroupRoles, hostgroup
+            )
+          )
+        end
+
+        render :json => @result
```
dLobatog: You might want to `process_response @result` instead, to provide some feedback if making the tasks went wrong

app/controllers/api/v2/play_roles_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+        @hosts = []
+        host_ids.uniq.each do |host_id|
+          @hosts.append(Host.find(host_id))
+        end
+        host_names.uniq.each do |host_name|
+          @hosts.append(Host.find_by(:name => host_name))
+        end
+      end
+
+      def find_hostgroups
```
dLobatog: Ditto, `find_resource` from core provides this functionality

app/controllers/api/v2/play_roles_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      protected
+
+      def resource_not_found(e)
+        @result = { :error => e.message }
+        render :json => @result, :status => :error
+      end
+
+      private
+
+      def find_hosts
```
dLobatog: Ditto, `find_resource` from core provides this functionality

app/controllers/foreman_ansible/concerns/hostgroups_controller_extensions.rb
```diff
@@ -0,0 +1,29 @@
+module ForemanAnsible
+  module Concerns
+    # Extra methods to enforce Ansible roles on a host or multiple hosts
+    module HostgroupsControllerExtensions
+      extend ActiveSupport::Concern
+      include ForemanTasks::Triggers
+
+      def play_roles
```
dLobatog: If we're going to use the API for this - mind to extend the api/v2/hostgroups_controller instead of the regular GET action? Other than that it looks great

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
...
+    class PlayHostgroupRoles < Actions::EntryAction
+      include ::Actions::Helpers::WithContinuousOutput
+      include ::Actions::Helpers::WithDelegatedAction
+
+      def plan(hostgroup, proxy_selector = ::ForemanAnsible::ProxySelector.new)
+        if hostgroup.hosts.empty?
+          raise ::Foreman::Exception.new(N_('hostgroup is empty'))
+        end
+        input[:hostgroup] = { :id => hostgroup.id, :name => hostgroup.name }
+        proxy = proxy_selector.determine_proxy(hostgroup.hosts[0])
```
dLobatog: Not sure `hostgroup.hosts[0]` is the right choice here? It's possible that the proxy isn't even able to reach other hosts. I think an easy way to circumvent the issue here would be to start one `PlayHostRoles` task per host instead of creating the playbook yourself

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
+module Actions
+  module ForemanAnsible
+    # Actions that initiaztes the playbook run for roles assigned to
```
dLobatog: typo, initiaztes, doest

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
...
+          raise ::Foreman::Exception.new(N_('hostgroup is empty'))
+        end
+        input[:hostgroup] = { :id => hostgroup.id, :name => hostgroup.name }
+        proxy = proxy_selector.determine_proxy(hostgroup.hosts[0])
+        inventory_creator = ::ForemanAnsible::InventoryCreator.new(hostgroup.hosts)
+        role_names = []
+        hostgroup.hostgroup_ansible_roles.each do |ansible_role|
+          role_names.append(ansible_role.ansible_role_name)
+        end
+        playbook_creator = ::ForemanAnsible::PlaybookCreator.new(role_names)
```
dLobatog: This playbook just contains hostgroup ansible roles, but not all roles for a host. That's unexpected IMO (when I click on 'Play roles' I expect all roles applicable to a host to run, host-specific and hostgroup inherited)

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
...
+        ::Dynflow::Action::Rescue::Fail
+      end
+
+      def humanized_input
+        _('on hostgroup %{name}') %
+          { :name => input.fetch(:hostgroup, {})[:name] }
+      end
+
+      def humanized_name
+        _('Play Ansible roles')
```
dLobatog: `Host group Ansible roles`?

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
...
+        return unless delegated_output[:exit_status].to_s != '0'
+        error! _('Playbook execution failed')
+      end
+
+      def rescue_strategy
+        ::Dynflow::Action::Rescue::Fail
+      end
+
+      def humanized_input
+        _('on hostgroup %{name}') %
```
dLobatog: host group

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def humanized_output
+        continuous_output.humanize
+      end
+
+      def continuous_output_providers
+        super << self
+      end
+
+      def fill_continuous_output(continuous_output)
```
dLobatog: Seems like a good deal of the actions here are shared with PlayRolesHost, they should be extracted to a concern

app/overrides/hostgroup_play_roles.rb
```diff
@@ -0,0 +1,16 @@
+# Displays Ansible roles tab on Hostgroup form
+Deface::Override.new(
+  :virtual_path => 'hostgroups/index',
+  :name => 'ansible_row_title',
+  :insert_after => "erb[loud]:contains('Actions')",
+  :partial => 'foreman_ansible/ansible_roles/ansible_row_title',
+  :original => '016f2f86bd199275e7f3bebef77fefa90dd218f1'
```
dLobatog: original isn't really needed, and it's unreliable, it can change upon the slightest change on the code related with the view, please remove them

app/controllers/api/v2/play_roles_controller.rb
Gerrit91: Hey, first of all, thank you very much for the review! I'm looking forward to learn a lot from the input. :)

Unfortunately, this is the first time I work with Ruby, which is why I would appreciate some more information regarding these changes. I saw that `find_resource` was used in the other controllers, but I was not sure how this mechanism can be applied to my controller as well. Do you mean that I should remove the play_roles_controller and instead add the methods to the API v2 hosts_controller as hosts_controller_extensions (respectively hostgroups_controller and hostgroups_controller_extensions)?

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
Gerrit91: Btw, the typos are also contained in play_host_roles in the master.

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
Gerrit91: I just want to make clear that we want to play roles on a group of hosts with Ansible, like:

```
PLAY [all] *********************************************************************

TASK [install : run installer] ***

ok: [node01]
ok: [node02]
```

Starting a single play for each host is not wanted because it implies several problems. Basically, the control over the orchestration gets lost. One problem is that each host proceeds in its own tempo, but this is actually what [Ansible strategies](http://docs.ansible.com/ansible/playbooks_strategies.html) are responsible for. Another problem is that a failure of one host does not stop the entire provisioning process and leaves the hostgroup in an uncontrolled state. Sometimes, timing between the tasks is essential and Ansible tasks strictly need to be executed linearly. 
 
I also see that you are a bit concerned with the implementation here because it does not honour the concept of smart proxies well. However, I don't know how a better solution should look like and if the playbook fails because hosts are unreachable, I think is reasonable. With this background, do you have another idea of how to do this better?

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
Gerrit91: This change request is a bit complicated considering playing the roles on a hostgroup as described above. If we really want to execute individual host plays, we need to trigger them **after** the roles of the hostgroup were executed successfully. What about nested hostgroups anyway?

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
Gerrit91: If I do that it would say "Host group Ansible roles on host group", which is a bit redundant to me. Can we leave it with "Play Ansible roles on host group"?

app/overrides/hostgroup_play_roles.rb
Gerrit91: Ok, I will remove them, but it produces as a warning in the console: 

```
Deface: [WARNING] No :original defined for 'ansible_row_title', you should change its definition to include:
16:36:28 rails.1   |  |  :original => '016f2f86bd199275e7f3bebef77fefa90dd218f1'
```

I thought that it's useful if there was an alert upon the slightest change in the view because the override could ruin the layout.

app/controllers/api/v2/play_roles_controller.rb
dLobatog: @Gerrit91 You got it right, exactly that - adding the methods to the hosts/hostgroups controllers allow you to use `find_resource` and also the API paths will be similar to those of Puppet/Salt and other paths in Foreman.

app/controllers/foreman_ansible/concerns/hostgroups_controller_extensions.rb
dLobatog: By the regular GET action I mean /hostgroups/play_roles - it should be /api/v2/hostgroups/:id/play_roles

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
dLobatog: Oh good catch! Mind to fix them there too?

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
dLobatog: Understood - we should provide a way to choose strategies too, no need to do that in this PR. Let's keep this discussion for foreman-dev or further PRs. Keep it as is now for 1.4.0 😄

app/lib/actions/foreman_ansible/play_hostgroup_roles.rb
dLobatog: My bad - you're right!

app/overrides/hostgroup_play_roles.rb
dLobatog: Fair enough - honestly I normally remove them as they change so often, but if you'll be on the lookout for the alerts, that's a valid reason

dLobatog:
@kuffner-avid @Gerrit91 Hey thanks again for the PR! The changes requested are basically to move the actions to `api/v2/hosts/:id/play_roles` and `api/v2/hostgroups/:id/play_roles` - let me know what you think of it. If you think this can be finished soon, I'll release your latest fixes of ansible_connection plus this as 1.4.0, otherwise I'll take the latest fixes and put it as 1.3.3.

dLobatog:
[test]

Gerrit91:
@dLobatog Your change requests are great and I already prepared all your requested changes. Unfortunately, I still have some questions before I can commit my changes:

1. If I add the methods to the hostgroups_controller as an extensions, how to provide the lines for (I suppose) documentation?

    ```
    module Api
      module V2
        module HostsControllerExtensions
          include ForemanTasks::Triggers

          api :POST, '/hosts/:id/play_roles', N_('Plays Ansible roles on hosts')
          param :id, String, :required => true

          def play_roles
          ...
    ```

    (throws: undefined method `api' for Api::V2::HostsControllerExtensions:Module)

2. Can I prevent the before_action `find_resource` for `multiple_play_roles` in the API v2 hosts_controller, which is triggered by default for all new methods in the controller? 

    From API v2 hosts_controller:

      ```
          before_action :find_resource, :except => [:index, :create, :facts]
      ```

    Even though it's possible to use `find_resources` with multiple ids, it would be nicer if the `find_multiple` from the regular hosts_controller or something similar could be reused.

Gerrit91:
I decided to show you what I've done, maybe it's easier for you to answer the questions then. Thanks for your help already. :)

dLobatog:
[test]

app/controllers/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -0,0 +1,70 @@
+module Api
+  module V2
+    # Extends the hostgroups controller to support playing ansible roles
+    module HostgroupsControllerExtensions
+      include ForemanTasks::Triggers
+
+      # TODO: How to add this to API v2 controller?
```
dLobatog: You need to do 2 things on both of these controllers:

* Add `extend ActiveSupport::Concern` to the module
* Wrap the methods in `included do end`. See an example of this [here](https://github.com/Katello/katello/blob/master/app/controllers/katello/concerns/api/v2/hostgroups_controller_extensions.rb) or [here](https://github.com/Katello/katello/blob/master/app/controllers/katello/concerns/api/v2/hosts_controller_extensions.rb)

app/views/foreman_ansible/ansible_roles/_ansible_row_buttons.html.erb
```diff
@@ -0,0 +1,15 @@
+<td>
+    <%=
+      if hostgroup.all_ansible_roles.empty?
```
dLobatog: Can you add the button directly to the list of actions of the hostgroup? http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin contains some help

dLobatog:
@Gerrit91 I need to look at find_multiple again - the 1st issue is fairly easy to solve I think, check out the inline comment. Thanks for the quick turnaround! (I'll keep on testing this as I haven't really finished)

app/controllers/api/v2/hostgroups_controller_extensions.rb
dLobatog: Ah I forgot.. if you can move this to the `ForemanAnsible` module that'd be even better. They'd be `ForemanAnsible::Api::V2::HostgroupsController`. Anything similar works. 

The reason is that plugin extensions should have namespaces to avoid clashing with one another:

e.g: foreman_discovery defines HostgroupExtensions, foreman_ansible defines also HostgroupExtensions, now we have two engines with the exact same module name, which could cause problems.

Gerrit91:
I think, the last commit contains your suggestions. However, the way I added the button to the action_buttons in the hostgroup_controller is quite bad because it brutally influences changes in the view... do you know how to do this a better way? Is this even possible with Deface?

dLobatog:
[test]

dLobatog:
ok to test

chilicat:
Rebased on master

chilicat:
@dLobatog seems like the version on master is not correct

lib/foreman_ansible/version.rb

  VERSION = '1.3.0'.freeze

But in the RPM packing repository expects '1.3.1'

dLobatog:
ok to test

dLobatog:
@Gerrit91 @chilicat (same person)? Thanks for the update! I've been testing it and it pretty much works from what I see. Sorry for taking so long to get back to you, the queue of foreman-related tasks I have to do these days is getting long 😥 

Anyway - my thoughts on this: since this is working code, and you mentioned "this is thje first time I work with Ruby" I tend  to think we should merge it. I see the 'code GPA' is taking a 0.25 hit (17 new issues), none of them are really a big deal so I'll just fix them on a PR myself post-merge unless you want to try it out. The code structure and reutilization which is what mattered the most to me is quite in line with the rest of the project & takes advantage of Foreman core APIs.

I think you're somewhere in the US - so in case you want to fix some of the codeclimate issues, I'll wait until tomorrow morning to merge 😄 The plan goes like:

- merge this 
- fix any remaining codeclimate issues to go back to 4.0 code quality
- test thoroughly 
- add new feature to run ansible roles after a host is provisioned
- test thoroughly
- release !

Thanks again! This is quite a milestone.

dLobatog:
@Gerrit91 Test failures indicate there are no permissions for non-admin users to use this:  http://ci.theforeman.org/job/test_plugin_pull_request/3259/#showFailuresLink

Check `foreman_ansible.register_plugin` on `lib/foreman_ansible/engine.rb` to understand how to add them - in any case I'll merge this today but I'd appreciate if you save me that 😄 

Thanks!

Gerrit91:
@dLobatog We're not the same person, we are just in the same company. I am living in Germany by the way.

It's great to hear that you are gonna merge the changes in. I'm also a bit low on time today unfortunately, although I would like to dig in into the final optimizations. I saw that there was a problem with the permissions already and I did not really understand the concept of the permissions to be honest at this point of time. I only tested the changes using `rubocop` and `rake test:foreman_ansible`. In both cases, this issue was not addressed. Can you give me a hint how I can specifically test permissions because I suppose I cannot do that by guessing.

Gerrit91:
Thanks again! This should solve the problem with the permissions. :)

dLobatog:
ok to test [test]

dLobatog:
Thanks @Gerrit91 and @chilicat  !

I'll merge and solve the codeclimate issues on my own (given this is your first time with Ruby, it's a crazy good job already). I hope to make a release next week with this patch for the 3 latest versions, which now include 1.14, 1.13 and 1.12

chilicat:
thanks @dLobatog really appreciated.

Are you also update the version in:
https://github.com/theforeman/foreman_ansible/blob/c6aa3ecfd2d2855ad11a0308d3d479230600fcbd/lib/foreman_ansible/version.rb

I was building for our internal use the plugin and I noticed that in foreman_packaging the version is set to 1.3.1 (no 1.3.0).

dLobatog:
@chilicat Yes, I will update it when I release the new version 1.4.0. You can use the 1.3.1 tag of this gitrepo to get what you're looking for though.

Issue #68, opened 2016-12-26
Provide option to set verbosity output


alejojo:
I have tried to set up several levels of verbosity but is not possible. An option on "Settings > Ansible" would be quite helpful

chilicat:
+1

Gerrit91:
I think, this is a really good idea and I created a pull request for that.

dLobatog:
Closing as the PR was merged

PR #71, opened 2017-01-17
Fixes #18108 - Verbosity level for Ansible runs

Gerrit91:
In issue https://github.com/theforeman/foreman_ansible/issues/68, a setting for the level of verbosity for Ansible is proposed. I think this feature can be very useful as it helps a lot with debugging Ansible from within Foreman.


dLobatog:
ok to test 
[test]

@Gerrit91 Thanks - I'll merge if tests pass. It'd be nice to write tests for this feature though, I'll check out if I can carve some time to do this myself.

dLobatog:
@Gerrit91 This PR didn't work originally , I fixed it on https://github.com/theforeman/foreman_ansible/pull/73 so that we can send arbitrary options to ForemanAnsibleCore to pass verbosity & others

PR #69, opened 2017-01-17
Fixes #18104 - Ad hoc Ansible role execution

Gerrit91:
Especially when using Foreman as a backend service, there is yet no
possibility to assign roles to hosts over the API. In addition to that,
some scenarios (like maintenance tasks) require only one-time-execution
of a role or rare role execution from time to time. The need to change
role assignments for each host to do these kind of tasks is impractical.
Therefore, an API where there is no need of specific role assignments
and where an imported role can be executed directly against a certain
host helps to broaden the scopes of application of the Foreman Ansible
plugin. Hopefully, the community is also interested in such
functionality.

PR #73, opened 2017-01-19
Fixes #18150 - Ansible roles run after a host is provisioned

dLobatog:


Gerrit91:
@dLobatog Some things are not working for me with these changes. 

Firstly, when verbosity_level is set to 0, the return value of setup_verbosity is nil. The command that's going to executed results in

```
Running command [{"JSON_INVENTORY_FILE"=>"/tmp/.../foreman-inventories/..."}, "ansible-playbook", "-i", "/root/foreman_ansible/bin/json_inventory.sh", nil, "/tmp/.../foreman-playbook-....yml"]
```

Throwing an error in dynflow:

```
[foreman-tasks/dynflow] [E] error while dispatching request to runner ...:TypeError no implicit conversion of nil into String:
```

A possible fix is contained in #69.

Secondly, the playbooks cannot be found any longer. The tests succeed of course, but looking at the UI reveals: "StandardError: Script execution failed" due to the error with Ansible "ERROR! the playbook:  could not be found".

Gerrit91:
The second problem was happening due to my implementation in #69. Actually, the verbosity should not be inserted into the command array at all if the level is smaller than 1.

Gerrit91:
By the way, I really like how you solved that problem with the proxy solution! 

The only thing that I think is a little odd, is that in the task UI it says `Play Ansible roles on host <host_name> through not_defined` even though we now that Ansible is running on the localhost, which is **defined**. If there was no proxy found, I think it should just state `Play Ansible roles on host <host_name>`.

Gerrit91:
By the way and just for your personal information (yes, I heard your sprint review ☺️), theoretically it's possible to have an arbitrarily high verbosity level in modules, so -vvvv is not the maximum verbosity. It's also briefly mentioned [in this video](https://www.youtube.com/watch?v=YqaTx1s8i6k#t=17m19s).

iNecas:
Ad verbosity level: I've filed the issue for this, just for make it easier for others to find the reason of the problem, before we have a fix http://projects.theforeman.org/issues/18278

PR #74, opened 2017-01-27
Fixes #18278 - Validate verbosity levels

dLobatog:
Verbosity levels were defaulting to the wrong one (0, which caused an
error). Now it offers a sensible list of possible verbosities with the
description of what they are.


dLobatog:
I'll make a 1.4.2 release just for this, so we offer a proper default at least. Some people are reporting on mailing lists and IRC the '0' default option does not work at all.

dLobatog:
[test] ok to test

PR #75, opened 2017-01-31
fixes #18309 - handle run verbosity zero

thomasmckay:


dLobatog:
I don't think this condition should happen with https://github.com/theforeman/foreman_ansible/commit/0ac98d55deabe80e62a1977c355c181202b21d03 but I'll merge just in case. Thanks @thomasmckay !

PR #77, opened 2017-02-09
Fixes #18311 - Read /etc/ansible/ansible.cfg roles_path

dLobatog:
Currently it's reading the hardcoded /etc/ansible/roles directory
instead of the /etc/ansible/ansible.cfg roles_path, which makes it
confusing for users who have custom roles_paths.

The implementation for reading roles_path should go in
ForemanAnsibleCore so that the proxy and Foreman can reuse the same
gem for reading roles

### DIFF
--- a/lib/foreman_ansible_core/roles_reader.rb
+++ b/lib/foreman_ansible_core/roles_reader.rb
@@ -1,0 +1,45 @@
+module ForemanAnsibleCore
+  # Implements the logic needed to read the roles and associated information
+  class RolesReader
+    class << self
+      DEFAULT_CONFIG_FILE = '/etc/ansible/ansible.cfg'.freeze
+
+      def list_roles
+        logger.info('[foreman_ansible] - Reading roles from '\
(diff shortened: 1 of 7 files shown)


lib/foreman_ansible_core/roles_reader.rb
```diff
@@ -0,0 +1,36 @@
...
+        return default_path if roles_line.empty?
+        roles_path_key = roles_line.first.split('=').first.strip
+        # In case of commented roles_path key "#roles_path", return default
+        return default_path unless roles_path_key == 'roles_path'
+        # In case roles_path is there, and is not commented, return the value
+        roles_line.first.split('=').last.strip
+      end
+
+      def logger
+        Foreman::Logging.logger('foreman_ansible')
```
dLobatog: This does NOT log in the proxy properly, only in Foreman - which needs to be fixed.

dLobatog:
This is WoC as it needs tests and it requires some support for the proxy logger.

PR #85, opened 2017-04-11
Fixes #19240 - fix the inline help icon position

ares:
before
ansible_before

after
ansible_after

Note that this requires Foreman 1.15. I also added Gemfile.lock to .gitignore, I was lazy to open separate issue for that, let me know if you want me to separate it.


dLobatog:
Errors are unrelated - thanks @ares

PR #88, opened 2017-06-20
Fixes #20060 - correct apidoc for single hostgroup run roles

tbrisker:


dLobatog:
Thanks @tbrisker & @adamruzicka !

PR #94, opened 2017-07-25
fixes #20407 - adds two new options

mriedel-pgdx:
Fixes #20407 - adds two new options.
Adds ability to configure ansible_become & ansible_ssh_private_key_file options to foreman_ansible plugin. IDE also edited whitespace in ansible.rb file.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -34,7 +34,7 @@ class Setting
               N_('Foreman will use the sudo command to run roles on hosts '\
                  'You can override this on hosts by adding a parameter '\
                  '"ansible_become"'),
-              'true',
+              true,
               N_('Become')
             ),
             set(
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -42,7 +42,7 @@ class Setting
               N_('Use this password by default when running Ansible '\
                  'playbooks. You can override this on hosts '\
                  'by adding a parameter "ansible_ssh_pass"'),
-                 'ansible',
+              'ansible',
               N_('Password')
             ),
             set(
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -99,7 +99,7 @@ class Setting
             create(s.update(:category => 'Setting::Ansible'))
           end
         end
-
+        Setting::BLANK_ATTRS.push('ansible_ssh_private_key_file')
         true
       end
 
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -12,71 +12,86 @@ class Setting
         return unless super
         transaction do
           [
-            set(
-              'ansible_port',
-              N_('Use this port to connect to hosts '\
+              set(
+                  'ansible_port',
+                  N_('Use this port to connect to hosts '\
                  'and run Ansible. You can override this on hosts'\
                 ' by adding a parameter "ansible_port"'),
-              22,
-              N_('Port')
-            ),
-            set(
-              'ansible_user',
-              N_('Foreman will try to connect to hosts as this user by default'\
+                  22,
+                  N_('Port')
+              ),
+              set(
+                  'ansible_user',
+                  N_('Foreman will try to connect to hosts as this user by default'\
                  ' when running Ansible playbooks. You can override this '\
                  ' on hosts by adding a parameter "ansible_user"'),
-              'root',
-              N_('User')
-            ),
-            set(
-              'ansible_ssh_pass',
-              N_('Use this password by default when running Ansible '\
+                  'root',
+                  N_('User')
+              ),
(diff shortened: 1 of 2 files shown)


app/models/setting/ansible.rb
```diff
@@ -37,6 +46,15 @@ def load_defaults
               N_('Password')
             ),
             set(
+              'ansible_ssh_private_key_file',
+              N_('Use this to supply a path to an SSH Private Key '\
+                 'that Ansible will use in lieu of a password '\
+                 'Override with "ansible_ssh_private_key_file"'\
```
xprazak2: nit: needs a space after the closing `"`

app/models/setting/ansible.rb
```diff
@@ -15,20 +15,29 @@ def load_defaults
...
+                 '"ansible_user"'),
               'root',
               N_('User')
             ),
             set(
+              'ansible_become',
+              N_('Foreman will use the sudo command to run roles on hosts '\
+                 'You can override this on hosts by adding a parameter '\
+                 '"ansible_become"'),
+              'True',
```
xprazak2: try replacing `'True'` with `true`, our helper should recognize the boolean and use dropdropdown instead of text field

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -38,10 +38,18 @@ def connection_params(host)
...
         'ansible_user' => host_user(host),
-        'ansible_ssh_pass' => host_ssh_pass(host),
+        'ansible_become' => host_become(host),
         'ansible_connection' => connection_type(host),
         'ansible_winrm_server_cert_validation' => winrm_cert_validation(host)
       }
+      # Check if private_key is defined, if it is use that. Otherwise
+      # use ssh_pass
+      if !host.host_params['ansible_ssh_private_key_file'].empty? ||
+          !Setting[:private_key].empty?
```
xprazak2: I think this should be `Setting[:ansible_ssh_private_key_file]`. Note that condition will be true out of the box since you set the default value to `/tmp/private_key_file`, so I would recommend keeping it blank. You can also take advantage of your `host_private_key_file` method and change it to:

```ruby
unless host_private_key_file(host).empty?
# ...
else
# ...
end
```

xprazak2:
[test]

xprazak2:
Thank you for your patch, I left a few comments inline, but it looks very nice!

mriedel-pgdx:
Hi @xprazak2 ! Thanks for the feedback, and catching the 'private_key' I missed!

I've made the changes you suggested, except the "unless" because I'm using an "else" with it. My IDE raised a warning on that, and evidently there's some [controversy](http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/) over doing things that way.

Also, maybe this is a Foreman bug, but in my testing empty strings ('') as defaults caused the setting to disappear from the Ansible settings page in Foreman. The existing `Default verbosity level` setting does not appear in Foreman 1.15.2, and the same thing happened with `ansible_ssh_private_key_file`. Regardless, I've updated that too.

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -38,10 +38,18 @@ def connection_params(host)
...
         'ansible_user' => host_user(host),
-        'ansible_ssh_pass' => host_ssh_pass(host),
+        'ansible_become' => host_become(host),
         'ansible_connection' => connection_type(host),
         'ansible_winrm_server_cert_validation' => winrm_cert_validation(host)
       }
+      # Check if private_key is defined, if it is use that. Otherwise
+      # use ssh_pass
+      if !host.host_params['ansible_ssh_private_key_file'].empty? ||
+          Setting[:ansible_ssh_private_key_file].empty?
```
xprazak2: I think the `!` got lost, it was here before. You can keep `if`, I do not mind either way. My main point was that you can re-use the logic from `host_private_key_file` method and be more DRY:

```ruby
if !host_private_key_file(host).empty?
```

xprazak2:
I solved the missing setting mystery. We have to add a setting with a blank default into whitelist. If you add:

```ruby
Setting::BLANK_ATTRS.push('ansible_ssh_private_key_file')
```

the setting will turn up. Here is an [example](https://github.com/theforeman/foreman/blob/develop/app/models/setting/provisioning.rb#L15) how we use it in other places.

mriedel-pgdx:
@dLobatog Thanks! I merged the changes in and recommitted.

dLobatog:
Thanks @mriedel-pgdx! Note you can use any arbitrary ansible CLI option using the latest Foreman Ansible version (only available in git now, I will release a package next week probably). 

Basically any `ansible_` param will be passed on to the playbook so it can be used as a connection option, similar to what you did here, except other parameters won't have a global setting.

PR #95, opened 2017-07-31
Fixes #20464 - Allow arbitrary connection options via host parameters

dLobatog:


app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -81,6 +81,12 @@ def host_ssh_pass(host)
       host.host_params['ansible_ssh_pass'] || Setting[:ansible_ssh_pass]
     end
 
+    def ansible_extra_options(host)
+      host.host_params.select do |key, _|
+        /ansible_/.match(key) || Setting[key]
```
dLobatog: This method should be better tested to make sure it handles ALL current settings and allows us to remove the `host_` methods.

PR #98, opened 2017-09-09
Fixes #20884 - Allow importing roles from multiple paths

bastilian:
This allows importing from multiple paths, when 'roles_path' has multiple colon separated paths.

### DIFF
--- a/lib/foreman_ansible_core/roles_reader.rb
+++ b/lib/foreman_ansible_core/roles_reader.rb
@@ -3,31 +3,18 @@ module ForemanAnsibleCore
   class RolesReader
     class << self
       DEFAULT_CONFIG_FILE = '/etc/ansible/ansible.cfg'.freeze
+      DEFAULT_ROLES_PATH = '/etc/ansible/roles'.freeze
 
       def list_roles
-        logger.info('[foreman_ansible] - Reading roles from '\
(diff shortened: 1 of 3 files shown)


lib/foreman_ansible_core/roles_reader.rb
```diff
@@ -40,6 +27,32 @@ def logger
...
+                            " from #{roles_path} - #{e.message}"
+        raise ReadConfigFileException.new(exception_message)
+      end
+
+      def roles_path_from_config
+        File.readlines(DEFAULT_CONFIG_FILE).select do |line|
+          line =~ /roles_path/
+        end
+      rescue Errno::ENOENT, Errno::EACCES => e
+        logger.debug("[foreman_ansible] - #{e.backtrace}")
```
iNecas: `logger.error(e)` would be better, both from backtrace formating, as well for error level

iNecas: oh, I see your passing different exception, is n this case, the original approach is probably ok

iNecas:
👍 from me. Nice @bastilian!

bastilian:
@iNecas Thanks for reviewing! Those few lines didn't feel right, I've cleaned it up a bit. Mind taking another look? :)

dLobatog:
Thanks @bastilian !

PR #99, opened 2017-09-27
Fixes #21008 - Improve button to run ansible roles

bastilian:
The button to run roles is now without an icon, but proper wording instead:
gnome-shell-screenshot-k1oz6y


dLobatog:
Thanks @bastilian !

PR #101, opened 2017-10-05
Fixes #21005 - Better wording for roles import button

bastilian:
Changes the wording of the roles import button to the following wording when multiple sources for roles are available:

gnome-shell-screenshot-wunr7y

When only one is available it stays as is:

gnome-shell-screenshot-tdsm7y
 
_A change of `select_action_button` in core to allow those to be primary (blue) buttons is needed before this is ready_


bastilian:
Once https://github.com/theforeman/foreman/pull/4892 is merged the button will appear as primary:

gnome-shell-screenshot-ehto7y

dLobatog:
Thanks @bastilian !

PR #104, opened 2017-11-01
Fixes #21528 - start using factory bot

ares:

PR #105, opened 2017-11-01
Fixes rubocop

ares:
I added .rubocop_todo for things I don't want to fix just during rubocop upgrade, this should unblock master tests and #104


dLobatog:
Thanks @ares !

PR #107, opened 2017-11-13
Fixes #21003 - Improve table for ansible roles

bastilian:
Before:
gnome-shell-screenshot-7sit9y

After:
gnome-shell-screenshot-p3za9y

### DIFF
--- a/app/controllers/ansible_roles_controller.rb
+++ b/app/controllers/ansible_roles_controller.rb
@@ -5,6 +5,7 @@ class AnsibleRolesController < ::ApplicationController
   before_action :find_resource, :only => [:destroy]
   before_action :find_proxy, :only => [:import]
   before_action :create_importer, :only => [:import, :confirm_import]
+  before_action :default_order, :only => [:index]
 
   def index
     @ansible_roles = resource_base.search_for(params[:search],
@@ -38,6 +39,10 @@ class AnsibleRolesController < ::ApplicationController
   end
 
   private
+
+  def default_order
+    params[:order] ||= 'name ASC'
+  end
 
   def find_proxy
     return nil unless params[:proxy]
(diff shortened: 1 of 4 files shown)


app/controllers/ansible_roles_controller.rb
```diff
@@ -5,10 +5,11 @@ class AnsibleRolesController < ::ApplicationController
   before_action :find_resource, :only => [:destroy]
   before_action :find_proxy, :only => [:import]
   before_action :create_importer, :only => [:import, :confirm_import]
+  before_action :default_order, :only => [:index]
 
   def index
     @ansible_roles = resource_base.search_for(params[:search],
-                                              :order => params[:order]).
+                                              :order => ()).
```
dLobatog: Shouldn't this still be `params[:order]` in order to use `default_order`?

ares:
Could we also have columns containing number/date more thin? values there have fixed size, the name though can be quite long. Could you please add ordering to the rest of columns? At lesat imported_at if counts are too complicated.

app/controllers/ansible_roles_controller.rb
bastilian: Yes, it should. fixed.

bastilian:
@ares & @dLobatog:

* Adjusted the width of all colums
* Ordering by "Imported at" is now possible
* Added scoped_search for ansible_roles to `Host` to allow searching for `role =`
* Hosts count is linked to roles search of hosts

gnome-shell-screenshot-7mzo9y

dLobatog:
@bastilian I think https://github.com/theforeman/foreman_ansible/commit/62c71aa1ecbff4757584d8800d75b34ec0f4e943 was causing one test to fail. If so, we should see http://ci.theforeman.org/job/test_plugin_foreman_ansible_master/ green again (at least I do locally). If it was something else, I'll merge this & work on the error after.

dLobatog:
Thanks @bastilian !

PR #111, opened 2017-12-25
Remove alias_method_chain from hosts helper

tbrisker:


dLobatog:
Merged as part of https://github.com/theforeman/foreman_ansible/pull/112

PR #108, opened 2017-11-30
Fixes #21818 - Add Rails 5 style migrations

xprazak2:
To be merged after core moves to Rails 5.1

PR #112, opened 2017-12-25
Foreman 1.17 updates

mmoll:


mmoll:
includes GH-108 and GH-111

PR #113, opened 2017-12-27
Fixes #22018 - PlayHostRoles uses REX provider 

dLobatog:
Based on #18 - other tasks should follow suit.

### DIFF
--- a/app/controllers/foreman_ansible/concerns/hosts_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/concerns/hosts_controller_extensions.rb
@@ -23,8 +23,9 @@ module ForemanAnsible
 
       def play_roles
         find_resource
-        task = async_task(::Actions::ForemanAnsible::PlayHostRoles, @host)
-        redirect_to task
+        composer = job_composer(:ansible_run_host, @host)
+        composer.trigger
+        redirect_to job_invocation_path(composer.job_invocation)
       rescue Foreman::Exception => e
         error e.message
         redirect_to host_path(@host)
@@ -37,6 +38,17 @@ module ForemanAnsible
       rescue Foreman::Exception => e
         error e.message
         redirect_to hosts_path
+      end
+
+      private
+
+      def job_composer(feature_name, target)
+        composer = ::JobInvocationComposer.for_feature(feature_name, target)
+        return composer if composer.save
+        raise ::Foreman::Exception.new(
+          format(N_('Could not run Ansible roles for %{host}'),
+                 :host => @host)
+        )
       end
     end
   end
--- a/app/views/foreman_ansible/job_templates/ansible_roles.erb
+++ b/app/views/foreman_ansible/job_templates/ansible_roles.erb
@@ -1,0 +1,17 @@
+<%#
+kind: job_template
+name: Ansible Roles - Ansible Default
+job_category: Ansible Playbook
+description_format: 'Run Ansible roles for host'
+feature: ansible_run_host
+provider_type: Ansible
+%>
+
+---
+- hosts: <%= @host.name %>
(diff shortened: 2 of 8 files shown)


app/controllers/foreman_ansible/concerns/hosts_controller_extensions.rb
```diff
@@ -38,6 +43,11 @@ def multiple_play_roles
         error e.message
         redirect_to hosts_path
       end
+
+      #
+      # params[:feature]
+      # params[:host_ids]
```
ares: I guess this does not belong here :-)

app/controllers/foreman_ansible/concerns/hosts_controller_extensions.rb
```diff
@@ -23,8 +23,13 @@ def action_permission
...
       def play_roles
         find_resource
-        task = async_task(::Actions::ForemanAnsible::PlayHostRoles, @host)
-        redirect_to task
+        @composer = JobInvocationComposer.for_feature(:ansible_run_host, @host)
+        if @composer.save
+          @composer.trigger
+          redirect_to job_invocation_path(@composer.job_invocation)
+        else
+          raise ::Foreman::Exception.new(N_("Could not run Ansible roles for %s") % @host)
```
ares: IMHO we shouldn't raise an exception here, most likely there are validation problems which is why composer was not saved, since we rescue from this exception on line below, it seems we're using exceptions for controlling the flow. Compose should not raise Foreman::Exceptions, so I think we should just add the redirection to else branch. If there's other exception, it's probably code error and should case 500.

app/views/foreman_ansible/job_templates/ansible_roles.erb
```diff
@@ -0,0 +1,15 @@
...
+feature: ansible_run_host
+provider_type: Ansible
+%>
+
+---
+- hosts: <%= @host.name %>
+  tasks:
+    - debug: msg=These are the parameters "{{ foreman_params }}"
+  roles:
+    - <%= @host.all_ansible_roles.map { |role| role.name.strip }.join("\n  - ") %>
```
adamruzicka: This generates the roles wrongly if `@host.all_ansible_roles` is empty. I think it should generate an empty list so applying no roles would be successful. The way it is now it fails with

```
ERROR! Invalid role definition: None
Exit status: 4
```

adamruzicka:
I went through all the realted PRs, left some comments inline, but in general I really like the way this goes.

app/views/foreman_ansible/job_templates/ansible_roles.erb
dLobatog: I will change it, however notice that the "Run Ansible roles" button does NOT show up if a host has no Ansible roles

adamruzicka: That's possible, but there are other ways of triggering the run which don't have such safeguard.

dLobatog:
Some integration tests seem to be failing on postgres all the time (for 8 months now). Locally I'm not able to reproduce, to unblock development I think I will merge if sqlite/mysql pass.

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4548

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4560

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4561

adamruzicka:
tests running here http://ci.theforeman.org/job/test_plugin_matrix/4565/

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4566 tests running here now

PR #114, opened 2017-12-28
 Fixes #22019 - PlayHostsRoles using REX provider 

dLobatog:


ares:
In which PR do you want review comments? I commented in the last one, since the work is related I think we could work in single PR with multiple commits. But whatever you find easier for development...

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4574 after fixing a bunch of problems with fixtures

dLobatog:
A reproducer:

```
bundle exec ruby -I.:test -e "ARGV.each{|f| require f}" test/integration/host_js_test.rb  test/unit/plugin_test.rb
```

dLobatog:
Passing all tests now with https://github.com/theforeman/foreman/pull/5218 😆 

https://gist.github.com/dLobatog/a1423c423dd355a236909131e06360f1 (max 65536 chars on GitHub comments so I can't post it here 🤣 )

ares:
There's a plugin dsl for disabling incompatible core tests IIRC

dLobatog:
http://ci.theforeman.org/job/test_plugin_matrix/4609/ ought to be fine.

ares:
seems like there's some error which might be related
```
10:59:27 E, [2018-01-29T10:59:27.064111 #21277] ERROR -- /parallel-executor-core: cannot accept event: Dynflow::Director::Event[execution_plan_id: adc57198-9aeb-495b-b0e8-c65e75076027, step_id: 2, event: Dynflow::Action::WithPollingSubPlans::Poll, result: <#Concurrent::Edge::CompletableFuture:0x98cfcb8 pending>] core is terminating (Dynflow::Error)
```

dLobatog:
@ares I've seen this on the REX repo too, only on sqlite, I'll investigate.

Seems that other databases are working fine http://ci.theforeman.org/job/test_plugin_matrix/4609/database=mysql,ruby=2.4,slave=fast/consoleFull

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -5,25 +5,19 @@ module V2
...
             find_resource
-
-            @result = {
-              :hostgroup => @hostgroup, :foreman_tasks => async_task(
-                ::Actions::ForemanAnsible::PlayHostgroupRoles, @hostgroup
-              )
-            }
-
-            render_message @result
+            composer = job_composer(:ansible_run_host, @hostgroup.hosts)
```
ares: `@hostgroup.hosts` only links direct assignments ignoring the hostgroup inheritance, perhaps this is what we want but wanted to raise it here

lib/foreman_ansible/register.rb
```diff
@@ -0,0 +1,52 @@
...
+Foreman::Plugin.register :foreman_ansible do
+  requires_foreman '>= 1.16'
+
+  security_block :foreman_ansible do
+    permission :play_roles_on_host,
+               { :hosts => [:play_roles, :multiple_play_roles],
+                 :'api/v2/hosts' => [:play_roles,
+                                     :multiple_play_roles] },
+               :resource_type => 'Host'
+    permission :play_roles_on_hostgroup,
```
ares: do we need separate permission for this if at the end it's just playing roles on separate hosts? I think just "play_roles" should be enough. Or we could just rely on `create_job_invocation` which is what REX already does. I see this is not the new code so it's fair to keep it out of this PR.

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
dLobatog: Good point, will address on the Hostgroup PR - I think it makes more sense to run the whole tree

PR #115, opened 2017-12-28
 Fixes #22020 - Host group play roles using REX provider

dLobatog:


app/controllers/foreman_ansible/concerns/hosts_controller_extensions.rb
```diff
@@ -42,7 +42,7 @@ def multiple_play_roles
           @composer.trigger
           redirect_to job_invocation_path(@composer.job_invocation)
         else
-          #raise ::Foreman::Exception.new(N_("Could not run Ansible roles for %s") % @host)
+          raise ::Foreman::Exception.new(N_("Could not run Ansible roles %s") % @composer.errors)
```
ares: this is quite specific exception, should we create new child class of Foreman::Exception for it? something like AnsibleRolePlayError

app/controllers/foreman_ansible/concerns/hostgroups_controller_extensions.rb
```diff
@@ -18,6 +19,13 @@ def play_roles
 
       private
 
+      def check_hostgroup
+        return unless @hostgroup.hosts.empty?
```
ares: this is another place to be changed if we decide to run ansible on hosts assigned to children hostgroups as well but good for now

ares:
looking at last version, I'm not sure failures from http://ci.theforeman.org/job/test_plugin_matrix/4613/ were fixed, did you trigger another jenkins run? otherwise LGTM once we have green run on postgres/mysql

ares:
@dLobatog both seems to be merged now

dLobatog:
Yup, all green but sqlite (rex problem) - ci.theforeman.org/job/test_plugin_matrix/4630/ - merging

PR #116, opened 2018-01-03
Fixes #22125 - Restore correct output for ansible reports

bastilian:
This brings back correct rendering of reports imported from Ansible:

Before:
gnome-shell-screenshot-ce2acz

After:
gnome-shell-screenshot-hajtbz

This is a start to improve the view for reports, an idea to further improve it would be to add a json viewer (like [react-json-view](https://www.npmjs.com/package/react-json-view)) for both arguments and message.

Another idea to make it more useful would be to limit what is shown by default, while still allowing to view "raw" versions of the report logs.


ares:
Is there a way to change the data we get to STDOUT? While this is definitely an improvement, I don't think users would like to parse JSONs visually even using react-json-view, JSON is more for machine processing which we don't do here.

bastilian:
@ares since this relies on Ansible reporting to Foreman (rather than Foreman running roles), we would need to change the callback in Ansible to report with STDOUT, maybe in addition to JSON, as it provides more detailed information.

ares:
Ok then, I guess we should apply the same formatting to arguments column.

Could we also parse out _ansible_parsed keys? That seems like a noise.

For values that are hashes,  as a poor man solution for now, could we just  replace it with three dots and leave the value as hover bubble? Or using inline tooltip we have in forms. At least until we'd start receiving stdout from the hook. Long term, the hook might be replaced by the runner itself.

db/migrate/20180111130853_add_config_reports_origin.rb
```diff
@@ -0,0 +1,5 @@
+class AddConfigReportsOrigin < ActiveRecord::Migration[5.1]
+  def change
```
ares: this should be probably in core

app/models/concerns/foreman_ansible/config_report_extensions.rb
```diff
@@ -0,0 +1,35 @@
+module ForemanAnsible
+  module ConfigReportExtensions
+    extend ActiveSupport::Concern
+
+    included do
+      class << self
+        alias_method :orig_import, :import
+
+        def import(report, proxy_id = nil)
```
ares: I think the logic should be in code and plugins should only provide ways how to identify report origin

bastilian:
@ares Agreed on both. Currently implementing report scanner in core, that can be registered on the ReportImporter and get a report to set in this case the origin. the ConfigReport extensions are then gone and foreman-ansible will register a scanner.

bastilian:
@ares The JSON message is now properly split up and module args and invocations with module arguments only appear in the module arguments column and are cleared from the results column.

The "_ansible_XYZ" keys are cleared from out too, it is still very long, adding collapsing of certain keys should make it more compact, which should be enough for now. It would be great though to have input from UX team to make it more useful. 

Here is a screenshot of the current state:

reports

ares:
Ok, looks slightly better, few suggestions:
1) vertical alignment should be top if possible, otherwise to find the message I need to scroll to the middle of arguments
2) could we add max height to table cells? I think for facts we used to have expand button for too long texts, perhaps the same could be used here? See following (hope that could be used for multiline input, so we could eg limit it to 5 lines)

a
b

ares:
if 2 does not work well, at least hiding nested hashes should be done, the output length now makes is really hard to use

bastilian:
Adding collapsing is still a good short term solution, but it would be great to have reports show only important and useful information and not have the user parse the data in the long run.

lib/foreman_ansible/engine.rb
```diff
@@ -102,11 +105,6 @@ class Engine < ::Rails::Engine
       end
     end
     # rubocop:enable BlockLength
-
-    rake_tasks do
-      Rake::Task['db:seed'].enhance do
```
dLobatog: Why is this removed?

bastilian: Oh, that must have been a bad merge.

bastilian:
@dLobatog I'm working on making it compatible with 1.17 and 1.18 atm.

bastilian:
@dLobatog rubocop offences are fixed now. Only `remove_keys` makes codeclimate fail, any idea how that could be rewritten to be less complex? ...I don't see how it could...

dLobatog:
@bastilian No worries about it, I tried a bit on https://github.com/theforeman/foreman_ansible/pull/129 too, it's fine as long as tests pass. We can come back to it any time but ideally it'd be nice to release this soon 😄

ares:
@bastilian see https://github.com/bastilian/foreman_ansible/pull/1, this could work before we add https://www.npmjs.com/package/react-json-view unless you want to add it as part of this PR

ares:
also I played with the JS a bit and it would be quite simple to add truncation like we have it for facts, see https://github.com/theforeman/foreman_ansible/pull/116#issuecomment-363963106

we can't reuse the logic from core as it's too specific for facts DOM but it would be like 10 lines to get the same feature here

any of three solutions (scroll bars, json view, truncate) would work with 1.18 only, but that's OK

bastilian:
@dLobatog The `module_name` method is now slightly improved and only returns a module name if it occurs with a colon.

Just to verify, we agreed on merging this now as is (when green) and have separate PR based on this with being able to expand and collapse (via react-json-view)?

ares:
The only change I'd like to incorporate from my https://github.com/bastilian/foreman_ansible/pull/1 is new keys for ignoring - https://github.com/bastilian/foreman_ansible/pull/1/commits/de8171f9f35d0618f010b735f977acae94a60280#diff-d39af4594f5f160f09c00c40d3536e86 

we don't need to print stdout and stderr, it's already stored in REX output

dLobatog:
@ares Not all reports come from REX output, in fact many current users ask on IRC about how to use the callback with the *ansible/ansible-playbook CLI* so IMO they shouldn't be removed.

ares:
still, I don't think storing stdout/stderr is that useful if it's mixed in this JSON and takes most of the space but I'm OK with keeping it for now, folding should fix it later
for facts, I guess that's clear, if users have callback setup, they most likely also have facts parsed

dLobatog:
Thanks @bastilian @ares!

PR #117, opened 2018-01-08
 Refs #13812 - support connect by IP REX feature

ares:
Based on WIP rex provider

### DIFF
--- a/app/services/foreman_ansible/inventory_creator.rb
+++ b/app/services/foreman_ansible/inventory_creator.rb
@@ -16,13 +16,13 @@ module ForemanAnsible
     # more advanced cases). Therefore we have only the 'all' group
     # with all hosts.
     def to_hash
-      { 'all' => { 'hosts' => hosts.map(&:fqdn) },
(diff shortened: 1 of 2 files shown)


app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -16,13 +16,13 @@ def initialize(hosts)
     # more advanced cases). Therefore we have only the 'all' group
     # with all hosts.
     def to_hash
-      { 'all' => { 'hosts' => hosts.map(&:fqdn) },
+      { 'all' => { 'hosts' => hosts.map { |h| AnsibleProvider.find_ip_or_hostname(h) } },
```
adamruzicka: Considering this already depends on REX, couldn't you just use `SSHExecutionProvider.find_ip_or_hostname(h)` instead of copying the method out to `AnsibleProvider`? Or just making `AnsibleProvider` delagate the methods to `SSHExecutionProvider`?

ares: yes, right after https://github.com/theforeman/foreman_remote_execution/pull/303 gets in, I'm happy to use it

PR #118, opened 2018-01-18
Fixes #22126 - Always show 'Play Roles' button for Hostgroups

bastilian:
This will always show the "Play Roles" button as an action for hostgroups listed.
If no roles are assigned a tooltip is now giving that hint:
gnome-shell-screenshot-fpwedz

If a role is assigned to either the hostgroup itself or any of the hosts assigned, there will be no tooltip and the link can be used to run the roles. 

Currently only roles assigned directly to the hostgroup are used for the run, this behavior needs to be changed, but is likely a bigger change.

### DIFF
--- a/app/models/concerns/foreman_ansible/has_many_ansible_roles.rb
+++ b/app/models/concerns/foreman_ansible/has_many_ansible_roles.rb
@@ -6,7 +6,9 @@ module ForemanAnsible
 
     included do
       def all_ansible_roles
-        (ansible_roles + inherited_ansible_roles).uniq
+        result = (ansible_roles + inherited_ansible_roles).uniq
+        result += host_ansible_roles if is_a? Hostgroup
+        result
       end
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/concerns/foreman_ansible/has_many_ansible_roles.rb
```diff
@@ -6,7 +6,9 @@ module HasManyAnsibleRoles
 
     included do
       def all_ansible_roles
-        (ansible_roles + inherited_ansible_roles).uniq
+        result = (ansible_roles + inherited_ansible_roles).uniq
+        result += host_ansible_roles if is_a? Hostgroup
```
dLobatog: How about just overriding this method in the hostgroup? Also I think you need to call uniq here to avoid duplicates

```
def all_ansible_roles
  (super += host_ansible_roles).uniq
end
```

bastilian:
@dLobatog the `all_ansible_roles` is now overwritten in the hostgroups extensions.
I tried your way, but that gave me a syntax error for some reason.

bastilian:
@dLobatog I'm aware of the `TEST` var that is how these failures were missed. fixing it now.

bastilian:
@dLobatog I reverted back to checking if it's a `HostGroup` in the has many extensions, calling `super` in a module method won't call a method included via a module. (of course 🤦‍♂️) 

`rake test:foreman_ansible` ran green locally hopefully now in Jenkins too.

PR #119, opened 2018-01-23
Fixes #22378 - Insights remediation playbook macro 

dLobatog:
To be merged after #115 #114 #113

### DIFF
--- a/app/services/foreman_ansible/insights_plan_runner.rb
+++ b/app/services/foreman_ansible/insights_plan_runner.rb
@@ -1,0 +1,67 @@
+module ForemanAnsible
+  # Fetch information about a plan from RH Insights and run it
+  class InsightsPlanRunner
+    include RedhatAccess::Telemetry::LookUps
+
+    def initialize(organization, plan_id)
+      @organization = organization
+      @plan_id = plan_id
+    end
+
+    def run_playbook
+      rules = playbook
+      hostname_rules_relation = hostname_rules(rules)
+      hosts = hostname_rules_relation.keys.map do |hostname|
+        Host::Managed.find_by(:name => hostname)
+      end
+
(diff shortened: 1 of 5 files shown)


app/controllers/api/v2/insights_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      include ::Api::Version2
+      include RedhatAccess::Telemetry::LookUps
+
+      api :POST,
+        '/ansible/insights/:id/run',
+        N_('Run playbook to fix all systems in a Red Hat Access Insights plan')
+      param :id, :identifier, :required => true
+      def run
+        # Fetch playbook by querying
+        binding.pry
```
adamruzicka: This should probably go away

app/services/foreman_ansible/insights_plan_runner.rb
```diff
@@ -0,0 +1,88 @@
...
+      composer.save
+      composer.trigger
+    end
+
+    def playbook
+      resource = RestClient::Resource.new(
+        'https://api.access.redhat.com/r/insights/'\
+        "v3/maintenance/#{@plan_id}/playbook",
+        # "https://api.access.redhat.com/r/
+        # insights/v3/maintenance/34285/playbook",
```
adamruzicka: This should probably go away as well

app/services/foreman_ansible/insights_plan_runner.rb
```diff
@@ -0,0 +1,88 @@
...
+    def rules_to_hash(rules)
+      result = {}
+      rules.map do |rule|
+        rule['hosts'] = 'all'
+        result[rule['name']] = rule
+      end
+      result
+    end
+
+    # Irrelevant method now
```
adamruzicka: So do we need to keep it around?

PR #123, opened 2018-01-30
Fixes #22124 - Add report interval setting for Ansible

bastilian:
This adds the setting (`Setting[:ansible_interval]`) specifically for Ansible host reporting to allow a longer period to report back to Foreman until they are considered "out of sync".

Also see https://github.com/theforeman/foreman/pull/5221

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -72,6 +72,12 @@ class Setting
                  'foreman_params["host_parameter"] in the playbooks.'),
               true,
               N_('Top level Ansible variables')
+            ),
+	    set(
+              'ansible_interval',
+              N_('Timeout (in minutes) when hosts should have reported.'),
+              '30',
+              N_('Ansible report timeout')
             )
           ].compact.each do |s|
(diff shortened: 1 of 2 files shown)


app/models/setting/ansible.rb
```diff
@@ -106,6 +106,12 @@ def load_defaults
                  'foreman_params["host_parameter"] in the playbooks.'),
               true,
               N_('Top level Ansible variables')
+            ),
+	    set(
+              'ansible_interval',
+              N_('Timeout (in minutes) when hosts should have reported.'),
```
dLobatog: "0 if you do not require hosts to report via Ansible on a fixed interval", also 0 by default

bastilian:
@dLobatog The changes for this that are needed in core are actually in the ["origin_interval" PR](https://github.com/theforeman/foreman/pull/5221)

We will address the not reporting in via another option. @ares suggested the "reports_off" option.

ares:
you should probably link the commit to http://projects.theforeman.org/issues/22124 at some point

dLobatog:
Thanks @bastilian

PR #124, opened 2018-02-03
Fixes #22507 - Pass REX inputs as inventory variables

dLobatog:
To use inputs in JobTemplates, users would have to use ERB <%=
input('myinput') %> so that the input is actually parsed by Foreman.
Since Ansible users (most likely!) don't want to migrate their Ansible
playbooks which contain Ansible variables to ERB.
Ansible variables are substituted by the variables coming from the
inventory. Therefore, it's possible to simply just send the
resolved inputs to Ansible through the inventory, and Ansible will
substitute the 'variables'.


ares:
Nice and simple :-)

adamruzicka:
Works like a charm for me 👍

dLobatog:
Thanks @adamruzicka @ares !

PR #125, opened 2018-02-05
Fixes #22512 - Pass variables without foreman_params

dLobatog:


app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -35,8 +38,8 @@ def host_vars(host)
       {
         'foreman' => host_attributes(host),
         'foreman_params' => host_params(host),
-        'foreman_ansible_roles' => host_roles(host)
-      }.merge(connection_params(host))
+        'foreman_ansible_roles' => host_roles(host),
+      }.merge(connection_params(host)).merge(host_params(host))
     end
```
dLobatog: This line  is the key part about this PR

adamruzicka:
Works as expected with one slight catch. Ansible doesn't allow variable names to contain dashes, so any parameter with a dash in its name won't be usable from the top level. This might be obvious for Ansible people, but it caught me off guard.

### Playbook:
```yaml
---
- hosts: all
  tasks:
    - debug:
        var: thing
    - debug:
        var: foreman_params
    - debug:
        var: snake_case_thing
    - debug:
        msg: "This is |{{ foreman_params['kebab-case-thing'] }}|"
    - debug:
        var: kebab-case-thing
```

### Result:
image

dLobatog:
Thanks @adamruzicka !

PR #133, opened 2018-02-20
Fixes #22626 - hide stdout and stderr

ares:
note that we still display stdout_lines and stderr_lines which are properly formatted, since they are arrays, stdout and stderr are just long strings that makes the page extremely wide


dLobatog:
Merged as 5b4630f8c83264ce94edc647d892d0abb6710dd1, thanks @ares !

PR #135, opened 2018-02-20
Fixes #22634 - Reuse remote execution settings

dLobatog:
effective_user, remote_execution_ssh_passphrase,
remote_execution_ssh_user, remote_execution_ssh_port should substitute
the role that 'ansible_become', 'ansible_ssh_pass', 'ansible_user',
'ansible_port' had in previous versions. The variables should still be
overridable at the host params level, for example:
host_params['ansible_user'] overrides the 'remote_execution_ssh_user'
setting.

cc @ares @adamruzicka

### DIFF
--- a/app/services/foreman_ansible/inventory_creator.rb
+++ b/app/services/foreman_ansible/inventory_creator.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'securerandom'
 module ForemanAnsible
   # Service to list an inventory to be passed to the ansible-playbook binary
@@ -47,10 +49,16 @@ module ForemanAnsible
     end
 
     def connection_params(host)
-      params = ansible_settings.merge ansible_extra_options(host)
-      # Backward compatibility for Ansible 1.x
-      params['ansible_ssh_port'] = params['ansible_port']
-      params['ansible_ssh_user'] = params['ansible_user']
+      # Preference order is:
+      # 1st option: host parameters.
+      #   - If they're set to 'ansible_whatever' we use that over anything else
+      # 2nd option: REX options.
+      #   - both settings, ssh password, effective_user can be used
+      # 3rd option:
+      #   - other settings
+      params = ansible_settings.
+               merge(remote_execution_options(host)).
+               merge(ansible_extra_options(host))
       params
     end
 
@@ -68,8 +76,7 @@ module ForemanAnsible
 
     def ansible_settings
       Hash[
-        %w[port user ssh_pass connection
-           ssh_private_key_file become
+        %w[connection ssh_private_key_file
            winrm_server_cert_validation].map do |setting|
           ["ansible_#{setting}", Setting["ansible_#{setting}"]]
         end
@@ -82,6 +89,19 @@ module ForemanAnsible
       end
     end
 
+    def remote_execution_options(host)
+      params = {
+        'ansible_become' => @template_invocation.effective_user,
+        'ansible_user' => host_setting(host, 'remote_execution_ssh_user'),
+        'ansible_ssh_pass' => rex_ssh_password(host),
+        'ansible_port' => host_setting(host, 'remote_execution_ssh_port')
+      }
+      # Backward compatibility for Ansible 1.x
+      params['ansible_ssh_port'] = params['ansible_port']
+      params['ansible_ssh_user'] = params['ansible_user']
+      params
+    end
+
     def template_inputs(template_invocation)
       input_values = template_invocation.input_values
       result = input_values.each_with_object({}) do |input, vars_hash|
@@ -90,10 +110,19 @@ module ForemanAnsible
       result
     end
 
+    def rex_ssh_password(host)
+      @template_invocation.job_invocation.password ||
+        host_setting(host, 'remote_execution_ssh_password')
+    end
+
     private
 
     def render_rabl(host, template)
       Rabl.render(host, template, :format => 'hash')
+    end
+
+    def host_setting(host, setting)
+      host.params[setting.to_s] || Setting[setting]
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -82,6 +89,20 @@ def ansible_extra_options(host)
       end
     end
 
+    def remote_execution_options(host)
+      params = {
+        'ansible_become' => @template_invocation.effective_user,
+        'ansible_user' => host_setting(host, 'remote_execution_ssh_user'),
+        'ansible_ssh_pass' => host_setting(host,
+                                           'remote_execution_ssh_passphrase'),
```
adamruzicka: I believe this should be `remote_execution_ssh_password`

test/unit/services/inventory_creator_test.rb
```diff
@@ -16,31 +20,46 @@ class InventoryCreatorTest < ActiveSupport::TestCase
...
       Setting.expects(:[]).with('ansible_ssh_private_key_file').
         returns(nil).at_least_once
-      Setting.expects(:[]).with('ansible_port').returns(nil).at_least_once
-      Setting.expects(:[]).with('ansible_user').returns(nil).at_least_once
-      Setting.expects(:[]).with('ansible_ssh_pass').
+      Setting.expects(:[]).with('remote_execution_ssh_port').
+        returns(2222).at_least_once
+      Setting.expects(:[]).with('remote_execution_ssh_user').
+        returns('root').at_least_once
+      Setting.expects(:[]).with('remote_execution_ssh_passphrase').
```
adamruzicka: Ditto

test/unit/services/inventory_creator_test.rb
```diff
@@ -16,31 +20,46 @@ class InventoryCreatorTest < ActiveSupport::TestCase
...
-      assert_equal Setting['ansible_ssh_pass'],
+      refute_equal Setting['remote_execution_ssh_user'],
+                   connection_params['ansible_user']
+      assert_equal extra_options['ansible_user'],
+                   connection_params['ansible_user']
+      refute_equal Setting['remote_execution_ssh_port'],
+                   connection_params['ansible_port']
+      assert_equal extra_options['ansible_port'],
+                   connection_params['ansible_port']
+      assert_equal Setting['remote_execution_ssh_passphrase'],
```
adamruzicka: Ditto

dLobatog:
Thanks @adamruzicka for the review!

ares:
with quick tests it seems that I only can set the password via remote_execution_ssh_password host param but when I use the invocation form field, it does not work, should this work with this PR?

dLobatog:
@ares I think that should work to close the bug, so I will keep it open to update it with that.

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -90,10 +110,19 @@ def template_inputs(template_invocation)
       result
     end
 
+    def rex_ssh_password(host)
+      @template_invocation.job_invocation.password ||
```
dLobatog: Here's what I had to change to make the job invocation password available

test/unit/services/inventory_creator_test.rb
```diff
@@ -16,36 +23,60 @@ class InventoryCreatorTest < ActiveSupport::TestCase
...
+        returns('root').at_least_once
+      Setting.expects(:[]).with('remote_execution_ssh_password').
         returns('asafepassword').at_least_once
       Setting.expects(:[]).with('ansible_winrm_server_cert_validation').
         returns(true).at_least_once
       Setting.expects(:[]).with('ansible_connection').
         returns('ssh').at_least_once
       @host.expects(:host_params).returns(extra_options).at_least_once
-      inventory = ForemanAnsible::InventoryCreator.new(@host, nil)
+      @template_invocation.job_invocation.expects(:password).
```
dLobatog: This test tests that the REX setting is respected

test/unit/services/inventory_creator_test.rb
```diff
@@ -16,36 +23,60 @@ class InventoryCreatorTest < ActiveSupport::TestCase
...
+      assert_equal extra_options['ansible_port'],
+                   connection_params['ansible_port']
+      assert_equal Setting['remote_execution_ssh_password'],
                    connection_params['ansible_ssh_pass']
       assert_equal Setting['ansible_winrm_server_cert_validation'],
                    connection_params['ansible_winrm_server_cert_validation']
     end
 
+    test 'job invocation ssh password is passed when available' do
+      inventory = ForemanAnsible::InventoryCreator.new(@host,
```
dLobatog: This test tests that the job invocation password takes priority

adamruzicka:
> how are you unencrypting this in REX?

We use the `Encryptable` module[1] from Foreman core which should allow automatic encryption and decrpytion of model's attributes as they're accessed. I'm not sure why it's not getting decrypted for you.

[1] - https://github.com/theforeman/foreman/blob/develop/app/models/concerns/encryptable.rb

ares:
> for the SSH private key passphrase

In that case we need to update the inline help for the field in rex that this field is only applicable for SSH backend. I'd also recommend adding inline help to passphrase and explain how/whether the password is stored and when it's cleared.

dLobatog:
@adamruzicka I can unencrypt it here, it's no big deal - I already send the ssh_pass from host params or REX settings unencrypted.  

@ares Yeah, not only that but also a comment asking to use `ssh-agent` instead - I don't know why it was necessary to do it in REX

dLobatog:
@adamruzicka Actually, no need to unencrypt it here. I put a breakpoint and I see the params are sent properly - `{"ansible_become"=>nil, "ansible_user"=>"root", "ansible_ssh_pass"=>"asdasdsadsadassadsadsa", "ansible_port"=>"22"}`.

dLobatog:
[test default] [test] will merge if tests pass

PR #139, opened 2018-02-23
Fixes #22677 Use proper api host name to allow TLS mutual auth

lphiri:
API requests must be made against "cert-api.access.redhat.com" and not "api.access.redhat.com".
The former is configured to accept RHSM manifest certificates for authentication, which is the expected production authentication method, with mutual tls verification enabled.
Reading the value of the API from a config file also allows this plugin to share configuration with the Insights plugin.

### DIFF
--- a/app/services/foreman_ansible/insights_plan_runner.rb
+++ b/app/services/foreman_ansible/insights_plan_runner.rb
@@ -27,10 +27,8 @@ module ForemanAnsible
     # Fetches the playbook from the Red Hat Insights API
     def playbook
       resource = RestClient::Resource.new(
-        'https://api.access.redhat.com/r/insights/'\
-        "v3/maintenance/#{@plan_id}/playbook",
-        get_ssl_options_for_org(@organization, nil).
-        merge(:verify_ssl => OpenSSL::SSL::VERIFY_NONE)
+        "#{insights_api_host}/r/insights/v3/maintenance/#{@plan_id}/playbook",
+        get_ssl_options_for_org(@organization, nil)
       )
       response = resource.get
       YAML.safe_load(response.body)


app/services/foreman_ansible/insights_plan_runner.rb
```diff
@@ -27,10 +27,8 @@ def run_playbook
     # Fetches the playbook from the Red Hat Insights API
     def playbook
       resource = RestClient::Resource.new(
-        'https://api.access.redhat.com/r/insights/'\
-        "v3/maintenance/#{@plan_id}/playbook",
-        get_ssl_options_for_org(@organization, nil).
-        merge(:verify_ssl => OpenSSL::SSL::VERIFY_NONE)
+        "#{insights_api_host}/r/insights/v3/maintenance/#{@plan_id}/playbook",
```
dLobatog: @lphiri Where is the `insights_api_host` variable coming from?

lphiri: @dLobatog  its a method from the "RedhatAccess::Telemetry::LookUps" module - the same place you are getting " get_ssl_options_for_org".

dLobatog:
[test]

lphiri:
@dLobato The rubocop error (2:1) is nowhere near the code changes in this PR. Am I missing something?
I don't get any errors when I run ` rake foreman_ansible:rubocop`:
`[vagrant@centos7-dev-latest foreman]$ rake foreman_ansible:rubocop
/home/vagrant/foreman/lib/foreman.rb:8: warning: already initialized constant Foreman::UUID_REGEXP
/home/vagrant/foreman/lib/foreman.rb:8: warning: previous definition of UUID_REGEXP was here
Running RuboCop...
Inspecting 83 files
...................................................................................

83 files inspected, no offenses detected
`
I think the Firefox error is a script loading order issue (will not be an issue in production). I am looking into it.

PR #142, opened 2018-03-05
Fixes #22774 -  ssh_private_key_path setting does not show up

dLobatog:
The setting to set the SSH key path does not show up by default. This
makes foreman_ansible set an option the inventory
'ansible_ssh_private_key_path: nil' which makes the path always default
to `/usr/share/foreman-proxy/.ssh/id_rsa`. This is problematic because
the key setup by REX by default will be at
`/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy`, and besides, the
user has no way of configuring this.

This fix should make it into 1.4.7 and the next 2.0 version of Ansible.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -10,6 +10,7 @@ class Setting
       # rubocop:disable BlockLength
       def load_defaults
         return unless super
+        Setting::BLANK_ATTRS.push('ansible_ssh_private_key_file')
         transaction do
           [
             set(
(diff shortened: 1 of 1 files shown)


ares:
Makes sense and fixes the problem for me, :+1:

app/models/setting/ansible.rb
```diff
@@ -10,6 +10,7 @@ class << self
       # rubocop:disable BlockLength
       def load_defaults
         return unless super
+        Setting::BLANK_ATTRS.push('ansible_ssh_private_key_file')
```
dLobatog: The issue here was that it tried to create the setting as '' by default. The validation only allows this when the setting name is in `BLANK_ATTRS`. Since `BLANK_ATTRS` was only updated after trying to create the setting, instead of **before**, it didn't work.

ares: yup, I figured out, reviewed and tested yesterday, you've got my ACK if that helps merging :-)

PR #143, opened 2018-03-05
Fixes #22775 - add missing audits for user actions

ares:


app/models/hostgroup_ansible_role.rb
```diff
@@ -1,5 +1,6 @@
 # Join model that hosts the connection between hostgroups and ansible_roles
 class HostgroupAnsibleRole < ApplicationRecord
+  audited :associated_with => :hostgroup, :allow_mass_assignment => true
```
dLobatog: Apparently this needs some change to log the name of the role too

ares: the problem is in Foreman core or better to say inability to extend `AuditsHelper::MAIN_OBJECTS` from plugin, Kavita is working on it right now. We can merge as it is now so we start getting data and later we just change the way how it's presented. Or we can wait few days until the core DSL is available, though I'd prefer the first approach.

ares:
yeah, reproduced the same, I'll take a closer look, though the fix probably needs to land in core, data seems to be stored correctly

ares:
Fixed, this now depends on https://github.com/theforeman/foreman/pull/5322 which is much cleaner way to implement it.

ares:
core PR got in :-)

ares:
core PR was meged, this can be too now

dLobatog:
Thanks @ares !

PR #151, opened 2018-03-22
Fixes #22994 - Add refactor for ansible roles controller

ofedoren:


ares:
While I don't have strong preference on keeping prefix but since this is a breaking change of API, I wonder if it would be possible to keep old routs too and log a deprecation in case they are used.

ofedoren:
@ares I apologize for confusing caused by PR's name. Actually I'm changing the API documentation only, because it's obviously wrong (I figured it out just by using the generated documentation from API controller)

ares:
Thanks for explaining, now it makes sense. I tested it, rerendered apipie cache which now looks correct. :+1: for merging.

PR #154, opened 2018-03-28
Fixes #23048 - Proxy param from hash to identifier

ofedoren:


ares:
This in fact changes the existing API, but given the fact it was wrongly documented as POST action and it actually makes it more consistent with the rest of Foreman API, I'd say let's merge this without keeping also `proxy` parameter (which was even documented as hash).

PR #159, opened 2018-04-10
Fixes #23205 - Prefix job template categories with "Ansible"

bastilian:
and update from community templates

### DIFF
--- a/db/migrate/20180410125416_rename_ansible_job_categories.rb
+++ b/db/migrate/20180410125416_rename_ansible_job_categories.rb
@@ -1,0 +1,17 @@
+class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
+  def up
+    User.as_anonymous_admin do
+      updated_templates = ['Power Action - Ansible Default',
+                           'Puppet Run Once - Ansible Default']
+      JobTemplate.where(:name => updated_templates).all.each do |job_template|
+        next if job_template.job_category =~ /^Ansible/
+        job_template.job_category = "Ansible #{job_template.job_category}"
+        job_template.save!
+      end
+
+      service_template = JobTemplate.where(:name => 'Service Action - Ansible Default').first
+      service_template.job_category = 'Ansible Services'
+      service_template.save!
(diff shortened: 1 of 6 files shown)


db/migrate/20180410125416_rename_ansible_job_categories.rb
```diff
@@ -0,0 +1,11 @@
+class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
+  def change
+    User.as_anonymous_admin do
+      JobTemplate.search_for('name ~ "Ansible Default"').all.each do |job_template|
```
ares: I think it would be better to be specific and update only templates we seed. If users cloned them, it's up to them to update the category. Also I think the same is required for  "job_category: Ansible Services" change.

db/migrate/20180410125416_rename_ansible_job_categories.rb
```diff
@@ -0,0 +1,11 @@
+class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
+  def change
```
ares: I think this is up only migration

bastilian:
@ares updated the migration to be more specific and only touch those that need to be updated and added an update for the Services change as well.

db/migrate/20180410125416_rename_ansible_job_categories.rb
```diff
@@ -0,0 +1,17 @@
+class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
+  def up
+    User.as_anonymous_admin do
+      updated_templates = ['Power Action - Ansible Default',
+                           'Puppet Run Once - Ansible Default']
```
dLobatog: Looks like the `Service Action` template is also going to be updated

ares: that's handled by https://github.com/theforeman/foreman_ansible/pull/159/files#diff-fe9a9a343fbfce6ef20705c597e95427R13

PR #161, opened 2018-04-11
Fixes #23227 - Add endpoints for ansible roles list commands

ofedoren:


lib/foreman_ansible/register.rb
```diff
@@ -13,6 +13,14 @@
                  :'api/v2/hostgroups' => [:play_roles,
                                           :multiple_play_roles] },
                :resource_type => 'Hostgroup'
+    permission :show_roles_on_host,
```
ares: we usually don't introduce new permission for subset of host/hostgroup attributes or relations, I think just checking `view_hosts` and `view_hostgroups` should be enough

ofedoren:
@ares , @dLobatog , updated.

ofedoren:
@dLobatog updated.

ares:
My comment is not fully addressed, it seems the permission name was changed from show_roles_on_host to view_roles_on_host but I think we shouldn't introduce new permission, but rather use existing view_hosts and view_hostgroups that lives in core. I think it's ok to see host's roles if user has view_hosts permissions

ofedoren:
@ares I apologize for misunderstanding. Is it correct now?

config/routes.rb
```diff
@@ -1,5 +1,28 @@
 # rubocop:disable BlockLength
 Rails.application.routes.draw do
+  scope '/' do
```
bastilian: There is already an `:api` namespace (at the bottom). Was there any reason not to put it there?

If possible it would be nice to have it in there instead of creating a duplicate namespace.

test/functional/api/v2/hostgroups_controller_test.rb
```diff
@@ -38,6 +40,13 @@ class HostgroupsControllerTest < ActionController::TestCase
         response = JSON.parse(@response.body)
         assert_job_invocation_is_ok(response, target.map(&:id))
       end
+
+      test 'should list ansible roles for a host' do
+        @host3.hostgroup.ansible_roles = [@ansible_role1]
+        get :ansible_roles, :params => { :id => @host3.hostgroup.id }
+        response = JSON.parse(@response.body)
+        refute_empty response
```
bastilian: An assertion that the response includes the roles name would be better.

`refute_empty` would also pass when the response is an empty JSON Object/Array.

test/functional/api/v2/hosts_controller_test.rb
```diff
@@ -62,6 +63,13 @@ class HostsControllerTest < ActionController::TestCase
         assert_response :success
         assert assigns('host').ansible_roles, [@ansible_role2]
       end
+
+      test 'should list ansible roles for a host' do
+        @host3.ansible_roles = [@ansible_role1]
+        get :ansible_roles, :params => { :id => @host3.id }
+        response = JSON.parse(@response.body)
+        refute_empty response
```
bastilian: Same as in the hostgroups controller test.

test/functional/api/v2/hostgroups_controller_test.rb
bastilian: > `refute_empty` would also pass when the response is an empty JSON Object/Array.
That's not right, but it would be good to confirm that the right role is returned.

ofedoren:
Don't merge this for now, please. I guess there is another solution via https://github.com/theforeman/hammer-cli-foreman/pull/342.

bastilian:
@ofedoren What is that status of this PR?

ofedoren:
@bastilian I apologize for not answering sooner.

I rebased and slightly changed the PR. Now I guess it's ready to be merged if everybody is happy with everything.

FYI, you mentioned this PR in another one with similar additions, which is closed now. This one does not fully replace the closed one, but I'll add the missing features in a split PR.

bastilian:
@ofedoren I tried reviewing these PRs, mind adding notes to them, like which one should go in first and which are relying on them. 
It would also be an option to merge these all into one PR (with/out separate commits).

dLobatog:
Closing as per https://github.com/theforeman/foreman_ansible/pull/205#

PR #162, opened 2018-04-11
Fixes #23228 - Add scoped_search on relations

ofedoren:


ares:
Works like a charm, :+1:

PR #164, opened 2018-04-12
Fixes #23243 - Display disclaimer on Insights playbook

dLobatog:
Red Hat Access Insights comes with some comments on the playbook.
However, the way `insights_plan_runner.rb` works is by directly parsing
the YAML coming from the site. As all YAML parsers ignore comments, the
disclaimer and other comments are not shown.

This commit takes the disclaimer and puts it on top of every template
invocation. Other comments are still ignored.

This needs tests, and a bit more testing :smile:


app/services/foreman_ansible/renderer_methods.rb
```diff
@@ -9,13 +9,21 @@ def insights_remediation(plan_id, organization_id = Organization.current.id)
...
-                       reduce([]) do |acc, cur|
+      host_playbooks = individual_host_playbooks(hostname_rules_relation,
+                                                 global_rules)
+      "#{disclaimer}\n#{host_playbooks.to_yaml}"
+    end
+
+    private
+
+    def individual_host_playbooks(hostname_rules_relation, global_rules)
+      hostname_rules_relation[@host.name].
```
ares: rubocop makes this really hard to read, 80 chars limitation is really (not introduced by this commit, just randomly ranting)

dLobatog: That was just me, not enforced by rubocop. I'll update it & add tests

ares: oh, ok :-)

ares:
What an amount of code for one assertion :-) let's wait for green jenkins, but :+1: for now. I'll merge if it's green.

dLobatog:
@ares Hmm... the code won't even be able to run the test because `InsightsPlanRunner` is only defined when the redhat_access plugin is installed too. I'll keep the test under `defined? RedhatAccess` like the class, if anything for downstream/local testing... I tested locally and it passes but obviously I had the plugin.

test/unit/services/insights_plan_runner_test.rb
```diff
@@ -0,0 +1,36 @@
+require 'test_plugin_helper'
+
+if defined? RedhatAccess
```
ares: perhaps you could just add `skip unless defined?(RedhatAccess)` to the test itself, but this also works fine

ares:
Yeah, that makes sense. The Job was aborted, so let's see again [test]

dLobatog:
@ares Unfortunately it could be that plugin tests take >2h now.. which is why tests got aborted in this repo - still if you check the results they're usually fine

ares:
Retested, works great. Thanks @dLobatog, merging (as previous test was successful and just one combination timeouted)

PR #165, opened 2018-04-12
Fixes #23245 - Rubocop issues in master

dLobatog:
There are a couple of tests that are not passing too, which ought to be fixed here so that this comes back to green.

PR #166, opened 2018-04-12
Refs #23205 - Fix renaming migration for empty and/or unseeded DBs

bastilian:
This will prevent the migration from failing when the database is unseeded and there are no job templates that need to be updated.

### DIFF
--- a/db/migrate/20180410125416_rename_ansible_job_categories.rb
+++ b/db/migrate/20180410125416_rename_ansible_job_categories.rb
@@ -1,5 +1,9 @@
 class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
   def up
+    unless User.unscoped.find_by_login(User::ANONYMOUS_ADMIN)
+      puts "No ANONYMOUS_ADMIN found. Skipping renaming Ansbible jobs"
+      return
+    end
     User.as_anonymous_admin do
       updated_templates = ['Power Action - Ansible Default',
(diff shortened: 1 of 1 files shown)


dLobatog:
If tests are able to run the migration in http://ci.theforeman.org/job/test_plugin_pull_request/4886 I'll merge - it won't run much more as this needs to be fixed https://github.com/theforeman/foreman_ansible/pull/165

db/migrate/20180410125416_rename_ansible_job_categories.rb
```diff
@@ -1,5 +1,9 @@
 class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
   def up
+    unless User.unscoped.find_by_login(User::ANONYMOUS_ADMIN)
+      puts "No ANONYMOUS_ADMIN found. Skipping renaming Ansbible jobs"
```
dLobatog: I'll fix this on merge 'Ansible' - :smile:

dLobatog:
Tests were able to run the migration, merging

dLobatog:
Merged as 136e599c4eb32be910aa969b4c478dca2c4f2fbe

db/migrate/20180410125416_rename_ansible_job_categories.rb
bastilian: Thanks! One of the more creative typos. :D

PR #167, opened 2018-04-17
Refs #23205 - fix the migration on existing setups

ares:
The migration fails because `without_auditing` is not a named scope but actually required a block in which auditing is ignored.

### DIFF
--- a/db/migrate/20180410125416_rename_ansible_job_categories.rb
+++ b/db/migrate/20180410125416_rename_ansible_job_categories.rb
@@ -7,21 +7,23 @@ class RenameAnsibleJobCategories < ActiveRecord::Migration[5.1]
     User.as_anonymous_admin do
       updated_templates = ['Power Action - Ansible Default',
                            'Puppet Run Once - Ansible Default']
-      job_templates = JobTemplate.without_auditing.where(
-        :name => updated_templates
-      ).all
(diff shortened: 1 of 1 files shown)


db/migrate/20180410125416_rename_ansible_job_categories.rb
```diff
@@ -7,21 +7,23 @@ def up
...
-        service_template.job_category = 'Ansible Services'
-        service_template.save_without_auditing
+      JobTemplate.without_auditing do
+        job_templates = JobTemplate.where(
+          :name => updated_templates
+        ).all
+        job_templates.each do |job_template|
+          next if job_template.job_category =~ /^Ansible/
+          job_template.job_category = "Ansible #{job_template.job_category}"
+          job_template.save_without_auditing
```
tbrisker: if you disable auditing using block you don't need to `save_without_auditing` here and below

dLobatog: If you don't mind I think I'll merge as it's harmless and some people are already complaining about not being able to migrate

ares: sorry I was busy, I opened https://github.com/theforeman/foreman_ansible/pull/168 to address that

PR #168, opened 2018-04-19
Refs #23205 - remove redundant without auditing

ares:
follow up refactoring


dLobatog:
[test default]

bastilian:
[test default]

PR #171, opened 2018-05-10
Pass string to webpacked_assets

dLobatog:
The reports page is not loading the Foreman webpack chunk because it's a
symbol. It needs to be a string to be loaded properly


ares:
Thanks @dLobatog, would you mind attaching redmine issue so that @theforeman-bot is happy?

ares:
This also seems to be fixed in core - https://github.com/theforeman/foreman/pull/5552

PR #172, opened 2018-05-20
Fixes #23653 - Do not rename interfaces names

mmalchuk:
It is antipattern to change non-local variable passed as a parameter.
This commit bring the same behaviour used in the puppet_fact_parser
and also adds the same debug logging.

Signed-off-by: Maksim Malchuk <maksim.malchuk@gmail.com>


bastilian:
@mmalchuk mind adding a test that describes what the method should be getting and what it should return. The changes in here don't seem to change the output at all, the interface name is still renamed via `.tr`.

mmalchuk:
@dLobatog here is some logs:
`2018-05-20T10:46:17 88de1713 [app] [D] We have following interfaces 'br-private, enp2s0f0, enp2s0f1, bond1.1005, br-ipo, br-ipo-nic, enp130s0f0, enp130s0f1, enp129s0f0, enp129s0f1, bond0, bond1' based on facts
2018-05-20T10:46:17 88de1713 [app] [D] Saving br_private NIC for host cloud-lab15.example.org
2018-05-20T10:46:18 88de1713 [app] [W] Saving br_private NIC for host cloud-lab15.example.org failed, skipping because:
2018-05-20T10:46:18 88de1713 [app] [W]  Identifier has already been taken`
as you can see the first line reports 'br-private' but the second 'br_private'
I can write the tests, but this issue appears here: https://github.com/theforeman/foreman/blob/1.17-stable/app/services/fact_parser.rb#L71
So the test written here.
Let's merge as is cos this is the obvious antipattern.

bastilian:
@mmalchuk The transformation of `-` to `_` is needed in order for other methods of `FactParser` to find the interface information without requiring to transform again when looking for interface information. It happens with a `tr!` to modify the result that is assigned to the `@interfaces` variable, which is within the realm of a `FactParser` object.

The transformation is needed to correctly retrieve interface information from Ansible facts, but it should not change the actual fact value written to DB to have the interface device name not transformed. A test in [fact_parser_test.rb](https://github.com/theforeman/foreman_ansible/blob/master/test/unit/services/fact_parser_test.rb) would help to verify this.

lzap:
The parser code in core evolved into this terrible spaghetti and it needs to be redesigned. There must be clear contract defined for these parsers and we should avoid using ActiveRecord because this leads to various problems with associations and reload hacks.

Unfortunately this is like thirty man-days of work, amount of tests is huge in this area. Not sure when this happens.

lzap:
Sorry it took so long, I am willing to take it from here. What we need:

* rebase core and this PR
* create a test in this plugin to check the modified method (a test case with virtual NIC)
* that test must pass with the core PR

We must not change method parameters, that's really not what anyone want to see.

app/services/foreman_ansible/fact_parser.rb
```diff
@@ -51,15 +51,17 @@ def get_interfaces # rubocop:disable Naming/AccessorMethodName
...
+      interface = iface_name.tr('-', '_') # virbr1-nic -> virbr1_nic
       interface_facts = facts[:"ansible_#{interface}"]
       ipaddress = ip_from_interface(interface)
       ipaddress6 = ipv6_from_interface(interface)
-      HashWithIndifferentAccess[
+      iface_facts = HashWithIndifferentAccess[
         interface_facts.merge(:ipaddress => ipaddress,
                               :ipaddress6 => ipaddress6)
       ]
+      logger.debug { "Interface #{interface} facts: #{iface_facts.inspect}" }
```
lzap:
```suggestion
      logger.debug { "Ansible interface #{interface} facts: #{iface_facts.inspect}" }
```

mmalchuk: done

mmalchuk:
@lzap review and merge please.

ekohl:
ok to test

lzap:
Tests do not look good, anyone have an idea what is wrong? Probably a rebase would help.

ares:
Tests failures are caused by REX (dependency) which should have been resolved but don't have the release yet. If this works correctly, we can merge.

mmalchuk:
> Thanks for the fix @mmalchuk and sorry it took so long to merge.

Thank you. We still use 1.77 in production, its great if these changes would be packaged.

ekohl:
We have dropped all code to release to 1.20 and older. Should still be possible to do a manual build using mock or even patch locally but I'd strongly urge to upgrade to a supported version.

mmalchuk:
> We have dropped all code to release to 1.20 and older. Should still be possible to do a manual build using mock or even patch locally but I'd strongly urge to upgrade to a supported version.

will try 1.20. thanks!

PR #173, opened 2018-05-20
Fixes #23653 - Do not rename interfaces names

mmalchuk:
It is antipattern to change non-local variable passed as a parameter.
This commit bring the same behaviour used in the puppet_fact_parser
and also adds the same debug logging.

Signed-off-by: Maksim Malchuk <maksim.malchuk@gmail.com>
(cherry picked from commit f780f2ec6dce45d4130e002909e8b1a30c03ef68)

### DIFF
--- a/app/services/foreman_ansible/fact_parser.rb
+++ b/app/services/foreman_ansible/fact_parser.rb
@@ -54,11 +54,14 @@ module ForemanAnsible
       end
     end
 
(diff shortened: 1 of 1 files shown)


bastilian:
Usually cherry picks to stable branches should happen after they are merged into master/develop and mostly only if fixes need to be backported.

PR #174, opened 2018-05-28
Fixes #23715 - Verify OS facts before submitting them

dLobatog:
Before this PR, some roles would collect only a few facts from the
machine, but not 'os_major' 'os_minor' 'os_name', so these facts were
empty. The entire POST request would fail with a 500 in that case, as we
were submitting Operatingsystem.create! without any facts to create it.

### DIFF
--- a/app/services/foreman_ansible/operating_system_parser.rb
+++ b/app/services/foreman_ansible/operating_system_parser.rb
@@ -1,0 +1,73 @@
+module ForemanAnsible
+  # Methods to parse facts related to the OS
+  module OperatingSystemParser
+    def operatingsystem
+      args = { :name => os_name, :major => os_major, :minor => os_minor }
+      return @local_os if local_os(args).present?
+      return @new_os if new_os(args).present?
+      logger.debug do
+        'Ansible facts parser: No OS could be created with '\
+        "os_name='#{os_name}' os_major='#{os_major}' "\
+        "os_minor='#{os_minor}': "\
+        "#{@new_os.errors if @new_os.present?}"
+      end
+      nil
+    end
+
+    def local_os(args)
+      @local_os = Operatingsystem.where(args).first
+    end
+
+    def new_os(args)
+      return @new_os if @new_os.present?
+      @new_os = Operatingsystem.new(args.merge(:description => os_description))
+      @new_os if @new_os.valid? && @new_os.save
+    end
+
+    def os_name
+      facts[:ansible_distribution] ||
+        facts[:ansible_lsb] && facts[:ansible_lsb]['id']
+    end
+
+    def debian_os_major_sid
+      case facts[:ansible_distribution_major_version]
+      when /wheezy/i
+        '7'
+      when /jessie/i
+        '8'
+      when /stretch/i
+        '9'
+      when /buster/i
+        '10'
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


ares:
> @ares I don't think we should be syncing the OS if it wasn't found. 

I agree but I think we should add a log message in that case as many users expect that OS will be updated after any ansible run. At least debug log level.

app/services/foreman_ansible/operating_system_parser.rb
```diff
@@ -0,0 +1,69 @@
...
+  module OperatingSystemParser
+    def operatingsystem
+      args = { :name => os_name, :major => os_major, :minor => os_minor }
+      local_os = Operatingsystem.where(args).first
+      return local_os if local_os.present?
+      return @new_os if new_os(args).present?
+      logger.debug do
+        'Ansible facts parser: No OS could be created with '\
+        "os_name='#{os_name}' os_major='#{os_major}' "\
+        "os_minor='#{os_minor}'"
```
ares: I think it would also be great to log @new_os.errors unless @new_os.nil?

lovilak:
Hello,
I'm testing this fix on centos 7.5 foreman-1.17.1-1 tfm-rubygem-foreman_ansible_core-2.0.2-1.fm1_17.el7 and I've no operatingsystem information for microsoft os. 
Regards

lovilak:
I meant, that foreman do not detect the OS in OperatingSystem Column even with facts like 
ansible_distribution_major_version 6 and ansible_os_family Windows. But maybe its not the purpose of this patch.

puzzle-it-nu:
Just ran into this issue, probably, but patching my current, tfm-rubygem-foreman_ansible-2.0.1-1.fm1_17.el7.noarch, still gives the same error more or less.
Running ansible windows -m setup -i hosts.test gives a 500 Server Error: Internal server Error at the end.
I see the following error in foreman/production.log:

> 2018-06-12T13:37:34 932de692 [app] [D] Ansible facts parser: No OS could be created with os_name='Microsoft Windows Server 2016 Standard 
Evaluation' os_major='10' os_minor='0': #<ActiveModel::Errors:0x00007faf2d7fc568>

Facts from the run which are output to the console:

>        "ansible_os_family": "Windows", 
>        "ansible_os_name": "Microsoft Windows Server 2016 Standard Evaluation", 
>        "ansible_owner_contact": "", 
>        "ansible_owner_name": "Windows User", 

Trying to create an OperatingSystem through the webui containing a space is not allowed either.

A temporary fix might be to do a .tr(" ","_") on os_name. I tried that but my Ruby skills aren't good enough it seems.

dLobatog:
Thanks @puzzle-it-nu - I will write a test for it in another PR and mention you there. We needed that kind of feedback to be able to make it work. When names contain spaces, we could just put that as the OS description (in Foreman the OS description can have spaces, not the name), and the name of the OS can be just `MicrosoftWindowsServer`. It's the OS description what's shown on the UI, so everywhere in Foreman you would see `Microsoft Windows Server 2016 Standard Evaluation`

dLobatog:
@dgoetz @lovilak @puzzle-it-nu https://github.com/theforeman/foreman_ansible/pull/178 hopefully contain what you needed

ares:
Thanks @dLobatog, works fine. Merging!

PR #175, opened 2018-05-31
Change Plugin Manual link to direct to parent page

tek0011:
After spending a lot of time setting up ansible/foreman, many times I clicked on the plugin documentation.  However, it links to the original 1x version.  This could confuse new users.  Made change to documentation to direct users to parent page of the document, so users can see what the latest version is and choose it.


ares:
Makes sense, thanks! Merging.

PR #176, opened 2018-06-05
Fixes #23731 - Add host key to known_hosts if not found

dLobatog:
This ensures we are *always* able to run Ansible on remote hosts,
without compromising security too much (for example, disabling host key
verification).

1st: We look into all known hosts_files for the hosts in the inventory.
     If it's found, we just use it to prevent MITM attacks

2nd: If the host key isn't found, we fetch it with ssh-keyscan, and then
     add it to all known_hosts files.

### DIFF
--- a/foreman_ansible_core.gemspec
+++ b/foreman_ansible_core.gemspec
@@ -20,4 +20,5 @@ DESC
   s.add_development_dependency 'rubocop', '~> 0.52'
   s.add_dependency 'foreman-tasks-core', '~> 0.1'
   s.add_dependency 'foreman_remote_execution_core', '~> 1.1'
+  s.add_dependency 'net-ssh', '~> 4.0'
 end
--- a/lib/foreman_ansible_core/playbook_runner.rb
+++ b/lib/foreman_ansible_core/playbook_runner.rb
@@ -1,6 +1,7 @@
 require 'foreman_tasks_core/runner/command_runner'
 require_relative 'command_creator'
 require 'tmpdir'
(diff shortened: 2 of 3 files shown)


ekohl:
Your commit message is still the old one and doesn't reflect the current implementation.

ekohl:
@dLobatog https://github.com/theforeman/foreman/pull/5659

ares:
[test foreman_ansible] the core issue has been resolved

foreman_ansible_core.gemspec
```diff
@@ -20,4 +20,5 @@ DESC
   s.add_development_dependency 'rubocop', '~> 0.52'
   s.add_dependency 'foreman-tasks-core', '~> 0.1'
   s.add_dependency 'foreman_remote_execution_core', '~> 1.1'
+  s.add_dependency 'net-ssh', '~> 4.0'
```
ares: is the version pinned to 4.x because 5 requires Ruby >= 2.2.6?

dLobatog: Not really, I just looked at what we had in foreman-packaging (4.0.1) and tried to make it compatible

ares: I tested with 5.0.1 and all seems fine, perhaps >= 4.0 and < 6 would be fine?

ares:
Thanks @dLobatog, that was the case. Now it works just great! I think this would deserve mentioning in plugin manual. Merging! We'll need to bump net-ssh when the time comes :-)

PR #177, opened 2018-06-12
Fixes #23742 - share all templates to all taxonomies

ares:


db/seeds.d/75_job_templates.rb
```diff
@@ -1,3 +1,5 @@
+organizations = Organization.unscoped.all
+locations = Location.unscoped.all
```
bastilian: Would it make sense to allow passing in organisation and location IDs and/or even allow this assignment to be optional?

ares:
FTR, the error seems unrelated - `Sequel::DatabaseError: PG::DuplicateTable: ERROR:  relation "dynflow_execution_plans" already exists`

dLobatog:
Fixes for the failing tests can be found at https://github.com/theforeman/foreman/pull/5690/files

PR #179, opened 2018-06-14
Fixes #23545 - Move post-provisioning Ansible to REX

dLobatog:
Leftovers from when foreman_ansible was not a REX provider. It now uses
the :ansible_run_host REX feature to run the playbook with all roles.


app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -22,14 +23,13 @@ def inherited_ansible_roles
...
-          ::ForemanAnsible::ProxySelector.new,
-          :timeout => Setting['ansible_post_provision_timeout']
-        )
+        composer = job_composer(:ansible_run_host, self)
+        composer.trigger!
         logger.info("Task for Ansible roles on #{self} before_provision: "\
-                    "#{Rails.application.routes.url_helpers.task_path(task)}.")
+                    "#{job_invocation_path(composer.job_invocation)}")
+      rescue Foreman::Exception => e
+        logger.info("Error running Ansible roles on #{self} before_provision: "\
```
ares: not in this commit, but I'm thinking we should have "error" notification blueprint in core and this should render such notification, another example would be, facts parsing failed...

dLobatog: Just as a reminder to myself, audience for that kind of notification should be `host.owner`

ares:
tests are green on 4/5 combinations already, ruby 2.3 started just a moment ago, I'll risk that and merge :-) thanks @dLobatog!

PR #180, opened 2018-06-21
Fixes #23879 - Disable Ansible Callback when running commands

bastilian:
When a job is run via Ansible and is executing a command it should not change the configuration status, for this the callback is disabled via setting 'ANSIBLE_CALLBACK_WHITELIST' to an empty string and not run **any** callbacks, including the foreman callback that would create a report and update the configuration status.

This can be tested by running a "Run Ansible Command" job. Executing it via a Smart Proxy requires to include the `foreman_ansible_core` gem for [smart_proxy_ansible](https://github.com/theforeman/smart_proxy_ansible).

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -20,8 +20,18 @@ if defined? ForemanRemoteExecution
           super(template_invocation, host).merge(
             'ansible_inventory' => ::ForemanAnsible::InventoryCreator.new(
               [host], template_invocation
-            ).to_hash.to_json
+            ).to_hash.to_json,
+            :remote_execution_command => ansible_command?(
+              template_invocation.template
+            )
           )
+        end
+
+        private
+
+        def ansible_command?(template)
+          template.remote_execution_features.
+            where(:label => 'ansible_run_host').empty?
         end
       end
     end
--- a/lib/foreman_ansible_core/command_creator.rb
+++ b/lib/foreman_ansible_core/command_creator.rb
@@ -1,23 +1,34 @@
 module ForemanAnsibleCore
   # Creates the actual command to be passed to foreman_tasks_core to run
   class CommandCreator
-    attr_reader :command
-
     def initialize(inventory_file, playbook_file, options = {})
       @options = options
-      @command = [{ 'JSON_INVENTORY_FILE' => inventory_file }]
-      @command << 'ansible-playbook'
-      @command = command_options(@command)
-      @command << playbook_file
+      @playbook_file = playbook_file
+      @inventory_file = inventory_file
+      command
+    end
+
+    def command
+      parts = [environment_variables]
+      parts << 'ansible-playbook'
+      parts.concat(command_options)
+      parts << @playbook_file
+      parts
     end
 
     private
 
-    def command_options(command)
-      command.concat(['-i', json_inventory_script])
-      command.concat([setup_verbosity]) if verbose?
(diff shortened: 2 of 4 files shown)


app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -20,9 +20,16 @@ def proxy_command_options(template_invocation, host)
...
-            ).to_hash.to_json
+            ).to_hash.to_json,
+            :remote_execution_command => ansible_command?(template_invocation.template)
           )
         end
+
+        private
+
+        def ansible_command?(template)
+          template.job_category == 'Ansible Commands'
```
ares: Could we use the rex feature instead of a category? This would still generate reports for e.g. service restart or package install. We only want that for roles plays. So it would be something like unless feature is play roles.

bastilian:
@ares changed it to us the `ansible_run_host` feature for detection. When it is **not** using `ansible_run_host` it disables the callbacks. At the moment only the "Ansible Roles" job template is using this feature.

ares:
> At the moment only the "Ansible Roles" job template is using this feature.

And that was the goal I think, only these runs should generate the configuration report. I guess now it also means that facts won't be uploaded for any other runs right?

bastilian:
@ares yes, facts are **not** uploaded on these runs as well, only on those where the callback is left active.

The test failures of `PlaybookRunnerTest::working_dir#test_0001_creates temp one if not provided` and `PlaybookRunnerTest::roles dir#test_0001_reads default when none provided` are failing for me locally with master too and don't seem to be related to these changes as far as I can see. Maybe @dLobatog knows.

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -20,9 +20,16 @@ def proxy_command_options(template_invocation, host)
...
-            ).to_hash.to_json
+            ).to_hash.to_json,
+            :remote_execution_command => ansible_command?(template_invocation.template)
           )
         end
+
+        private
+
+        def ansible_command?(template)
+          template.remote_execution_features.where(:label => 'ansible_run_host').empty?
```
dLobatog: Shouldn't this line be `.present?`? Right now it's disabling the callback for *all* runs, except for ones related to the 'ansible run host' feature.

bastilian: It is correct to check if there are `any?` features with `ansible_run_host`. When the feature would be present? it would not be a `ansible_command?` and therefore not disable callback. Only those using the feature should have this set to "false" and "enable" the callback, by not setting the environment variable to empty.

bastilian:
[test foreman_ansible]

bastilian:
Not sure why it reports failed here, all tests that ran on Jenkins are :green_heart: 

@dLobatog have we resolved the confusion about when the callback is disabled? I'm honestly still confused a bit, but I think how it is implemented should be correct. 

If the job does **not** have the 'ansible_run_host' feature it is **not** a task triggered by "Play Roles" or the Ansible roles job (therefore `.empty?`). 
If that is true it is a "ansible command" or in `CommandCreator` a "rex_command". (We could align that and see if we can make the implementation more self explanatory.)
When true, we don't want a callback and set the `ANSIBLE_CALLBACK_WHITELIST` to an empty string.

We could make the implementation clearer if we introduce a new rex_feature called "ansible_callback" (or something), change the `empty?` to `.present?` and adjust the existing templates and job_compositions. When users would like to receive reports for a specific task they can add the feature to their jobs.

PR #181, opened 2018-06-25
Fixes #24061 - Make tests ansible_dir use Dir.home

dLobatog:
Since I changed :ansible_dir on foreman_ansible_core to point to
`Dir.home`, this is making tests fail in Jenkins.
The tests were not changed, so 'ansible_dir' is expected to be still
'/etc/ansible' instead of 'Dir.home'.


bastilian:
:+1: Thank @dLobatog! merged majority of test was already green.

PR #183, opened 2018-07-05
Fixes #24163 - Power Action does not do anything

dLobatog:
The playbook first runs 'echo' of 'shutdown host' - that fails because
that returns -1 'Failed to parse time specification:' ('host' is not a
valid time specification)

I don't find much of an use for that line, so the template currently
just runs the action and works.


ares:
Tests are green, it skipped mysql. The fix works as expected. Thanks @dLobatog, merging.

PR #185, opened 2018-07-11
Fixes #24207 - don't display run buttons for unauthorized users

ares:
This fixes the authorization for all means of playing roles - button on host detail page, button for hostgroup and bulk action on hosts list page. If user is missing create_job_invocations permission, they can't play roles. I was thinking of extending ansible roles manager role with this permission, but this permission grants also access to whole SSH stack so that would be probably unexpected. The minimal set of built in roles for using ansible is following:

* Viewer
* Ansible Roles Manager
* Remote Execution User


ares:
made rubocop happy

PR #189, opened 2018-07-17
Refs #24163 - shutdown with 1m delay so callback can return 

dLobatog:
Delaying the shutdown 1 minute so the Ansible job can return and the job
invocation can be green. Otherwise 'shutdown' may work but the job
invocation will fail.

https://github.com/theforeman/community-templates/pull/493 has been submitted too

cc @lpramuk


ares:
verified the same in community templates repo, thanks @dLobatog, merging!

PR #190, opened 2018-07-19
Fixes #24296 - Remove option to import roles from Foreman host

dLobatog:
In the past, the duplicity of being able to do everything through
proxies or Foreman itself has led to bugs and confusion in some users.
We are removing these methods one by one, to begin with, removing the
import roles option is very easy to do.


ares:
mysql error is unrelated, the patch works just fine, thanks @dLobatog, merging

PR #191, opened 2018-07-23
Fixes #24038 - fix host roles inheritance from hostgroup

ares:
The same problem still remained for host. To reproduce:
1) create hg H, assign role R1
2) create host A, assign role R2 and hg H
3) try to create host B, chose H as HG and switch to Ansible roles tab, you see both R1 and R2

it's even more scary since the inventory has the same problem. One of the reason why this happened is the fact HasManyAnsibleRoles was included both to host and hostgroup, while it has different meaning in both classes. Instead of adding more if conditions into all_ansible_roles, I copied out both methods which led to code simplification. The core of the change though is the change of #inherited_ansible_roles in Host extensions


bastilian:
The CI failures seem unrelated. `import_host` in core was changed (https://github.com/theforeman/foreman/pull/5855) and breaks the overwriting.

ares:
[test foreman_ansible] the fixing PR got in

I'll add tests today

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -15,12 +15,11 @@ module HostManagedExtensions
                     :only_explicit => true
 
       before_provision :play_ansible_roles
-      include ForemanAnsible::HasManyAnsibleRoles
       audit_associations :ansible_roles
 
       def inherited_ansible_roles
         return [] unless hostgroup
-        hostgroup.all_ansible_roles
+        hostgroup.inherited_and_own_ansible_roles
```
dLobatog: Isn't this the only thing that changed? As far as I can understand from the changes, the module `HasManyAnsibleRoles` was doing the same exactly that this is doing:

`(ansible_roles + inherited_ansible_roles + host_ansible_roles).uniq` for Hostgroup
`(ansible_roles + inherited_ansible_roles).uniq` for Host

ares: yes, but if I kept it in all_ansible_roles method, I'd need to add additional argument whether when I ask for host roles I want to include hostgroups' hosts and their role or not. That was second if in the method that would need to check for context where it's called, that is a good indication that this DRYification was not a good idea, since these two methods are different.

PR #192, opened 2018-07-24
Fixes #24361 - import_host arguments in foreman-core changed

bastilian:
`import_host`'s arguments[1] in core have changed and needs to be fixed.

This change doesn't require a specific amount of arguments and should be compatible with foreman-core that still has the third argument present.

[1] https://github.com/theforeman/foreman/pull/5855

### DIFF
--- a/app/models/concerns/foreman_ansible/host_managed_extensions.rb
+++ b/app/models/concerns/foreman_ansible/host_managed_extensions.rb
@@ -47,10 +47,12 @@ module ForemanAnsible
     # rubocop:enable Metrics/BlockLength
     # Class methods we may need to override or add
     module ClassMethods
-      def import_host(hostname, certname = nil, deprecated_proxy = nil)
-        host = super(hostname, certname, deprecated_proxy)
-        if IPAddress.valid?(hostname) && Nic::Interface.find_by(:ip => hostname)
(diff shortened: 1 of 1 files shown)


app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -43,14 +42,20 @@ def play_ansible_roles
...
+      end
     end
     # rubocop:enable Metrics/BlockLength
     # Class methods we may need to override or add
     module ClassMethods
-      def import_host(hostname, certname = nil, deprecated_proxy = nil)
-        host = super(hostname, certname, deprecated_proxy)
-        if IPAddress.valid?(hostname) && Nic::Interface.find_by(:ip => hostname)
-          host = Nic::Interface.find_by(:ip => hostname).host
+      def import_host(*args)
```
dLobatog: Does this break compatibility with any version? As far as I can see previous versions should be able to work with it too, but if you could confirm it it'd be great

bastilian: It rather improves compatibility. :) We simply pass on arguments to `super` without touching and later get the first argument as hostname.

I tested this with the definition of `import_host` before https://github.com/theforeman/foreman/pull/5855 and it works as well.

PR #193, opened 2018-07-24
Fixes #24362 - Ansible Variables can be imported as LookupKeys

dLobatog:
It would be really useful to be able to override parameters the same way
we do with Puppet: having smart matchers, default values, etc..
Since the LookupKey object has been written in a way that is really
reusable, we may essentially adapt Ansible Variables to be a kind of
LookupKey.

Pending tests, and a PR with the API should follow. It's critical to show these variables in the same fashion we show puppet params in the Host and Host Group form (Parameters tab).

It's necessary to have https://github.com/theforeman/smart_proxy_ansible/pull/9 for it to work.

[Screenshot%20from%202018-07-02%2016-55-42|690x276
Screenshot from 2018-07-02 16-55-42.png1910x765 85.7 KB
](https://community.theforeman.org/uploads/default/original/2X/b/babe9cd8ad6f2a7efcf79c8771e0489fd80a193c.png)

Menu entry:

[Screenshot%20from%202018-07-02%2016-54-17|690x388
Screenshot from 2018-07-02 16-54-17.png1920x1080 161 KB
](https://community.theforeman.org/uploads/default/original/2X/5/5a48dba669d4a16c83d472c592a30d8748a7e0c0.png)

Edit page:

[Screenshot%20from%202018-07-02%2016-56-23|690x370](https://community.theforeman.org/uploads/default/original/2X/9/9e79a91f3433a3b940280941c62241fe0be785b0.png)

### DIFF
--- a/app/controllers/ansible_roles_controller.rb
+++ b/app/controllers/ansible_roles_controller.rb
@@ -1,11 +1,7 @@
 # UI controller for ansible roles
 class AnsibleRolesController < ::ApplicationController
   include Foreman::Controller::AutoCompleteSearch
-
-  before_action :find_resource, :only => [:destroy]
-  before_action :find_proxy, :only => [:import]
-  before_action :create_importer, :only => [:import, :confirm_import]
-  before_action :default_order, :only => [:index]
+  include ForemanAnsible::Concerns::ImportControllerHelper
 
   def index
     @ansible_roles = resource_base.search_for(params[:search],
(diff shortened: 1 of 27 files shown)


bastilian:
@dLobatog cool!

When there are no new variables to import and no changes it fails with `undefined method `[]' for nil:NilClass` as the `params[:changes]` is empty.

Another issue, but it's more UX is that the role will need to be imported first, before users can import their variables. Maybe we can show a info when no variables to import to import roles first.
A cleaner solution might be to combine Roles and Variable importing into a service and on both Roles and Variable importing use that and always try to import both.

bastilian:
@dLobatog awesome! Works like a charm now!

PR #194, opened 2018-07-25
Unpin net-ssh dependency

ares:


ares:
as discussed at https://github.com/theforeman/foreman_ansible/commit/e258437698c7cc5115f5c7b3673bea4cb84273e6#commitcomment-29794543 the pin block deb build, I've tested with newer net-ssh, I think we should simply remove the version lock

foreman_ansible_core.gemspec
```diff
@@ -20,5 +20,5 @@ DESC
   s.add_development_dependency 'rubocop', '~> 0.52'
   s.add_dependency 'foreman-tasks-core', '~> 0.1'
   s.add_dependency 'foreman_remote_execution_core', '~> 1.1'
-  s.add_dependency 'net-ssh', '~> 4.0'
+  s.add_dependency 'net-ssh'
```
ekohl: Should it be `>= 4.0`?

ares: I don't think there's such requirement but why not, the version bundled in core for debs is probably >= 4.0, pushed new version

PR #196, opened 2018-08-01
Unpin net-ssh completely

ares:
as discussed here https://github.com/theforeman/foreman-packaging/pull/2776#issuecomment-408705494 and on Monday sync up, we should remove the pin compleltely for new version

hopefully this is a quick merge


dLobatog:
[test foreman-ansible]

dLobatog:
[test foreman_ansible]

PR #186, opened 2018-07-11
Fix Performance/HashEachMethods cop

ares:
this is reported by rubocop 0.58.1

PR #197, opened 2018-08-02
Fixes #24525 - Replace notice with info in UI controller

xprazak2:


app/controllers/ansible_roles_controller.rb
```diff
@@ -34,7 +34,7 @@ def import
 
   def confirm_import
     @importer.finish_import(params[:changed])
-    notice _('Import of roles successfully finished.')
+    info _('Import of roles successfully finished.')
```
bastilian: In this instance it might make more sense to use the `success` helper.

xprazak2:
I changed `info` to `success`. `each_value` is also not defined for controller params, so I replaced it with `each`

bastilian:
@xprazak2 cool! However... https://github.com/theforeman/foreman_ansible/pull/186 :) that is also why code climate is not liking it.

xprazak2:
Yeah... I went for a different solution to keep the #186 and make codeclimate happy as well.

PR #198, opened 2018-08-15
Fixes #24623 - Correct documented API endpoints for playing roles

ofedoren:
I've moved routes `/host(group)s/(multiple)_play_roles` from `/ansible/api` namespace to global `/api/` so they actually work. Because apidocs documents them without `/ansible` prefix there were 2 options:

1. Change documentation from e.g. `api :POST, '/hosts/play_roles'` to `api :POST, '/ansible/api/hosts/play_roles'`. See https://github.com/theforeman/foreman_ansible/blob/ec84e8b201ef4a9179c356608e930f1f0f58f5c6/app/controllers/api/v2/ansible_roles_controller.rb#L7.
or
2. Move routes from `/ansible/api` namespace to global `/api/` because those routes use documentation defined in the Foreman's `hosts` resource. In other words: after including the `HostsControllerExtensions` the `base_url` for all new endpoints will be `/api/` not the `/ansible/api`. So. if use the first one approach, Apipie will generate something like `/api/ansible/api/` which is wrong.

### DIFF
--- a/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
@@ -12,7 +12,7 @@ module ForemanAnsible
         included do
           api :POST, '/hostgroups/:id/play_roles',
               N_('Plays Ansible roles on a hostgroup')
-          param :id, String, :required => true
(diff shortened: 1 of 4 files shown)


dLobatog:
Closing as per https://github.com/theforeman/foreman_ansible/pull/205#

PR #200, opened 2018-08-16
Fixes #18951 - Assigning Ansible roles over the API

ofedoren:
Based on https://github.com/theforeman/foreman_ansible/pull/81

PR #199, opened 2018-08-16
Fixes #24626 - Import selected roles only via API

ofedoren:

PR #81, opened 2017-03-17
Fixes #18951 - Assigning Ansible roles over the API

Gerrit91:
At the moment, it is only possible to assign Ansible roles from the
Foreman UI. Due to this reason, Foreman cannot be used as a pure backend
service and it cannot be configured externally. In this feature, new
endpoints are added to the API, which allow setting roles as well as
listing assigned roles.

PR #205, opened 2018-09-12
Fixes #23227, #24623, #24626, #18951 - Ansible Roles API

ofedoren:
Cherry-picked and rebased from [PR 161](https://github.com/theforeman/foreman_ansible/pull/161), [PR 198](https://github.com/theforeman/foreman_ansible/pull/198), [PR 199](https://github.com/theforeman/foreman_ansible/pull/199), [PR 200](https://github.com/theforeman/foreman_ansible/pull/200).

P.S. [Hammer plugin](https://github.com/theforeman/hammer-cli-foreman-ansible) currently doesn't have commands for these changes/endpoints, but I've prepared them and could create a PR there with them for testing. Anyway I was going to do it, but after this is merged (for a case of additional changes in here).

### DIFF
--- a/app/controllers/api/v2/ansible_roles_controller.rb
+++ b/app/controllers/api/v2/ansible_roles_controller.rb
@@ -12,8 +12,8 @@ module Api
       end
 
       before_action :find_resource, :only => [:show, :destroy]
-      before_action :find_proxy, :only => [:import, :obsolete]
-      before_action :create_importer, :only => [:import, :obsolete]
+      before_action :find_proxy, :only => [:import, :obsolete, :fetch]
+      before_action :create_importer, :only => [:import, :obsolete, :fetch]
 
       api :GET, '/ansible_roles/:id', N_('Show role')
       param :id, :identifier, :required => true
@@ -33,8 +33,9 @@ module Api
 
       api :PUT, '/ansible_roles/import', N_('Import Ansible roles')
       param :proxy_id, :identifier, N_('Smart Proxy to import from')
+      param :role_names, Array, N_('Ansible role names to import')
       def import
-        @imported = @importer.import!
+        @imported = @importer.import!(role_names)
       end
 
       api :PUT, '/ansible_roles/obsolete', N_('Obsolete Ansible roles')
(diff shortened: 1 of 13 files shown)


app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb
```diff
@@ -6,31 +6,63 @@ module HostsControllerExtensions
...
+          api :GET, '/hosts/:id/ansible_roles',
+              N_('List all Ansible roles for a host')
+          param :id, :identifier, :required => true
+
+          def ansible_roles
+            return unless @host
+            @ansible_roles = @host.all_ansible_roles
+          end
+
+          api :POST, '/hosts/:id/assign_ansible_roles',
```
dLobatog: Why this endpoint? It seems to me that we can definitely do this already by sending a POST request to `api/v2/hosts/:id` with the 'ansible_roles' attribute

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -6,31 +6,59 @@ module HostgroupsControllerExtensions
...
+              N_('List all Ansible roles for a hostgroup')
+          param :id, :identifier, :required => true
+
+          def ansible_roles
+            find_resource
+            return unless @hostgroup
+            @ansible_roles = @hostgroup.all_ansible_roles
+          end
+
+          api :POST, '/hostgroups/:id/assign_ansible_roles',
```
dLobatog: Same question as for the one for this endpoint in hosts.

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -41,8 +42,27 @@ def obsolete
         @obsoleted = @importer.obsolete!
       end
 
+      api :GET, '/ansible_roles/fetch',
```
dLobatog: Nice job!

app/controllers/foreman_ansible/concerns/job_invocation_helper.rb
```diff
@@ -7,10 +7,14 @@ module JobInvocationHelper
       def job_composer(feature_name, target)
         composer = ::JobInvocationComposer.for_feature(feature_name, target)
         return composer if composer.save
-        raise ::Foreman::Exception.new(
-          format(N_('Could not run Ansible roles for %{host}'),
-                 :host => target)
-        )
+        msg = if target.blank?
+                format(N_('There are no Ansible roles to play'),
```
dLobatog: Why passing `:target => target.class` to format, if it's not used at all?

app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb
ofedoren: Sure we can, It's just added "from" https://github.com/theforeman/foreman_ansible/pull/81/files#diff-e3a172920758b4851cabcb23ced54a22R64 and also for easier use from hammer. If you think this is redundant, tell me and I'll remove that.

ofedoren:
@dLobatog ping. I rebased, but tests are red even though 0 skipped, 0 failed?

ares:
@ofedoren we discussed this today with bigger group, it would be great to send a separate PR with assign_* endpoints. Sorry for back and forth. I'm merging this meanwhile, test failure is only on sqlite because of some machine problem.

PR #206, opened 2018-09-13
Refs #24262 - bump min version of REX

ares:
For 2.2-stable, I'll bump to 1.5.6 which will also get the fix today.


bastilian:
Thank you, @ares!

PR #211, opened 2018-09-25
Fixes #25030 - importer works with Parameters

lzap:
The problem is that Rails Parameters object is passed into the UiRolesImporter instead of hash and that does not have each_value. Result: `each_value' for <ActionController::Parameters:0x00007faa5d0cb3b8> Did you mean? each_pair"

### DIFF
--- a/app/services/foreman_ansible/ui_roles_importer.rb
+++ b/app/services/foreman_ansible/ui_roles_importer.rb
@@ -14,13 +14,13 @@ module ForemanAnsible
     end
 
     def create_new_roles(changes)
-      changes.each_value do |new_role|
+      changes.each_pair do |_, new_role|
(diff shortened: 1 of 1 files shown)


ares:
interesting, so this is not a regression but probably a failure after rails upgrade? or did you find introducing commit?

lzap:
I haven't digg into history, probably Rails upgrade regression, these were introduced in 4.2 I think? Weird, really weird.

ares:
ok, let's wait for tests, I'll look at recent changes just to be sure

ares:
ah, this is a regression caused by the rubocop fix introduced in f6bf9d46 in 2.2.7

lzap:
Let's see: https://github.com/rails/rails/pull/33979 ;-)

PR #213, opened 2018-09-28
Fixes #25066 - Remove foreman_ansible_core

dLobatog:
All of this functionality is now exclusively used in the proxy, hence it
makes no sense to keep the code in both places, especially since it has
caused problems in packaging such as #24879.

Tests and new features should be moved to smart_proxy_ansible

cc @bastilian @ares @ekohl


dLobatog:
Same as https://github.com/theforeman/foreman_ansible/pull/216#issuecomment-432820677 , 2 tests failed at the 'Junit result report' step after passing everything. 2.3 and 2.4 with postgresql.

ares:
Thanks @dLobatog, merging!

PR #214, opened 2018-10-08
Fixes #22279 - Do not duplicate hosts on callback

xprazak2:
Prevents creating duplicate hosts for Ansible callback when using partially qualified name. If there are multiple matches on partial host name, we have no way of knowing which host the report belongs to and we create a new one as before.

### DIFF
--- a/app/services/foreman_ansible/ansible_report_importer.rb
+++ b/app/services/foreman_ansible/ansible_report_importer.rb
@@ -15,6 +15,19 @@ module ForemanAnsible
           @host = Nic::Interface.find_by(:ip => hostname).host
         end
         super
+        partial_hostname_match(hostname)
+      end
+
+      def partial_hostname_match(hostname)
+        return @host unless @host.new_record?
+        hosts = Host.where(Host.arel_table[:name].matches("#{hostname}.%"))
+        if hosts.count > 1
+          msg = "More than 1 host found for name #{hostname}, "
+          msg += 'please use host FQDN when uploading reports'
+          Rails.logger.warn msg
+          return @host
+        end
+        @host = hosts.first || @host
       end
     end
   end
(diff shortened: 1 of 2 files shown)


xprazak2:
[test foreman_ansible]

ares:
The error seems consistent

```

Error:
AnsibleReportImporterTest#test_0004_creates new host when multiple hosts for partially qualified name are found:
NoMethodError: undefined method `new_record?' for nil:NilClass
    /var/lib/workspace/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/test/unit/services/ansible_report_importer_test.rb:34:in `block in <class:AnsibleReportImporterTest>'
```

xprazak2:
[test foreman_ansible]

app/services/foreman_ansible/ansible_report_importer.rb
```diff
@@ -15,6 +15,23 @@ def host
...
+      def partial_hostname_match(hostname)
+        return @host unless @host.new_record?
+        hosts = Host.search_for("name ~ #{hostname}")
+        if hosts.count > 1
+          msg = "More than 1 host found for name #{hostname}, "
+          msg += 'please use host FQDN when uploading reports'
+          Rails.logger.warn msg
+          return @host
+        end
+        found_host hosts.first
```
ares: the code in method above also sets `@host = $found_host`, if we find a host based on search_for, should we set it there too?

also I wouldn't extract this logic to a separate method, just return `hosts.first || @host` or even better, `return @host if hosts.empty?` above the if

xprazak2:
Updated, I use `where` instead of scoped search query and set the `@host` properly.

app/services/foreman_ansible/ansible_report_importer.rb
```diff
@@ -15,6 +15,19 @@ def host
           @host = Nic::Interface.find_by(:ip => hostname).host
         end
         super
+        partial_hostname_match(hostname)
+      end
+
+      def partial_hostname_match(hostname)
+        return @host unless @host.new_record?
+        hosts = Host.where(Host.arel_table[:name].matches("%#{hostname}%"))
```
ares: this should be matching `hostname.%` and perhaps start with short name exact match first

xprazak2:
I removed the leading `%` and added `.`

We do not need to search by full name, superclass does that for us: https://github.com/theforeman/foreman/blob/develop/app/services/report_importer.rb#L69-L74

PR #217, opened 2018-10-25
Fixes #25309 - Switcher for ansible roles

xprazak2:
* ~needs tests~
* ~LoadingState was taken from katello and should be moved to core~
* ~Pagination was taken from katello because the one we have in foreman relies on being rendered directly in erb template~
* pagination is not disabled when loading another page, should be fixed in patternfly-react
* ~handle errors when loading roles fails~


xprazak2:
needs https://github.com/theforeman/foreman/pull/6171

xprazak2:
@Rohoover, @terezanovotna, any thoughts?

ares:
also @theforeman/ui-ux code comments would be appreciated

tstrachota:
I didn't test it live but it looks great. Much cleaner than before.

I think that adding an input for filtering by role names would be a big usability improvement. We could reuse the recently added search box component for that.

tstrachota:
FYI I just merged the dependency into the core.

Rohoover:
Looks great. There are a few details here that are missing, which are present in the Repositories page that this is modeled after. This includes the lines between the rows, and that the Enabled Side is grey. I think it would be in our best interests to keep those details as close as possible so avoid inconsistencies. 

We have a few changes that will happen to the Repos page design overtime, including reducing the font size of the rows. We would want to keep an eye on those changes.

Laviro:
@xprazak2 @Rohoover 
Hey guys, I have been working on something very similar on Patternfly - [Dual List Selector](https://github.com/patternfly/patternfly-react/pull/395), 
and it is close to be finished.. maybe check if it would be a good fit for you,
so we won't have duplication.

xprazak2:
Comparison with the gray background:




The main difference compared to repo pages is that we are in a form with grey border. The form padding between the border and grey background looks strange to me.

I cleaned up a bit and added selectors and linting. Strangely enough the you-dont-need-lodash-underscore plugin did not trigger any offences to me. 

@Laviro: Dual List Selector looks impressive, I am just a bit worried about performance on a large sets as it is not paginated.

xprazak2:
LoadingState componetn is now in patternfly-react and Pagination was moved to core, I opened https://github.com/theforeman/foreman/pull/6322 to bump the versions. I cleaned up the code and added tests.

xprazak2:
Added a fix for available roles count in pagination.

app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
```diff
@@ -1,15 +1,28 @@
...
+    <% roles = f.object.is_a?(Hostgroup) ? roles_attrs(f.object.inherited_and_own_ansible_roles) : roles_attrs(f.object.all_ansible_roles) %>
+    <% class_name = f.object.is_a?(Hostgroup) ? 'Hostgroup' : 'Host' %>
+    <%= mount_react_component('AnsibleRolesSwitcher', '#ansible_roles_switcher', { :initialAssignedRoles => roles,
+                                                                                   :inheritedRoleIds => f.object.inherited_ansible_roles.map(&:id),
+                                                                                   :availableRolesUrl => ui_ansible_roles_path,
+                                                                                   :resourceId => f.object.id,
+                                                                                   :resourceName => class_name
+                                                                                    }.to_json) %>
+  <% else %>
+    <div class="blank-slate-pf">
```
ohadlevy: would it makes more sense to use empty state component instead? I think it might give you more power over time to manage the entire page / life cycle from react?

webpack/components/AnsibleRolesSwitcher/AnsibleRole.js
```diff
@@ -0,0 +1,33 @@
...
+import './styles.scss';
+
+const clickHandler = (onClick, role) => (event) => {
+  event.preventDefault();
+  onClick(role);
+};
+
+
+const actionButton = (icon, onClick, role) => (
+  <button
```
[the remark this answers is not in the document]
xprazak2: It has additional styling which does not look very good in this case:

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherReducer.js
```diff
@@ -35,22 +37,16 @@ const initAssignedRoles = (state, initialAssignedRoles, inheritedRoleIds) => {
...
+  state.merge(addItemNewState(state, payload.role));
 
 const ansibleRoleRemove = (state, payload) =>
-  state.set('assignedRoles', removeItem(state.assignedRoles, payload.role))
-       .set('itemCount', state.itemCount + 1);
-
-const addItem = (list, item) => ([...(list || []), item]);
-
-const removeItem = (list, item) => list.filter(listItem => item.id !== listItem.id);
+  state.merge(removeItemNewState(state, payload.role));
```
ohadlevy: hmm.. sorry, maybe I wasn't clear, the idea was to consider using reselect instead of storing it in the store? /cc @sharvit

xprazak2:
@sharvit, thank you for detailed review, main changes:

* the file structure was improved
* react-redux-test-utils is used in tests
* several new components were extracted into separate files (AnsiblePermissionDenied, AnsibleRoleActionButton, AnsibleRolesSwitcherError)
* reducer was reduced to updating store, logic was moved to selectors using recompose
* still using the camel/snake case helpers to keep this working until they get into [core](https://github.com/theforeman/foreman/pull/6433)

xprazak2:
Using the helpers for changing case from core now.

ares:
@sharvit mind to take another look? we'd like to get this in soon

.eslintrc
```diff
@@ -1,6 +1,8 @@
 {
   "root": true,
-  "extends": "airbnb-base",
+  "extends": [
```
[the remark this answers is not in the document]
sharvit: This is an easy change I see as important because it automatically finds many useful errors while it will keep the code style unified with core.

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherReducer.js
```diff
@@ -0,0 +1,91 @@
...
+
+  switch (action.type) {
+    case ANSIBLE_ROLES_REQUEST:
+      return state.set('loading', true);
+    case ANSIBLE_ROLES_SUCCESS:
+      return ansibleRolesSuccess(state, payload);
+    case ANSIBLE_ROLES_FAILURE:
+      return state.merge({ error: payload.error, loading: false });
+    case ANSIBLE_ROLES_ADD:
+      return ansibleRoleAdd(state, payload);
```
sharvit: IMHO those cases will be more readable as:

```js
case ANSIBLE_ROLES_ADD:
  return state
    .set('assignedRoles', state.assignedRoles.concat(payload.role))
    .set('itemCount', state.itemCount + 1)
```

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherReducer.js
```diff
@@ -0,0 +1,91 @@
...
+
+const removeItemNewState = (state, role) => ({
+  assignedRoles: removeItem(state.assignedRoles, role),
+  results: addItem(state.results, role),
+  itemCount: state.itemCount + 1,
+});
+
+const addItemNewState = (state, role) => ({
+  assignedRoles: addItem(state.assignedRoles, role),
+  itemCount: state.itemCount - 1,
```
sharvit: Am I missing something? shouldn't it do `itemCount + 1` while the remove should do the `itemCount - 1`?

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherSelectors.js
```diff
@@ -0,0 +1,62 @@
...
+const switcherState = state => state.foremanAnsible.ansibleRolesSwitcher;
+
+const markInheritedRoles = (roles, inheritedRoleIds) => roles.map(role => (
+  includes(inheritedRoleIds, role.id) ?
+    { ...role, inherited: true } :
+    role
+));
+
+export const selectResults = state =>
+  Immutable.asMutable(uniq(switcherState(state).results)).sort(compare);
```
sharvit: Why does it need to be `mutable`? If you are going to break to immutability, IMO it is better to do it in the component.

xprazak2: because `sort` mutates original array. I wrap it with `Immutable` to keep the immutability.

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherReducer.js
xprazak2: `itemCount` tracks the item count of the unassigned roles for pagination to display correct values, so if we add a role to a host, we need to decrement the unassigned roles by 1.

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.scss
```diff
@@ -0,0 +1,25 @@
+@import '~patternfly/dist/sass/patternfly/_color-variables';
+@import '~patternfly/dist/sass/patternfly/_variables';
+@import '~patternfly/dist/sass/patternfly/_loading-state';
```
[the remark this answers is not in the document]
xprazak2: It makes a difference:

ls-wop
ls-wp

xprazak2:
Rebased and updated.

ares:
And also I liked the functionality that I could click anywhere on role name to move it to the other column, now I need to click the icon :-)

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherReducer.js
```diff
@@ -0,0 +1,77 @@
...
+});
+
+const ansibleRoles = (state = initialState, action) => {
+  const { payload } = action;
+
+  switch (action.type) {
+    case ANSIBLE_ROLES_REQUEST:
+      return state.set('loading', true);
+    case ANSIBLE_ROLES_SUCCESS:
+      return ansibleRolesSuccess(state, payload);
```
sharvit: Could you embed it inside as well?

```js
case ANSIBLE_ROLES_SUCCESS:
  return state.merge({
    loading: false,
    itemCount: Number(payload.subtotal),
    pagination: { page: Number(payload.page), perPage: Number(payload.perPage) },
    results: payload.results,
    assignedRoles: payload.initialAssignedRoles,
    inheritedRoleIds: payload.inheritedRoleIds,
  });
```

.eslintrc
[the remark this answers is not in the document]
sharvit: Any update here?

sharvit:
> I've seen the discussion about Dual List Selector which is now in patternfly-react. I share the concern with having hundreds or even thousands of items in this. But at the same time, this new component does needs a bit more styling. Today it's very hard to use with more than 7 roles, user need to scroll down even on full HD resolution. Could we style it so it's more compact? It would be great if it could look as similar as dual list selector as possible.

I think we can create a workaround and update the styling to use a **compact mode**, and I think it should be done with collaboration with patternfly. @Laviro?

> the main benefit of having having own implementation is that we can have the selector working with hundred of records, but we don't provide search functionality. I think it's fine to keep it out of this PR, but would it be possible to add it too? I don't want to list page by page, to find specific ansible role.

If it implement the scoped search I think we can like it with the new `AutoComplete` component. @Laviro?

xprazak2:
I added tooltips and clicking on anywhere on list item will move it, which makes the buttons kind of redundant, but I left them there for now. I also decreased the padding for list items to make everything more vertically compact. Current look:

compact-switcher

If I decrease font and icon size, I can go even smaller, but it seems too small and crowded for me:

tiny-switcher

Searching was mentioned [before](https://github.com/theforeman/foreman_ansible/pull/217#issuecomment-433321801) and I [agree](https://github.com/theforeman/foreman_ansible/pull/217#issuecomment-446237991) it would be very helpful. I have a draft ready, but I would like to do it as a follow-up, because this is already big enough as it is.

ares:
Few more design changes as discussed locally:

1) the font shouldn't be bold (disabling font-weight: 600 did the job)
1) the font should be smaller (disabling font-size: 16px did the job)
1) default pagination should be 10, so users still see submit button
1) hovering on the row should keep cursor: hand, same as the plus icon

otherwise :+1:

xprazak2:
Updated, font is smaller and not bold, pagination set to 10 by default and list items have pointer as cursor on hover.

ares:
Design wise :+1: , now the problem is that when on right side I have more than 10 records, whenever I go to next page in left selector or change pagination number, it drops everything from right table besides first page. I showed that live to make it clear.

xprazak2:
Updated, the assigned roles should stay assigned when users change pagination for available roles.

ares:
I can confirm it now works as expected. Ready for merge once we have ack from @sharvit and mr jenkins confirms.

sharvit:
I have tested it and it works well.

My only concern about the code is the face you are not using `eslint` with our recommended configurations.

I tried running it and it found very useful 42 problems, most of them are about missing `prop-types` which are important.

You can use this [.eslintrc](https://gist.github.com/sharvit/05b8f0c98623fecfa9cdbcca98524eec) file to make it work.

xprazak2:
I changed the eslint config and fixed linting errors.

ares:
Tests are red now

ares:
Thanks @xprazak2 and @sharvit for the review, merging.

Rohoover:
"If I decrease font and icon size, I can go even smaller, but it seems too small and crowded for me:"

Agreed too crowded.

tbrisker:
Looks like this pulled in different npm module versions to what we have in core, causing issues with packaging (and possibly also functionality). see https://github.com/theforeman/foreman-packaging/pull/3650 for details.

ares:
But switcher was merged in 2.2 already, the 3.0 release only adds enc and inventory changes IIRC

PR #218, opened 2018-10-28
Fixes #25319 - Add authorized importer claim

UXabre:


UXabre:
BTW, this fixes the issue in which smart-proxies don't seem to be able to send their facts to Foreman (out-of-the-box) work around is to add the smart-proxy to the trusted_host array; which should not be needed normally. This is related to the work that has been done in PR https://github.com/theforeman/foreman/pull/1975

ares:
Thanks @UXabre, merging!

PR #219, opened 2018-10-30
Fixes #25246 - Imports IPv6 facts

UXabre:
If we don't do this, seems the IPv6 address is cleared, resulting in other unwanted side-effects (rebuild not working etc.)

### DIFF
--- a/app/services/foreman_ansible/fact_parser.rb
+++ b/app/services/foreman_ansible/fact_parser.rb
@@ -55,7 +55,11 @@ module ForemanAnsible
       interface.tr!('-', '_') # virbr1-nic -> virbr1_nic
       interface_facts = facts[:"ansible_#{interface}"]
       ipaddress = ip_from_interface(interface)
-      HashWithIndifferentAccess[interface_facts.merge(:ipaddress => ipaddress)]
+      ipaddress6 = ipv6_from_interface(interface)
+      HashWithIndifferentAccess[
+        interface_facts.merge(:ipaddress => ipaddress,
+                              :ipaddress6 => ipaddress6)
+      ]
     end
 
     def ipmi_interface; end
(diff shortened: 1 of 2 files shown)


app/services/foreman_ansible/fact_parser.rb
```diff
@@ -72,6 +76,11 @@ def ip_from_interface(interface)
       facts[:"ansible_#{interface}"]['ipv4']['address']
     end
 
+    def ipv6_from_interface(interface)
+      return if facts[:"ansible_#{interface}"]['ipv6'].blank?
```
ares: this will now break on rubocop, new line is required after return

UXabre: Hi @ares , thank you for the review! Will do the changes later today

ares: thanks for letting me know, ok, I'll wait with the new release then

UXabre: Hi @ares, didn't want to be the blocking factor so changed it right away :-)

ares: Thanks! I'll give it a short test tomorrow and will merge and release 2.3 if everything goes right!

app/services/foreman_ansible/fact_parser.rb
```diff
@@ -72,6 +76,12 @@ def ip_from_interface(interface)
...
+
+      facts[:"ansible_#{interface}"]['ipv6'].first['address']
```
ares: somehow I'm not seeing these facts, is there some role I need to add?

ares: oh pls ignore me, looking at wrong interface

ares:
the ipv6 address is not being update on my setup, I'll continue testing tomorrow

UXabre:
Strange, since this is fairly similar to IPv4 address retrieval. Let me know if you didn't find it working, after which I'll re-investigate (but retraced my steps, so far it looks like it should do it). Sorry for the inconvenience. If possible, could you send me the retrieved facts? Perhaps something isn't where I expected it to be :-)

ares:
Sorry, my bad. The IPv6 addresses I had were fe80::* which we ignore during import. After some debugging I realized the fact is passed correctly. For some reason it's still not set to my interfaces but I blame my local IPv6 configuration, from this plugin point of view, everything is correct.

Thanks @UXabre, merging this!

PR #221, opened 2018-11-15
Fixes #25474 - String refresh

bkearney:
This pull request does the following:
1) Refresh the string files
2) Adds a .tx folder to the plugin root to allow the makfile in locale to work
3) Pulls in the latest translatoins from transifex that contain Red Hat translations


tbrisker:
@bkearney https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Translating describes the whole thing, basically the plugin maintainers will need to pull in new translations before cutting a new version.

PR #222, opened 2018-11-15
top_level_ansible_vars not complete

pierew:
Top Level Ansible Vars should include all other vars, since the foreman inventory plugin makes. foreman_location_name or by disabling the prefix 'location_name' foreman at the moment sets foreman.location_name in a own list/dict. so with this change the roles can now be executed by hand on a foreman-server or by foreman itself.

### DIFF
--- a/app/services/foreman_ansible/inventory_creator.rb
+++ b/app/services/foreman_ansible/inventory_creator.rb
@@ -44,6 +44,8 @@ module ForemanAnsible
       }.merge(connection_params(host))
       if Setting['top_level_ansible_vars']
         result = result.merge(host_params(host))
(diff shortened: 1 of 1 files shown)


ares:
[test foreman_ansible]

ares:
While it looks reasonable, won't it conflict with other variables that can be defined via roles? E.g. "ip", "mac", "domain_name", "realm_id", "enabled", "managed", "global_status" seems as quite generic terms. Also it will double the inventory size. Would it be acceptable to have it either in top level xor namespaced?

ares:
should be resolved by #258

ares:
closing per previous comment

PR #225, opened 2018-11-21
Fixes #26161 - Assigning Ansible roles over the API

ofedoren:


lib/foreman_ansible/register.rb
```diff
@@ -8,12 +8,14 @@
     permission :play_roles_on_host,
                { :hosts => [:play_roles, :multiple_play_roles],
                  :'api/v2/hosts' => [:play_roles,
-                                     :multiple_play_roles] },
+                                     :multiple_play_roles,
+                                     :assign_ansible_roles] },
```
ares: this should rather require edit_hosts permission I believe

lib/foreman_ansible/register.rb
```diff
@@ -8,12 +8,14 @@
...
                :resource_type => 'Host'
     permission :play_roles_on_hostgroup,
                { :hostgroups => [:play_roles],
                  :'api/v2/hostgroups' => [:play_roles,
-                                          :multiple_play_roles] },
+                                          :multiple_play_roles,
+                                          :assign_ansible_roles] },
```
ares: this should rather require edit_hostgroups permission I believe

ofedoren:
@ares, changed as you suggested.

app/controllers/foreman_ansible/concerns/api_common.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module ForemanAnsible
+  module Concerns
+    # Shared task methods between api controllers
+    module ApiCommon
+      extend ActiveSupport::Concern
+
+      def find_ansible_roles
+        ids = params.fetch(:ansible_role_ids, []) || []
+        @ansible_roles = []
```
ares: This way is not very efficient, it loads each role with separate SQL query, also the current way does not authorize resources. How about this single line?

```
AnsibleRole.authorized(:view_ansible_roles).find(ids)
```

it also raises ActiveRecord::RecordNotFound in case some of the id does not exist, so the rescue part could remain the same

ofedoren:
@ares, updated with your suggestions, thanks!

ares:
Test failures are now related, though I don't fully understand why it complains about ansible variable update route now. Also it would be great if you could open and use separate redmine issue for this, the original ticket is linked to foreman-ansible 2.2.10 release already.

ares:
so tests failures are unrelated, sorry for confusion, please just use separate redmine issue and it's good to be merged

ares:
I know you are off this week and I didn't want to wait any longer so I updated the ticket and merged. Thanks @ofedoren!

PR #228, opened 2018-12-18
Fixes #25726 - Show actual variable types

xprazak2:


app/views/ansible_variables/index.html.erb
```diff
@@ -26,7 +26,7 @@
           variable.ansible_role.name,
           hash_for_ansible_variables_path(:search => "ansible_role = #{variable.ansible_role}")
         ) %></td>
-        <td class="ellipsis"><%= "String" %></td>
+        <td class="ellipsis"><%= variable.key_type || _('string') %></td>
```
ares: should `string` be really translated? if key_type was string, it wouldn't be translated, correct?

xprazak2: Correct, I removed the translation.

app/views/ansible_variables/index.html.erb
```diff
@@ -26,7 +26,7 @@
           variable.ansible_role.name,
           hash_for_ansible_variables_path(:search => "ansible_role = #{variable.ansible_role}")
         ) %></td>
-        <td class="ellipsis"><%= "String" %></td>
+        <td class="ellipsis"><%= variable.key_type || 'String' %></td>
```
ares: so now we're getting String vs string, could you please change the default to `string`?

PR #229, opened 2018-12-20
Fixes #25741 - remove leftover requirement

ares:


mmoll:
test failures related :broken_heart:

adamruzicka:
Test failures are related. In `foreman-tasks` we setup the settings for `*Core` plugins in an initializer[1]. By moving the `require 'foreman_ansible_core` into `test_plugin_helper.rb` the settings dont get set up correctly (`test_plugin_helper` being loaded after the initializer was run?)

[1] - https://github.com/theforeman/foreman-tasks/blob/master/lib/foreman_tasks/engine.rb#L134

ares:
modified so we require only in test env, :crossed_fingers:

adamruzicka:
The `should fetch` failures appear in other PRs too so those are most likely unrealted, not 100% sure if the `should create with recurrence` falls in the same boat

ares:
let's see [test foreman_ansible]

ares:
`Api::V2::AnsibleRolesControllerTest.test_0003_should fetch` is unrelated as it can be observed in other PRs too, e.g. https://github.com/theforeman/foreman_ansible/pull/229

ares:
Thanks for taking a look, merging to unblock 2.3.1 release.

PR #230, opened 2019-01-02
Fix rubocop offence in ansible_runner

xprazak2:


lib/foreman_ansible_core/remote_execution_core/ansible_runner.rb
```diff
@@ -54,7 +54,8 @@ def unknown_host_key_fingerprint?
         last_output = @continuous_output.raw_outputs.last
         return if last_output.nil?
         last_output_raw = last_output['output']
-        return last_output_raw && last_output_raw.lines.last == 'Are you sure you want to continue connecting (yes/no)? '
+        last_line = 'Are you sure you want to continue connecting (yes/no)? '
```
adamruzicka: How about making this a constant and freezing it?

PR #231, opened 2019-01-02
Fixes #25764 - add ansible_sudo_pass during inventory creation

snagoor:
Without this setting, Ansible Jobs would fail to execute in an environment, where NOPASSWD and SSH Keys are not allowed.


adamruzicka:
But the tests need a bit of stubbing

ares:
[test foreman_ansible]

PR #232, opened 2019-01-03
Fixes #25765 - Show hostname in job output

adamruzicka:
Even if remote_execution_connect_by_ip is enabled

Requires:
- [ ] https://github.com/theforeman/foreman_remote_execution/pull/402


adamruzicka:
Before:
before

After:
after

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -20,7 +20,7 @@ def initialize(hosts, template_invocation)
     # with all hosts.
     def to_hash
       hosts = @hosts.map do |h|
-        RemoteExecutionProvider.find_ip_or_hostname(h)
+        AnsibleProvider.find_ip_or_hostname(h)
```
iNecas: Does it make sense to call it `find_ip_or_hostname` if we never expect `ip` in here? I would suggest calling `find_fqdn` directly

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -31,7 +31,7 @@ def to_hash
     def hosts_vars
       hosts.reduce({}) do |hash, host|
         hash.update(
-          RemoteExecutionProvider.find_ip_or_hostname(host) => host_vars(host)
+          AnsibleProvider.find_ip_or_hostname(host) => host_vars(host)
```
iNecas: Same here: if we don't expect ip address to be here, let's call `find_fqdn` instead.

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -95,7 +95,8 @@ def remote_execution_options(host)
         'ansible_user' => host_setting(host, 'remote_execution_ssh_user'),
         'ansible_ssh_pass' => rex_ssh_password(host),
         'ansible_ssh_private_key_file' => ansible_or_rex_ssh_private_key(host),
-        'ansible_port' => host_setting(host, 'remote_execution_ssh_port')
+        'ansible_port' => host_setting(host, 'remote_execution_ssh_port'),
+        'ansible_host' => RemoteExecutionProvider.find_ip_or_hostname(host)
```
iNecas: If we use `find_fqdn` in other places, we don't have to override `find_ip` in `AnsibleProvider` and we could call `AnsibleProvider.find_ip_or_hostname(host)`, which would be less surprising for me to see.

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -126,6 +127,10 @@ def ansible_or_rex_ssh_private_key(host)
       end
     end
 
+    def find_ip_or_hostname(host)
```
iNecas: Is this used anywhere?

app/services/foreman_ansible/inventory_creator.rb
adamruzicka: Calling `find_ip_or_hostname` will raise an exception in case the host doesn't have fqdn associated with any of its interfaces. Not sure if that can even happen. We might simplify this even further and just use `host.name` instead

iNecas:
Btw. tested and works nicely. It seems it also fixes issue with 

```
Unprocessable entity ConfigReport (id: new):
 |   Report time has already been taken
```

which I was hitting without this set of patches.

app/services/foreman_ansible/inventory_creator.rb
iNecas: I would be ok with that

ares:
we'll need to bump rex dependency too

adamruzicka:
This doesn't need the REX thing anymore, since we're not overriding the `#find_ip` method on the `AnsibleProvider` anymore

adamruzicka:
Great :tada:

ares:
[test foreman_ansible]

PR #234, opened 2019-01-04
Update versions in readme

iNecas:
The matrix table was updated according to what was available in yum repos for corresponding versions.


iNecas:
@ares anything against merging this?

ares:
Not at all, thanks @iNecas, merging

PR #237, opened 2019-01-18
Fixes #25455 - Add variables into host_params

xprazak2:
This puts values for Ansible variables from Classification's `values_hash` into `host_params`.

foreman_scap_client_{port,server} come from oprazak.foreman_scap_client:

### DIFF
--- a/app/models/concerns/foreman_ansible/host_managed_extensions.rb
+++ b/app/models/concerns/foreman_ansible/host_managed_extensions.rb
@@ -4,51 +4,70 @@ require 'ipaddress'
 module ForemanAnsible
   # Relations to make Host::Managed 'have' ansible roles
   module HostManagedExtensions
-    extend ActiveSupport::Concern
-    include ::ForemanAnsible::Concerns::JobInvocationHelper
-
     # rubocop:disable Metrics/BlockLength
-    included do
-      has_many :host_ansible_roles, :foreign_key => :host_id
-      has_many :ansible_roles, :through => :host_ansible_roles,
-                               :dependent => :destroy
-      scoped_search :relation => :ansible_roles, :on => :name,
-                    :complete_value => true, :rename => :role,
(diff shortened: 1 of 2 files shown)


ares:
seems tests failures are related, it's not just parser errors

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -4,51 +4,69 @@
...
+    end
+    # rubocop:enable Metrics/AbcSize
 
-      def all_ansible_roles
-        (ansible_roles + inherited_ansible_roles).uniq
+    def all_ansible_roles
+      (ansible_roles + inherited_ansible_roles).uniq
+    end
+
+    def host_params_hash
```
ares: Would this be possible to implement using [Host Info providers](https://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Host-info-providers)?

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -4,51 +4,69 @@
...
+      values = AnsibleVariable.where(:ansible_role_id => all_ansible_roles.pluck(:id)).
+               values_hash(self).raw.values
+
+      transformed = values.each_with_object({}) do |item, memo|
+        item.map do |name, hash|
+          memo[name] = { :value => hash[:value], :safe_value => hash[:value], :source => 'global' }
```
ares: should the source link the specific ansible variable? I suppose that would mean introducing new Parameter type from this plugin though, so global may be ok here for now

safe_value though is more interesting, this means we'll always display the value right? even if the ansible variable is marked as hidden, we should probably use https://github.com/theforeman/foreman/blob/develop/app/models/concerns/hidden_value.rb here

PR #238, opened 2019-01-22
Fix fact parser test

xprazak2:
Lets see if this fixes the failing parser tests...

PR #239, opened 2019-01-25
Make deb OS facts parsing more reliable

ares:
@xprazak2 this should be more robust way for https://github.com/theforeman/foreman_ansible/pull/238 that should also support ubuntu, mind to take a look? Instead of hardcoding the mapping, we do what core does for puppet, looking at lsb info that ansible setup modules provide under ansible_distribution_release

### DIFF
--- a/test/unit/services/fact_parser_test.rb
+++ b/test/unit/services/fact_parser_test.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 require 'test_plugin_helper'
 
 module ForemanAnsible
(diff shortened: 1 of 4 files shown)


xprazak2:
[test foreman_ansible]

test/unit/services/fact_parser_test.rb
```diff
@@ -126,7 +132,7 @@ class DebianFactParserTest < ActiveSupport::TestCase
       as_admin do
         os = @facts_parser.operatingsystem
 
-        assert_equal '10', os.major
+        assert_equal '8', os.major
```
xprazak2: Maybe we should change the test name as well - test no longer tries to parse facts for debian unstable.

xprazak2:
Looking ok, just minor things + rubocop again :(

ares:
New version pushed

xprazak2:
Code related to fact parsing looks good.

ares:
rubocop is now happy, the failing tests seems to be core test for search bar, which is sad but unrelated, thanks for review @xprazak2, merging

PR #241, opened 2019-02-04
Fixes #25977 - Use host info provider for variables

xprazak2:


xprazak2:
This should also handle nested values for variables correctly.

ares:
Now the output ends in root level, e.g.
```
foreman_ansible:
  a: tada
parameters:
  kt_env: 
  kt_cv: 
```
which is no longer valid ENC output for puppet. Previously it was merged under parameters which I believe end in the ansible inventory.

```
parameters:
  kt_env: 
  kt_cv: 
  a: tada
```

So I suppose we need to wrap it in parameters? The `foreman_ansible` prefix should be added based on settings.

xprazak2:
Updated, I added a new setting that determines whether to add namespace to ENC. It has no impact on variables in inventory file.

xprazak2:
Updated, I removed the setting for namespace in ENC.

ares:
test failures seems related

ares:
Sorry I meant unrelated, it only needs to be rebased on current master I guess.

So with this patch I think the behavior changed (but I think in a good direction), we now add all ansible variables to inventory and ENC, their default value is used if no matcher is matching. That is good but user has no way to keep default for ansible, we're missing "override" checkbox we have in smart class parameters.

I would be fine with merging this as is, if you could also add "override" flag in separate PR. Based on today lunch discussion, it shouldn't be too complicated.

xprazak2:
I pushed a second commit that allows us to use `override` and `omit` with ansible variables.

Then I realized omit will need additional changes.

xprazak2:
`omit` will not work without additional changes in core, but having `override` should be enough for now.

I also disabled editing of the ansible variable name in form, I am not sure if it was enabled intentionally.

app/services/foreman_ansible/ansible_info.rb
```diff
@@ -10,7 +10,7 @@ def ansible_params
 
       variables.each_with_object({}) do |var, memo|
         value = values[var]
-        memo[var.key] = value if value
+        memo[var.key] = value if value && var.override
```
ares: could we use override few lines above when we search for AnsibleVariables? or do we need to load all in order to correctly compute values_has?

app/views/ansible_variables/_fields.erb
```diff
@@ -0,0 +1,67 @@
...
+    <%= text_f f, :key, :disabled => true, :size => "col-md-8" %>
+    <%= f.hidden_field :key %>
+    <%= textarea_f f, :description, :rows => :auto, :size => "col-md-8", :class => "no-stretch" %>
+    <%= select_f f, :ansible_role_id, [f.object.ansible_role], :id, :to_label, {}, :label => _("Ansible Role"), :disabled => true %>
+  </fieldset>
+  <fieldset>
+    <h2><%= _("Default Behavior") %></h2>
+    <h6><%= _("Override the default value of the Ansible variable.") %></h6>
+    <%= checkbox_f(f, :override, :onchange => 'toggleOverrideValue(this)', :size => "col-md-8",
+                   :label_help => _('Whether the Ansible variable value is managed by Foreman.')
```
ares: Could we describe this a bit more? Something like:

"Mark the variable to be managed by Foreman. When the Ansible role of this variable is assigned to a host, the default value will be added to Ansible inventory as a host variable. To set a different value for such variable, specify matchers."

ares:
oh actually one more thing, we should highlight the overriden somehow in the table, either a new column or the flag we use for puppet

xprazak2:
I search only for overriden variables, changed the label text and added flags:

overriden-vars

san7ket:
@xprazak2 I don't know what I am missing, but the ansible variables are all empty for me, I mean there is no default value set in them. Could you point me in right direction please

ares:
@san7ket we don't import the default value, once you import the variable, you need to go to its edit form and check "override". Once you assign this variable's role to the host, the configured default value is set. They can be overridden using matchers at the bottom of the form. Hope that helps

ares:
not sure if this is caused by the parameter types introduced in core or some change here, but the variable type selector is now visually broken 
vars

xprazak2:
Yes, caused by changes in core, should be fixed now. 

I also opened https://github.com/theforeman/foreman/pull/6506 to enable/disable type field based on override checkbox.

ares:
[test foreman_ansible]

ares:
[test foreman_ansible]

ares:
triggered manually, the comment does not seem to work http://ci.theforeman.org/job/foreman_ansible-pull-request/287/

ares:
[test foreman_ansible]

ares:
APJ

xprazak2:
```
rake aborted!
ActiveRecord::AdapterNotSpecified: 'production' database is not configured. Available: ["development", "test"]

```

ares:
Failures on sqlite does not seem related to this PR, thanks @xprazak2, merging!

PR #242, opened 2019-02-05
Fixes #25989 - move play roles button

ares:
before
before

after
after


xprazak2:
Works well.

ares:
Tests failures unrelated, thanks for review, merging

PR #243, opened 2019-02-07
Use absolute_import for foreman_callback plugin

Akasurde:
Ansible plugin directory contains `json` callback plugin next to
foreman_callback plugin. Without absolute import foreman_callback
plugin fails with json errors (which occurs due to namespace collision).

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/extras/foreman_callback.py
+++ b/extras/foreman_callback.py
@@ -15,6 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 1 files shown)


tbrisker:
ok to test

adamruzicka:
We removed the callback from our repo in https://github.com/theforeman/foreman_ansible/pull/358 , so I'll go ahead and close this. We now favor the callback from [foreman ansible modules](https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/callback/foreman.py) where the change you were proposing is already present. Thank you @Akasurde

PR #244, opened 2019-02-17
Fixes #26064 - add missing search route

ares:

PR #245, opened 2019-02-18
Fix rubocop offenses

tbrisker:


xprazak2:
Test failures caused by missing route, which is fixed by #244.

Thanks!

PR #246, opened 2019-02-20
Fixes #26093 - restore searching by role

ares:


ares:
the issue is only reproducible with katello, which adds "role" for searching by system purpose

xprazak2:
Fixes the issue for me, test failures unrelated, should be fixed in #247

ares:
[test foreman_ansible] should be green now

xprazak2:
Maybe rebase on top of master with the fix for missing permissions will help.

ares:
yeah, just rebased

xprazak2:
The only failing test is AuditExtensions, I do not think it is related.

xprazak2:
Thanks!

PR #249, opened 2019-03-01
Fixes #26202 - Uses hardware MAC-address whenever possible

UXabre:
I've noticed that, when a Bonded interface has been created, the facts seem to not take into consideration the fact that the interfaces take over the MAC-address of the bonded interface; however, ansible supports this scenario by adding a special perm_macaddress field which holds the hardware mac address.

Using this hardware mac address allows us to update facts for the individual interfaces. If we don't do this, however, we wind up with multiple NICs with the same name (which foreman does not like)

### DIFF
--- a/test/unit/services/fact_parser_test.rb
+++ b/test/unit/services/fact_parser_test.rb
@@ -71,13 +71,12 @@ module ForemanAnsible
           '_timestamp' => '2018-10-29 20:01:51 +0100',
           'ansible_facts' =>
           {
-            'ansible_interfaces' => [
-                'eth0'
-            ],
+            'ansible_interfaces' => %w(eth0 eth1 eth2 bond0),
             'ansible_eth0' => {
               'active' => true,
               'device' => 'eth0',
-              "macaddress" => '52:54:00:04:55:37',
+              'macaddress' => '52:54:00:04:55:37',
+              'perm_macaddress' => '52:54:00:04:55:37',
               'ipv4' => {
                   'address' => '10.10.0.10',
                   'netmask' => '255.255.0.0',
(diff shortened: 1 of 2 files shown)


UXabre:
could somebody check out this PR please? :-)

ares:
Looks reasonable, I won't be able to test with real bonds setup, but I guess if we don't see any regressions, we can merge without testing.

But first, please fix the rubocop issue

```
Inspecting 103 files
......................................................................................................C

Offenses:

/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/test/unit/services/fact_parser_test.rb:74:37: C: Style/WordArray: Use %w or %W for an array of words.
            'ansible_interfaces' => [ ...
                                    ^
/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/test/unit/services/fact_parser_test.rb:137:27: C: Style/WordArray: Use %w or %W for an array of words.
              'slaves' => [ ...
                          ^
```

UXabre:
Rubocop issues should be fixed now, thanks for pointing these out :-)

UXabre:
Could this be re-reviewed please? :-)

ares:
[test foreman_ansible] should be fixed by https://github.com/theforeman/foreman_ansible/pull/274

ares:
[test foreman_ansible]

UXabre:
Hi, 
I've update my ansible plugin on foreman to v3.0.5; which includes the mentioned fix; however, I still have the same issue that uploading facts of a bonded interface, where both the bond and the primary interface within my bond, share the same mac address, results in my primary interface being renamed to the secondary interface name.

To elaborate this example, this is what I see in foreman before running ansible:
Bond0, with MAC A
Eth0 with MAC A
Eth1 with MAC B

After Uploading:
Bond 0 with MAC A
Eth1 with MAC A
Eth1 with MAC B

UXabre:
Hi, I had to rebase since there was a conflict.
I also noticed that a few (seemingly?) unrelated tests seem to fail as of now?
If someone could be so kind to review this and let me know the next steps for this PR

test/unit/services/fact_parser_test.rb
```diff
@@ -93,6 +92,46 @@ class NetworkFactParserTest < ActiveSupport::TestCase
...
+                  {
+                      'address' => 'fd00::5054:00ff:fe04:5538',
+                      'prefix' => '64',
+                      'scope' => 'host'
+                  }
+              ],
+              'mtu' => 1500,
+              'promisc' => false,
+              'slaves' => %w(eth1 eth2)
+              ],
```
ares: the lcosing square bracket here is unexpected

UXabre: Oh, must've been a merge error probably. I'll fix it and run tests on my system before pushing again. Terribly sorry for this!

UXabre:
Hi, I've tested this on my machine; all tests seem to pass so far and the syntax error is fixed as well.

ares:
Thank you @UXabre, ran tests locally and can confirm they are green. Merging this, once again, sorry for the delay.

UXabre:
Thanks @ares ! As always, a pleasure doing business with you guys :-) Don't worry about the delay, I was more worried that I was annoying people instead!

PR #251, opened 2019-03-07
Fixes #26272 - Use cdn_proxy to fetch playbook from Red Hat Insights

hao-yu:


ares:
Thank you very much, I linked the upstream issue to it to track the version.

PR #252, opened 2019-03-07
Fixes #26264, #25721 - Detect variable types on import

xprazak2:


app/controllers/ansible_variables_controller.rb
```diff
@@ -49,11 +50,19 @@ def resource_params
...
-    params[:changed] ? params[:changed][:obsolete].as_json : {}
+    fetch_vars :obsolete
+  end
+
+  def updated_vars
+    fetch_vars :update
+  end
+
+  def fetch_vars(key)
+    params[:changed] ? (params[:changed][key].as_json || {}) : {}
```
ares: since you touch this already, `params.fetch(:changed, {}).fetch(key).try(:as_json) || {}` is I think equivalent and slightly safer (in case `params[:changed]` is not a hash)

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -96,5 +99,25 @@ def local_variables
     def remote_variables
       proxy_api.all_variables
     end
+
+    def infer_key_type(value)
+      {
```
ares: could we extract this to a constant in case we'll need to use it elsewhere too?

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -96,5 +99,25 @@ def local_variables
...
+        TrueClass => 'boolean',
+        FalseClass => 'boolean',
+        Integer => 'integer',
+        Float => 'real',
+        Array => 'array',
+        Hash => 'hash'
+      }.reduce('string') { |memo, (key, type)| value.is_a?(key) && memo == 'string' ? type : memo }
```
ares: I'm probably missing something, but can't we add String => 'string' to the hash and then just do `hash[value.class] || 'string'`?

xprazak2:
```ruby
13.class #=> Fixnum
```

I had that previously, rubocop was complaining I should be using Integer instead of Fixnum. Not sure which one is better, let me know your preference.

ares: as discussed, this would be the pattern
```
{
'Fixnum' => 'integer',
'Integer' => 'integer'
}
h[value.class.to_s] || 'string'
```

xprazak2: I made updates as suggested.

ares:
I think it changes with this PR, when I modify the default value of the variable and reimport variables. It shows up as updated, if I mark that for import, it reset the default value. That is probably undesired behavior.

xprazak2:
I made changes as suggested, default value is no longer updated for the variables that already exist. This means that when users want to update variables using import, they need to delete them in Foreman and newly import.

ares:
rubocop failure

```
.....................................................C................................................C

Offenses:

/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/app/services/foreman_ansible/variables_importer.rb:52:9: C: Style/MultilineIfModifier: Favor a normal if-statement over a modifier clause in a multiline statement.
        variable.assign_attributes(:default_value => variable_default, ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/test/unit/services/ansible_variables_importer_test.rb:15:50: C: Layout/SpaceInsideHashLiteralBraces: Space inside } missing.
      new_role.name => { 'new_var' => 'new value'},
                                                 ^
```

PR #254, opened 2019-03-18
Fixes #26391 - Allow custom Ansible variables

xprazak2:


app/views/ansible_variables/_fields.erb
```diff
@@ -1,10 +1,16 @@
...
+    <%= text_f f, :key, :disabled => f.object.persisted?, :size => "col-md-8" %>
     <%= textarea_f f, :description, :rows => :auto, :size => "col-md-8", :class => "no-stretch" %>
-    <%= select_f f, :ansible_role_id, [f.object.ansible_role], :id, :to_label, {}, :label => _("Ansible Role"), :disabled => true %>
+    <%= ansible_role_select f, f.object.persisted? %>
+    <% is_custom = f.object.new_record? ? true : f.object.ansible_custom %>
+    <%= f.hidden_field :ansible_custom, :value => is_custom %>
+    <%= checkbox_f(f, :ansible_custom,
+                      :label => _('Custom variable'),
+                      :checked => is_custom,
+                      :label_help => _('Whether this variable was created in Foreman as a custom variable rather than through import from role'),
```
ares: `import from smart proxy` instead of `import from role`

app/views/ansible_variables/index.html.erb
```diff
@@ -1,14 +1,16 @@
...
+    documentation_button('#4.3Variables', :root_url => ansible_doc_url) %>
 
 <table class="<%= table_css_classes 'table-fixed' %>">
   <thead>
     <tr>
       <th class='col-md-6'><%= sort :name, :as => s_('Variable|Name') %></th>
       <th class='col-md-2'><%= sort :ansible_role, :as => s_('Variable|Role') %></th>
-      <th class='col-md-2'><%= _('Type') %></th>
+      <th class='col-md-1'><%= _('Type') %></th>
+      <th class='col-md-1'><%= _('Custom variable?') %></th>
```
ares: this is variables table, no need to repeat it here, just say `_('Custom')`

perhaps it would be better to say `Imported`, that is more descriptive, but of course changes the meaning and would need to change on many places, would that be too much to ask? :-)

app/models/ansible_variable.rb
```diff
@@ -12,8 +12,8 @@ def ansible?
     true
   end
 
-  def self.humanize_class_name
-    'Ansible variable'
+  def self.humanize_class_name(name = 'Ansible variable')
```
ares: where do we use this new feature of this method? IIRC this is some rails class which expects 0 arguments. We override it in core for variable lookup key but accepting options. If we need this, should we follow the same pattern? But still, I don't see this parameter being used anywhere, so first I'm trying to understand the use of it.

xprazak2: I was getting wrong number of arguments, but I can no longer reproduce, so I reverted to the original state.

xprazak2:
I changed the new attribute to `imported`. I also no longer set the default db value as it would set value for Puppetclass lookup keys as well.

app/helpers/foreman_ansible/ansible_roles_helper.rb
```diff
@@ -11,6 +11,20 @@ def ansible_proxy_links(hash, classes = nil)
       end.flatten
     end
```
ares: ok, we need the humanize method, the same we have for smart variables would do

```ruby
  def self.humanize_class_name(options = nil)
    if options.present?
      super
    else
      "Ansible variable"
    end
  end
```

ares:
Existing imported ansible variables are no longer set as imported, or at least in the form, the checkbox is not checked, I believe we need to update them, otherwise they would not be removed on role update.

Also I wonder now after more testing, why the imported in the form is actually a checkbox? It's always disabled, why we don't simply show check mark, like in the table? User is not expected to ever change it right?

db/migrate/20190318094437_add_imported_attr_to_ansible_variables.rb
```diff
@@ -0,0 +1,15 @@
+class AddImportedAttrToAnsibleVariables < ActiveRecord::Migration[5.2]
+  def up
+    add_column :lookup_keys, :imported, :boolean
+
+    AnsibleVariable.find_in_batches do |batch|
+      batch.map do |variable|
+        variable.update_attribute :imported, true
```
xprazak2: Migrations should take care of updating the existing variables, it works for me without problems.

xprazak2:
Yes, user will never be able to change the imported checkbox, because it is always disabled. It makes the whole form look consistent.

I changed it to a check icon, but it looks strange to my eyes because we now have 2 different ways of showing boolean values in 1 form and there is another checkbox just a few lines below it:

imported-check

And when variable is not imported, it looks like there is something missing:

impurted-uncheck

xprazak2:
I improved the form and removed the check for `imported` attribute:

xprazak2:
I added a scoped search on `imported`.

ares:
sorry, I'd suggest disabling the cop

```
RuboCop failed!
Running RuboCop...
Inspecting 105 files
................................W........................................................................

Offenses:

/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/app/models/ansible_variable.rb:8:56: W: Lint/BooleanSymbol: Symbol with a boolean name - you probably meant to use true.
  scoped_search :on => :imported, :complete_value => { :true => true, :false => false }
                                                       ^^^^^
/home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/app/models/ansible_variable.rb:8:71: W: Lint/BooleanSymbol: Symbol with a boolean name - you probably meant to use false.
  scoped_search :on => :imported, :complete_value => { :true => true, :false => false }
```

xprazak2:
Adding fix for rubocop.

ares:
Tests are green on one combination, I've seen it working previously. Thanks @xprazak2, merging!

PR #256, opened 2019-03-25
Fixes #26436 - restore compatibility with 1.21

ares:
broken by https://github.com/theforeman/foreman/commit/62f2fd886b35f3543ecea134efec840ccfb36aa6#diff-6c69b0cc93d393f1605b2749fda757bf

### DIFF
--- a/app/views/ansible_variables/_fields.erb
+++ b/app/views/ansible_variables/_fields.erb
@@ -12,7 +12,13 @@
     <%= checkbox_f(f, :override, :onchange => 'toggleOverrideValue(this)', :size => "col-md-8",
                    :label_help => _('Mark the variable to be managed by Foreman. When the Ansible role of this variable is assigned to a host, the default value will be added to Ansible inventory as a host variable. Specify matchers to set a different value for such variable.')
         ) %>
-    <%= param_type_selector(f, false, :onchange => 'keyTypeChange(this)', :disabled => !f.object.override) %>
+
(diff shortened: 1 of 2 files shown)


app/views/ansible_variables/_fields.erb
```diff
@@ -12,7 +12,13 @@
     <%= checkbox_f(f, :override, :onchange => 'toggleOverrideValue(this)', :size => "col-md-8",
                    :label_help => _('Mark the variable to be managed by Foreman. When the Ansible role of this variable is assigned to a host, the default value will be added to Ansible inventory as a host variable. Specify matchers to set a different value for such variable.')
         ) %>
-    <%= param_type_selector(f, false, :onchange => 'keyTypeChange(this)', :disabled => !f.object.override) %>
+
+    <% version = Foreman::Version.new %>
+    <% if version.major.to_i >= 1 && version.minor.to_i >= 22 %>
```
tbrisker: since 1.22 hasn't been released yet, does it make sense keeping the old signature in core?

ares: I don't see helpers as (plugin) API but don't have strong opinion, I don't think other plugins use this selector and I'm fine with the fix here

ares:
@xprazak2 mind to take a look? failure is on the sqlite (too many version in single SQL, not related)

xprazak2:
Thanks!

PR #258, opened 2019-03-28
Fixes #26068 - Improve inventory structure

xprazak2:


ares:
Few comments after a bit broader discussions:
* let's allow both GET and POST
* let's move API request to separate PR, we need to test them with ansible inventory script
* the `foreman:` shouldn't contain the whole ENC but just content of `parameters`, not puppet env or puppet classes
* ansible variables should win over parameters

xprazak2:
* I removed the API endpoints
* `foreman` key contains only content of `parameters`
* ansible variabales winning over host params

test-inv

ares:
Thanks @xprazak2, merging!

PR #260, opened 2019-04-17
Fixes #25722 - Redirect to index when no changes in vars

xprazak2:


ares:
tests need to be updated, it now changes response to 302, perhaps two tests would be now good, one for empty second for some roles reported by proxy

xprazak2:
I added tests.

PR #261, opened 2019-04-17
Fixes #22804 - Parse origin for failed runs

xprazak2:


xprazak2:
Needs https://github.com/ansible/ansible/pull/55437

app/services/foreman_ansible/ansible_report_scanner.rb
```diff
@@ -15,8 +15,7 @@ def scan(report, logs)
       def ansible_report?(logs)
         return false if logs.blank?
         logs.any? do |log|
-          log['log'].fetch('messages', {}).
-            fetch('message', '') =~ /"_ansible_parsed"/
+          log['log'].fetch('reporter', nil) == 'ansible'
```
adamruzicka: Isn't `a_hash.fetch('reporter', nil)` essentialy equivalent to `a_hash['reporter']`?

xprazak2: Yeah, updated.

adamruzicka:
Looks it should work with the linked ansible changes, will need to test once the ansible changes land.

ares:
I raised a question in ansible PR, I think it would be better to add origin as top-level attribute than have it as part of each and every log

xprazak2:
It would be, but report scanners do not have access to the top-level report attributes.

sbernhard:
Any update on this PR?

ezr-ondrej:
> Any update on this PR?

The ansible part still didn't make it in.

xprazak2:
Now needs https://github.com/theforeman/foreman-ansible-modules/pull/896 and https://github.com/theforeman/foreman/pull/7819

webpack/components/ReportJsonViewer.js
```diff
@@ -10,7 +10,7 @@ const theme = {
 
 const ReportJsonViewer = ({ data }) => (
   <div className="report-json-viewer">
-    <JSONTree data={this.props.data} hideRoot theme={theme} />
+    <JSONTree data={data} hideRoot theme={theme} />
```
ezr-ondrej, line 13: is this fix related?

app/services/foreman_ansible/ansible_report_scanner.rb
```diff
@@ -5,19 +5,14 @@ module ForemanAnsible
...
     class << self
-      def scan(report, logs)
-        if (is_ansible = ansible_report?(logs))
-          report.origin = 'Ansible'
-        end
-        is_ansible
+      def add_reporter_data(report, raw); end
+
+      def identify_origin(raw)
+        raw['reporter']
```
ezr-ondrej, line 11: This could accidently return something else, in case another config mgmt tool would use this field, right?
Following seems bit safer:
```suggestion
        'Ansible' if ansible_report?(raw)
```

webpack/components/ReportJsonViewer.js
xprazak2: Yes, report is not displayed properly without it:

data-without

data-within

app/services/foreman_ansible/ansible_report_scanner.rb
xprazak2: Updated.

webpack/components/ReportJsonViewer.js
ezr-ondrej: Thanks for explanation :+1:

ezr-ondrej:
I've fixed the fixture for this, hopefully the tests are going to pass.

app/services/foreman_ansible/ansible_report_scanner.rb
```diff
@@ -5,19 +5,14 @@ module ForemanAnsible
...
       end
 
-      def ansible_report?(logs)
-        return false if logs.blank?
-        logs.any? do |log|
-          log['log'].fetch('messages', {}).
-            fetch('message', '') =~ /"_ansible_parsed"/
-        end
+      def ansible_report?(raw)
+        raw['reporter'] == 'ansible'
```
ezr-ondrej, line 15: Should we keep the backward compatibility?
By renaming the old method to `ansible_27_report?` and
```suggestion
        raw['reporter'] == 'ansible' || ansible_27_report?(raw['logs'] || [])
```

ezr-ondrej: And we should then have two fixtures and two tests for it I'd guess :thinking:

xprazak2:
I kept the old version of method as well.

PR #262, opened 2019-04-18
Fixes #26651 - Mark switcher for remount after AJAX

xprazak2:


xprazak2:
needs https://github.com/theforeman/foreman/pull/6699

ezr-ondrej:
Hi, to be less destructive and more supportive, I have came up with two opposing solutions - both needs bit of cleansing, both solves the issue though.

One in #264, it is not really nice solution, but I believe it's bit better than mounting (and thus rendering) the components twice.

**Or**

If you would like to take a look, there is a web components solution, which works as well ( :eyes: :heart: )

- In core enables web components for old browser and in new ones for babel compiled code (our case): https://github.com/ezr-ondrej/foreman/commit/7c899ddb1f5b6d72f64c7191c61f2093c448ccd7
- And defines `<ansible-roles>` component and uses it: https://github.com/ezr-ondrej/foreman_ansible/commit/faa24a7d494f33261976a3ead76575a35bde8296

xprazak2:
Closing in favour of #264.

PR #263, opened 2019-04-24
Fixes #26670 - Change node packages to be compatible with develop

xprazak2:


ekohl:
Not sure how this could have caused the test failures, but python is included when the ansible plugin is present so it may be relevant.

PR #264, opened 2019-04-26
Fixes #26645 - delay React mount after jquery inserts DOM

ezr-ondrej:
Solution proposal to be constructive, not just destructive on #262.
I have investigated a bit and it is not really easy to do, I believe the `Vue.js` is solving it by integrating itself into https://github.com/w3c/webcomponents

In this case it means just wait a bit before mounting. This is proposing a solution.

### DIFF
--- a/app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
+++ b/app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
@@ -2,14 +2,22 @@
 <%= webpacked_plugins_css_for :foreman_ansible %>
 
 <div class='tab-pane' id='ansible_roles'>
-  <div id='ansible_roles_switcher'></div>
   <% roles = f.object.is_a?(Hostgroup) ? roles_attrs(f.object.inherited_and_own_ansible_roles) : roles_attrs(f.object.all_ansible_roles) %>
   <% class_name = f.object.is_a?(Hostgroup) ? 'Hostgroup' : 'Host' %>
-  <%= mount_react_component('AnsibleRolesSwitcher', '#ansible_roles_switcher', { :initialAssignedRoles => roles,
-                                                                                 :inheritedRoleIds => f.object.inherited_ansible_roles.map(&:id),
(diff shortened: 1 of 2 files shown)


xprazak2:
Works nicely. I would go with this solution for the moment as it does not need any changes in core. We can switch to web components when we have support them in core.

@ezr-ondrej, could you change ticket number to [26645](https://projects.theforeman.org/issues/26645) please?

ezr-ondrej:
Gladly, just as it was a proposal I believe we should solve two things:

- would be nice to clean up to separate file
- not sure about the double event registering, maybe we could handle it in one somehow.

ezr-ondrej:
Okey, I have solved the second, and that shrunk the code in `index.js` so I believe it is not necessary to create new file for one function? not sure there :)

xprazak2:
:+1: I think you can keep it as it is, test failures unrelated

any concerns @ares, @sharvit?

ezr-ondrej:
- Ad 1. it is custom event which we are using to signalize "almost full page reloaded" so I believe it is only for usecase on host create page, though the concern in general makes sense, I think we are on a safe side with this one (as long as we do not add more triggers of this event) :)
- Ad 2. I believe it should not be implemented any more and considered as a workaround. So merging should depend on how badly we want to get rid of the bug :)

I agree we should aim to get all the data necessary for rendering the page through a graphQL, but I do not believe it is happening any time soon, so we will have to find a solution how to do it for server rendered html ajax calls as well.

POC of web components - brief blog post with kinda POC:
https://community.theforeman.org/t/react-components-mounted-by-web-components/13767
If you can get involved, it will be lovely :heart:

ares:
Do I read that correctly it's 'blinking on partial page reload on hidden tab' vs 'the switcher hides on ajax request and user needs to reload host form'? I'm fine with hack fix for now if you guys continue working on proper fix.

ezr-ondrej:
Yeah, you should not even see that blink, though you are not on that tab.
I believe @sharvit is worried that the event could be fired after another content load (what is understandable as it's name is very misleading), but it is actually fired only if the whole host form is re-rendered.

PR #268, opened 2019-05-13
Fixes #26786 - fix reverted routes

ezr-ondrej:
I have managed to spin up the 1.21 and test it, but it had to be in a clean environment, so it would deserve another spin.
Fixes the root problem though.


ares:
Thanks @ezr-ondrej, merging!

PR #273, opened 2019-05-25
add config/as_deprecation_whitelist.yaml

mmoll:


xprazak2:
The test failures might be relevant. Why do we need the whitelist file if it is empty?

mmoll:
@xprazak2 I'll put out post to discourse in a few days... To fix the current master test failues, new gem releases of -tasks and _rex would be needed anyway, and https://projects.theforeman.org/issues/26882 the next one.

mmoll:
[test foreman_ansible]

PR #274, opened 2019-05-25
Fixes #26881 - use dynamic attrs for Factory Bot 5.x compat

mmoll:

PR #275, opened 2019-06-10
Fixes #27007 - Fix passing command to shell module

adamruzicka:
Ansible < 2.8.0 parses the command in a weird way and adds a
single space at the beginning of each line. This breaks scripts
which contain heredocs.

For details see: https://github.com/ansible/ansible/issues/39137#issuecomment-383453939

### DIFF
--- a/app/views/foreman_ansible/job_templates/run_command_-_ansible_default.erb
+++ b/app/views/foreman_ansible/job_templates/run_command_-_ansible_default.erb
@@ -17,7 +17,8 @@ model: JobTemplate
 ---
 - hosts: all
   tasks:
-    - shell: |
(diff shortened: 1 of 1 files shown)


ares:
[test foreman_ansible] should be fixed by https://github.com/theforeman/foreman_ansible/pull/274

ares:
[test foreman_ansible]

ares:
Thanks @adamruzicka, could you please also open the same patch against https://github.com/theforeman/community-templates/blob/develop/job_templates/run_command_-_ansible_default.erb where we sync templates from? Merging meanwhile so it gets into 3.0.3

PR #279, opened 2019-06-30
Fixes #28073 - parse boot time fact

ares:
needs theforeman/foreman#6843 but does not hurt to merge earlier

### DIFF
--- a/app/services/foreman_ansible/fact_parser.rb
+++ b/app/services/foreman_ansible/fact_parser.rb
@@ -68,6 +68,10 @@ module ForemanAnsible
 
     def ipmi_interface; end
 
+    def boot_timestamp
+      Time.zone.now.to_i - facts['ansible_uptime_seconds'].to_i
+    end
+
     private
 
     def ansible_interfaces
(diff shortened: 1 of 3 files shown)


app/services/foreman_ansible/fact_parser.rb
```diff
@@ -64,6 +64,10 @@ def get_facts_for_interface(interface)
 
     def ipmi_interface; end
 
+    def boot_time
+      Time.zone.now.to_i - facts['ansible_uptime_seconds'].to_i
```
adamruzicka: Are we ok with our idea of boot time being off by a couple of seconds (or minutes, depending on the how long the job run)?

ares: That is the case for every facts parser, it will differ by the time facter was ran and we parsed it on satellite side. It was the case for uptime as well which even got older faster.

ares:
@ezr-ondrej I belive you tested this recently, perhaps you could ack :-) first commit fixes rubocop in this repo

ares:
@ezr-ondrej this should now turn green and should get to 4.0, can you take a look pls? it's a small thing

adamruzicka:
Test failures seem unrelated, merging

adamruzicka:
Thanks you @ares !

PR #280, opened 2019-07-04
Fixes #27236 - Send verbosity level to the smart proxy

adamruzicka:


app/lib/actions/foreman_ansible/helpers/host_common.rb
```diff
@@ -29,10 +29,6 @@ def fill_continuous_output(continuous_output)
         rescue StandardError => e
           continuous_output.add_exception(_('Error loading data from proxy'), e)
         end
-
-        def find_options
```
adamruzicka: The entire module seems like a leftover from when foreman ansible implemented its own actions. A quick grep didn't find it being used anywhere. How about I drop it entirely?

adamruzicka:
[test foreman_ansible]

ares:
[test foreman_ansible] rekick

adamruzicka:
[test foreman_ansible] one more time. _Daft Punk starts playing in the background_

ares:
Thanks @adamruzicka, merging, the failures are caused by templates snapshots.

PR #283, opened 2019-07-18
Parameterized roles in ENC (Feature #27285)

avylove:
This pull request adds a new setting, ansible_parameterized_roles_in_enc, which defaults to false. When this setting is enabled, the output of  /api/hosts/id:/enc will list roles and role variables hierarchically. Something like.

```
"parameters": {
  "domainname": "example.com",
  ...
}
"roles": {
  "role1": {
    "foo": "bar" 
  }
}
```

This will allow for richer API interaction, particularly in cases where playbooks are not pushes from the Foreman server.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -90,6 +90,13 @@ class Setting
               :collection => lambda do
                 Hash[%w[ansible-playbook ansible-runner].map { |x| [x, x] }]
               end
+            ),
+            set(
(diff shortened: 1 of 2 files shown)


avylove:
Closing and will resubmit

ares:
is there a reason why we'd like to avoid adding this info to ENC? and could you please describe a bit more your usecase? the only downside I see may be the performance, when we generate ENC for many hosts

PR #284, opened 2019-07-18
Fixes #27285 - Parameterize roles in ENC

avylove:
This pull request adds a new setting, **ansible_parameterized_roles_in_enc**, which defaults to `false`. When this setting is `true`, the output of **/api/hosts/id:/enc** will list roles and role variables hierarchically. Something like.

```
"parameters": {
  "domainname": "example.com",
  ...
}
"roles": {
  "role1": {
    "foo": "bar" 
  }
}
```

This will allow for richer API interaction, particularly in cases where playbooks are not pushes from the Foreman server and mirrors a similar capability provided by Foreman for Puppet.

### DIFF
--- a/app/services/foreman_ansible/ansible_info.rb
+++ b/app/services/foreman_ansible/ansible_info.rb
@@ -1,7 +1,11 @@
 module ForemanAnsible
   class AnsibleInfo < ::HostInfo::Provider
     def host_info
-      { 'parameters' => ansible_params }
+      if Setting[:ansible_parameterized_roles_in_enc]
+        { 'roles' => ansible_roles }
+      else
+        { 'parameters' => ansible_params }
+      end
     end
 
     def ansible_params
@@ -12,6 +16,20 @@ module ForemanAnsible
         value = values[var]
         memo[var.key] = value if value
         memo
+      end
+    end
+
+    def ansible_roles
+      roles = host.all_ansible_roles.pluck(:id, :name)
+
+      roles.each_with_object({}) do |role, rmemo|
+        variables = AnsibleVariable.where(:ansible_role_id => role[0], :override => true)
+        values = variables.values_hash(host)
+
+        rmemo[role[1]] = variables.each_with_object({}) do |var, memo|
+          value = values[var]
+          memo[var.key] = value if value
+        end
       end
     end
   end
--- a/app/services/foreman_ansible/ansible_info.rb
+++ b/app/services/foreman_ansible/ansible_info.rb
@@ -19,4 +19,3 @@ module ForemanAnsible
     end
   end
 end
-
--- a/app/services/foreman_ansible/ansible_info.rb
+++ b/app/services/foreman_ansible/ansible_info.rb
@@ -1,22 +1,7 @@
 module ForemanAnsible
   class AnsibleInfo < ::HostInfo::Provider
     def host_info
-      if Setting[:ansible_parameterized_roles_in_enc]
-        { 'roles' => ansible_roles }
-      else
-        { 'parameters' => ansible_params }
-      end
-    end
-
-    def ansible_params
-      variables = AnsibleVariable.where(:ansible_role_id => host.all_ansible_roles.pluck(:id), :override => true)
-      values = variables.values_hash(host)
-
-      variables.each_with_object({}) do |var, memo|
-        value = values[var]
-        memo[var.key] = value if value
-        memo
-      end
+      { 'ansible_roles' => ansible_roles }
     end
 
     def ansible_roles
@@ -34,3 +19,4 @@ module ForemanAnsible
     end
   end
 end
+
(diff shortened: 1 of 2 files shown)


app/services/foreman_ansible/ansible_info.rb
```diff
@@ -1,7 +1,11 @@
 module ForemanAnsible
   class AnsibleInfo < ::HostInfo::Provider
     def host_info
-      { 'parameters' => ansible_params }
+      if Setting[:ansible_parameterized_roles_in_enc]
+        { 'roles' => ansible_roles }
```
xprazak2: I would change the key to `'ansible_roles'` to make it more obvious.

app/services/foreman_ansible/ansible_info.rb
```diff
@@ -14,5 +18,19 @@ def ansible_params
         memo
       end
     end
+
+    def ansible_roles
+      roles = host.all_ansible_roles.pluck(:id, :name)
```
xprazak2: If `pluck` is removed, you will be able to use attribute names below instead of indexes.

app/services/foreman_ansible/ansible_info.rb
```diff
@@ -14,5 +18,19 @@ def ansible_params
...
+
+      roles.each_with_object({}) do |role, rmemo|
+        variables = AnsibleVariable.where(:ansible_role_id => role[0], :override => true)
```
xprazak2: The content of the whole block looks very similar to what is in `ansible_params` method above. Do you think it would be possible to extract it into a separate method used by both?

xprazak2:
Thanks for your patch! I left a few comments inline.

ares:
I asked in the original PR but I'll rather add a question here as well - what is the reason for having a setting to avoid adding this info to ENC? Could you please describe a bit more your usecase? The only downside I see may be the performance impact, when we generate ENC for many hosts, but I'd prefer to not add too many settings. The ENC should ideally be consistent everywhere.

app/services/foreman_ansible/ansible_info.rb
avylove: Makes sense, I'll update it.

app/services/foreman_ansible/ansible_info.rb
avylove: Wouldn't that be less efficient from a memory perspective? My thought was an array is more memory efficient than a hash and only half the fields are needed. I think the other fields are timestamps, but there could be more fields in the future.

app/services/foreman_ansible/ansible_info.rb
avylove: You're right. I'll take look and see if I can do that.

avylove:
@ares I guess the advantage of the setting is to avoid breaking anyone who is depending on the old behavior. Adding the roles to the ENC won't break anyone, but this also removes the flattened Ansible variables from parameters as well. I think the behavior is better overall because many roles use the same variable names and some data could be lost flattening it.

The Puppet settings have a similar setting, but they enable it by default now. Best software practice is probably to add a setting in case it does break someone, then remove it after it has been the default for a while. It could default to enabled if you don't think it will break much.

ares:
I think we reached a point where we have more than 100 settings and it's very confusing for users and hard to support for developers. I appreciate you're thinking about backwards compatibility, but I think having two versions of ENC is making it ever worse. If it won't cause big increase of rendering time, I'd be happier to avoid making this optional. At the same time, we should bump the major version of a plugin to indicate backwards incompatible change.

avylove:
Ok, I can remove it. Is there anywhere else this code gets called? For the ENC, I think that gets generated on demand. I don't think there would be much change in processing time since the same data is parsed in both cases.

app/services/foreman_ansible/ansible_info.rb
avylove: Based on what @ares said about only having one behavior, I pulled out the old method.

ares:
@lpramuk, any concern with this change regarding backwards compatibility?

I'll need to play with this a bit and see how this feels. I'd like to avoid more ENC format changes in near future, so let's not rush this.

avylove:
What are the chances we'd want to include any metadata on the roles now or down the road? If so, it might make sense to drop the variables down a level, so it would be `ansible_roles.role1.variables.foo` instead of `ansible_roles.role1.foo`

ares:
Is there some metadata that you have in mind? Right now, I can't imagine useful example.

avylove:
The main thing that comes to mind is possibly some versioning information. Ansible versioning is still fairly primitive, but that could change in the future. Not saying we have anything else to put in now, just wondering if we sound be prepared for the eventuality so we wouldn't need to rework the ENC again.

adamruzicka:
What is the status here, apart from needing a rebase?

stejskalleos:
I'm closing the PR, it's been a while since the last update, feel free to update it and reopen new PR if you wish to continue.

PR #287, opened 2019-07-23
Fix obsolete cop names

adamruzicka:


ares:
Failures unrelated, thanks @adamruzicka, merging

PR #288, opened 2019-07-23
Fixes #27388 - migrate to @theforeman/vendor pkg

sharvit:
Migrate to use the new [@theforeman/vendor](https://github.com/theforeman/foreman-js/tree/master/packages/vendor) package from `npm`.

Don't forget to `rm -rf node_modules package-lock.json` and `npm i` before you test it.

### What to test?

1. `npm run lint` should work properly with no errors.
2. `npm test` should work properly with no errors.
3. running foreman dev-server should work
4. We shouldn't notice any UI changes. especially in the scss file that changed.

Thanks

### DIFF
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -2,7 +2,7 @@
 
 # rubocop:disable BlockLength
 Foreman::Plugin.register :foreman_ansible do
-  requires_foreman '>= 1.22'
+  requires_foreman '>= 1.23'
 
   security_block :foreman_ansible do
     permission :play_roles_on_host,
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   },
   "dependencies": {
     "@theforeman/vendor": "^0.1.1",
-    "react-json-tree": "^0.11.0"
-    "react-bootstrap": "^0.32.1",
+    "react-json-tree": "^0.11.0",
+    "react-bootstrap": "^0.32.1"
   },
   "devDependencies": {
     "@theforeman/vendor-dev": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -7,22 +7,12 @@
     "test": "test"
   },
   "dependencies": {
-    "babel-polyfill": "^6.26.0",
-    "classnames": "^2.2.5",
-    "lodash": "^4.17.11",
-    "patternfly": "^3.58.0",
-    "patternfly-react": "^2.25.4",
-    "prop-types": "^15.6.2",
-    "react": "^16.6.3",
-    "react-bootstrap": "^0.32.1",
-    "react-dom": "^16.6.3",
-    "react-json-tree": "^0.11.0",
-    "reselect": "^3.0.1",
-    "seamless-immutable": "^7.1.3"
+    "@theforeman/vendor": "^0.1.1",
+    "react-json-tree": "^0.11.0"
   },
   "devDependencies": {
+    "@theforeman/vendor-dev": "^0.1.1",
     "babel-eslint": "^8.2.1",
-    "babel-plugin-lodash": "^3.3.2",
     "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
@@ -63,6 +53,7 @@
   "jest": {
     "verbose": true,
     "moduleDirectories": [
+      "node_modules/@theforeman/vendor-core/node_modules",
       "node_modules",
       "webpack"
     ],
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,14 @@
   "plugins": [
     "transform-class-properties",
     "transform-object-rest-spread",
-    "transform-object-assign",
-    "lodash"
-  ]
+    "transform-object-assign"
+  ],
+  "env": {
+    "test": {
+      "presets": ["@theforeman/vendor-dev/babel.preset.js"]
+    },
+    "storybook": {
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -7,22 +7,12 @@
...
-    "patternfly": "^3.58.0",
-    "patternfly-react": "^2.25.4",
-    "prop-types": "^15.6.2",
-    "react": "^16.6.3",
-    "react-bootstrap": "^0.32.1",
-    "react-dom": "^16.6.3",
-    "react-json-tree": "^0.11.0",
-    "reselect": "^3.0.1",
-    "seamless-immutable": "^7.1.3"
+    "@theforeman/vendor": "^0.1.1",
```
ares: I suppose this is only available with Foreman 1.23?

sharvit: Right

xprazak2:
Works for me, we can merge.

xprazak2:
I found one bug - the spinner stylesheets are not present: 

spinner

xprazak2:
https://github.com/theforeman/foreman-js/pull/38 adds the stylesheets for loading state

sharvit:
Thanks @xprazak2 and @ares 

- package.json updated to contain theforeman/foreman-js#38
- added `requires_foreman '>= 1.23'`

sharvit:
Updated to use vendor version ^1.4.0

sharvit:
Anything you would like me to change so it can be merged?

sharvit:
@adamruzicka any chance of helping reviewing/merging this PR?

adamruzicka:
I'm still on PTO, can take a look tomorrow when I get back if it is still pending

lib/foreman_ansible/register.rb
```diff
@@ -2,7 +2,7 @@
 
 # rubocop:disable BlockLength
 Foreman::Plugin.register :foreman_ansible do
-  requires_foreman '>= 1.22'
+  requires_foreman '>= 1.23'
```
adamruzicka: Shoudln't this be `1.24`?

package.json
```diff
@@ -7,22 +7,12 @@
...
-    "patternfly": "^3.58.0",
-    "patternfly-react": "^2.25.4",
-    "prop-types": "^15.6.2",
-    "react": "^16.6.3",
-    "react-bootstrap": "^0.32.1",
-    "react-dom": "^16.6.3",
-    "react-json-tree": "^0.11.0",
-    "reselect": "^3.0.1",
-    "seamless-immutable": "^7.1.3"
+    "@theforeman/vendor": "^1.4.0",
```
tbrisker: Does this mean we have no version of ansible plugin that works with foreman 1.23?

sharvit: This is the original PR where I tried to add the vendor from the beginning. Because it didn't merge yet, I updated to use the latest vendor.

Would you suggest creating 2 different PRs for `1.23` and `1.24`?

tbrisker: Yes. We need a version that is compatible with 1.23, otherwise users won't be able to upgrade.

sharvit:
Now it should work with `>=1.23`.

Once it will get merged I will create another PR for `>=1.24`.
Let me know if I do right.

ares:
We don't have a separate branch for 1.23 (was not needed yet), since master already requires 1.24 (which caused the rebase) I'd say can we make this 1.24 only? or I can create a 1.23 branch and base it on today 1.22, that would though mean, the new release needs to be 5.0. So question to @tbrisker, do we need a release for 1.23?

tbrisker:
@ares afaik we need a release that uses vendor 0.1 to be compatible with 1.23, but i may be wrong and the previous release is still compatible even though it doesn't use vendor at all. In tasks I saw several places that were broken without vendor. You can check the latest release against 1.23 RC2 and verify that all of the webpacked UI works properly (and no errors in console).

tbrisker:
What's the status here? 1.23 has been out for a while already. If foreman_ansible works with 1.23 without this, i'm fine with updating only for 1.24 but that should be done soon so we can release a version that's compatible with 1.24.

sharvit:
> What's the status here? 1.23 has been out for a while already. If foreman_ansible works with 1.23 without this, i'm fine with updating only for 1.24 but that should be done soon so we can release a version that's compatible with 1.24.

I'm not sure, can someone please verify that?
This PR aims to work with `1.23` and there is another PR should be done for `1.24`

mmoll:
The current master is 1.24-only since a while anyway. Also, this needs a rebase.

ares:
Sorry for delay on this, we were busy with other stuff. We could cherry-pick this to 3.0-stable branch and do one more 3.x release with this change. But it would be good to make this through master branch, so it would be great to rebase, then we can merge. Also 1.24 update will be needed just after.

package.json
```diff
@@ -7,23 +7,13 @@
...
-    "classnames": "^2.2.5",
-    "jquery": "~2.2.4",
-    "lodash": "^4.17.11",
-    "patternfly": "^3.58.0",
-    "patternfly-react": "^2.25.4",
-    "prop-types": "^15.6.2",
-    "react": "^16.6.3",
+    "@theforeman/vendor": "^0.1.1",
+    "react-json-tree": "^0.11.0"
     "react-bootstrap": "^0.32.1",
```
xprazak2: There should be no trailing commas.

ares:
yeah, sorry updated

xprazak2:
Thanks everyone!

sharvit:
Thanks for pushing it forward!
Will create a follow-up PR soon to do the next vendor update

tbrisker:
@sharvit it's already open at https://github.com/theforeman/foreman_ansible/pull/308

PR #289, opened 2019-07-23
Fixes #27389 - fix eslint

sharvit:


sharvit:
btw, now travis will fail PR that break linting rules.

ares:
Does this now require Foreman 1.23? If so, we need to bump major version here. (note for myself, we'll need to branch 3.0-stable

ares:
if the answer is yes, please bump https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible/version.rb

sharvit:
@ares this change only affect the development environment

ares:
> @ares this change only affect the development environment

@sharvit isn't https://github.com/theforeman/foreman_ansible/pull/289/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R12 actually changing the dependency? I suppose it only formalizes the requirement, but I'm rather cautious. I suppose the jquery is consumed from the core, so just checking whether this would still work if we rebuild against older version of Foreman.

sharvit:
> > @ares this change only affect the development environment
> 
> @sharvit isn't https://github.com/theforeman/foreman_ansible/pull/289/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R12 actually changing the dependency? I suppose it only formalizes the requirement, but I'm rather cautious. I suppose the jquery is consumed from the core, so just checking whether this would still work if we rebuild against older version of Foreman.

As you assumed, it only formalizes the requirement to satisfy eslint. In any case the actual version will be supplied by foreman-core or by [@theforeman/vendor](https://github.com/theforeman/foreman-js/tree/master/packages/vendor) since `1.23`.

This will only affect your dev/test/storybook env where `jquery` wasn't available for you.

sharvit:
Rebased.
Can we merge it?

sharvit:
Can we merge it please?

sharvit:
@tbrisker any chace helping reviewing/merging this PR?

tbrisker:
[test foreman_ansible]

ares:
I'm sorry for the delay, merging now, thanks @sharvit

PR #291, opened 2019-07-25
Fixes #27435 - icon_path instead origin to override

lzap:
This needs https://github.com/theforeman/foreman/pull/6585


ares:
After merging this, we need to bump Foreman dependency to 1.24 and version to 4.0 as it requires PR from 1.24

ares:
fixes the issue :+1:

PR #292, opened 2019-07-27
Fixes 25159 - Ansible variables overrides on hosts/hostgroups edit page

avylove:
This updates the host and group edit pages to display and allow editing of Ansible variables on the parameters tab.

The omit checkbox will set the omit value, but it is not honored yet as @xprazak2 mentioned in https://github.com/theforeman/foreman_ansible/pull/241#issuecomment-465555679.

Ansible variables also seem to default to hidden. I'm not sure if that was intentional, but should probably not be the default.

### DIFF
--- a/app/helpers/foreman_ansible/ansible_variables_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_variables_helper.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  # Helper methods for host and hostgroup varibales
+  module AnsibleVariablesHelper
+
+    def ansible_lookup_key_with_diagnostic(obj, values_hash, lookup_key, lookup_value)
+      value, matcher = ansible_value_matcher(obj, values_hash, lookup_key)
+      inherited_value = LookupKey.format_value_before_type_cast(value, lookup_key.key_type)
+      effective_value = lookup_value.lookup_key_id.nil? ? inherited_value.to_s : lookup_value.value_before_type_cast.to_s
+      warnings = lookup_key_warnings(lookup_key.required, effective_value.present?)
+      popover_value = lookup_key.hidden_value? ? lookup_key.hidden_value : inherited_value
+
+      parameter_value_content(
+        "#{parameters_receiver}_lookup_values_attributes_#{lookup_key.id}_value",
+        effective_value,
+        :popover => diagnostic_popover(lookup_key, matcher, popover_value, warnings),
+        :name => "#{lookup_value_name_prefix(lookup_key.id)}[value]",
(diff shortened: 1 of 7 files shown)


avylove:
Not exactly sure how to hand those two Unescaped model attribute Code Climate issues. Would appreciate some insight.

It looks like what breaks omit is update_generic_matcher() in [foreman/app/services/classification/values_hash_query.rb](https://github.com/theforeman/foreman/blob/9adae642efd8bfd8e88d49062430e239cd3307a9/app/services/classification/values_hash_query.rb#L100)

`computed_lookup_value[:managed] = lookup_value.omit if lookup_value.lookup_key.puppet?`

The lookup_key shows up as a PuppetclassLookupKey.

Option are to overwrite the method or to set `puppet?` to `true` in [foreman_ansible/app/models/ansible_variable.rb](https://github.com/theforeman/foreman_ansible/blob/master/app/models/ansible_variable.rb) It seems to inherit the values.

xprazak2:
Thank you for this nice feature and time you spent implementing it. You can ignore the remaining Code Climate issues. The 'omit' is too tied to Puppet, so I would just skip dealing with it right now and I would not show the checkbox at all - we need to refactor in core so that we can have a clean solution.

Everything works for hosts with existing roles, however newly added Ansible roles with variable overrides are not displayed:

1) go to host edit page
2) on 'Ansible roles' tab, add a role that has overridable variables
3) no variables are displayed for newly added role on 'Parameters' tab
4) save the host
5) go to edit page again, variables are now displayed on 'Parameters' tab

Even though steps 4 and 5 provide a workaround, this workflow is not very user friendly and makes it look like the form is broken because it does not show the variables that should be visible.

The reason for this is that erb views are rendered server-side and are not updated when you add a role. There are plans to have the variable overrides as a React component which would be able to solve this problem. It would also allow us to avoid Deface as now we have a way how to extend pages using [Slot & Fill](https://github.com/theforeman/foreman/pull/6849).

I started [initial investigation](https://github.com/xprazak2/foreman_ansible/tree/vars-over) on how this could be achieved, but did not finish due to other responsibilities. The code in my branch is kind of messy and 'Slot' extension point needs to be added to the hosts form, but I am willing to provide any support that I can if you are willing to continue where I left off.

avylove:
It would definitely be nicer if it was more dynamic. If that can get done, it would be great, but if that's a longer effort, it would be good to have something sooner. I'll try to look into it when I get a chance, but I'm not sure I'll have a chance to dig into it in the near term.

ares:
This will probably be superseeded by https://github.com/theforeman/foreman_ansible/pull/304

ezr-ondrej:
This feature is implemented as part of the new Host details page. Closing this one.
Thanks a lot for bringing up the topic @avylove ! :+1:

PR #294, opened 2019-07-31
Fixes #27021 - make runner default implementation

ezr-ondrej:
Installer PR (theforeman/puppet-foreman_proxy#515) got merged, so we can proceed with this one I believe.


ares:
rubocop failed but tests were green, thanks @ezr-ondrej, works fine, merging!

PR #295, opened 2019-08-01
Fixes #27491 - Ensure ssh and sudo password are not leaked

adamruzicka:
Before this change the were stored in the inventory, but that meant they were shown in task export as these fields were not hidden.

This patch moves the secrets under a special key which is hidden while exporting. The downside is we have to rebuild the inventory on the smart proxy's side.

Requires:
- [ ] https://github.com/theforeman/foreman_remote_execution/pull/425

### DIFF
--- a/lib/foreman_ansible_core/runner/playbook.rb
+++ b/lib/foreman_ansible_core/runner/playbook.rb
@@ -14,7 +14,7 @@ module ForemanAnsibleCore
 
       def initialize(inventory, playbook, options = {}, suspended_action:)
         super :suspended_action => suspended_action
-        @inventory = inventory
+        @inventory = rebuild_secrets(inventory, options[:secrets])
         unknown_hosts.each do |host|
           add_to_known_hosts(host)
         end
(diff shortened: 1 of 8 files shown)


adamruzicka:
If the user provides secrets per host using the `ansible_$SOMETHING` format, they would still show up, so far I'm inclined towards putting in a blacklist of some sort.

lib/foreman_ansible_core/runner/playbook.rb
```diff
@@ -118,6 +118,20 @@ def add_to_known_hosts(host)
...
+
+          new_secrets = {
+            'ansible_ssh_pass' => inventory['ssh_password'] || per_host['ansible_ssh_pass'],
+            'ansible_sudo_pass' => inventory['sudo_password'] || per_host['ansible_sudo_pass']
+          }
+          inventory['_meta']['hostvars'][name].update(new_secrets)
+        end
+
+        inventory
+      end
```
lzap: Is this covered in "when using secrets" test? I am not sure because it's stubbed many times.

PR #298, opened 2019-09-01
Fixes #27747 - Remove taxonomy enabled checks

tbrisker:


ares:
Thanks both @tbrisker and @adamruzicka, merging!

PR #299, opened 2019-09-10
Fixes #27937 - API for Ansible Inventory template scheduling

ofedoren:
Requires https://github.com/theforeman/foreman/pull/6875.

To use the endpoint there should exist the `Ansible Inventory` report template (part of this PR)

### DIFF
--- a/app/controllers/api/v2/ansible_inventories_controller.rb
+++ b/app/controllers/api/v2/ansible_inventories_controller.rb
@@ -31,16 +31,66 @@ module Api
         show_inventory :hostgroup_ids, :hostgroup_id
       end
 
+      api :POST, '/ansible_inventories/schedule',
+          N_('Schedule generating of Ansible Inventory report')
+      param :input_values, Hash, N_('Hash of input values of type input=>value')
+      param :report_format, ReportTemplateFormat.selectable.map(&:id),
+            N_("Report format, defaults to '%s'") % 'json'
+      example <<-EXAMPLE
+      POST /ansible/api/ansible_inventories/schedule
+      {
+        "input_values": {
+          "Organization": "yes",
+          "Location": "yes",
+          "IPv4": "yes",
+          "Facts": "no"
+        }
+      }
+      200
+      {
+        "job_id": UNIQUE-REPORT-GENERATING-JOB-UUID
+        "data_url": "/api/v2/report_templates/1/report_data/UNIQUE-REPORT-GENERATING-JOB-UUID"
+      }
+      EXAMPLE
+
+      def schedule
+        @composer = ReportComposer.from_api_params(schedule_params)
+        if @composer.valid?
+          job = @composer.schedule_rendering
+          response = { :job_id => job.provider_job_id }
+          response[:data_url] = report_data_api_report_template_path(
+            @report_template, :job_id => job.provider_job_id
+          )
+          render :json => response
+        else
+          @ansible_inventory = @composer
+          process_resource_error(:resource => @ansible_inventory)
+        end
+      rescue StandardError => e
+        render_error 'standard_error', :status => :internal_error,
+                                       :locals => { :exception => e }
+      end
+
       def action_permission
         case params[:action]
         when 'hosts', 'hostgroups'
           :view
+        when 'schedule'
+          :generate
         else
           super
         end
       end
 
       private
+
+      def schedule_params
+        template_name = Setting::Ansible.find_by(:name => 'ansible_inventory_template').value
+        @report_template = ReportTemplate.find_by!(:name => template_name)
+        params[:id] = @report_template.id
+        params[:report_format] = 'json' if params[:report_format].blank?
+        params
+      end
 
       def show_inventory(ids_key, condition_key)
         ids = params.fetch(ids_key, []).uniq
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -71,6 +71,9 @@ Foreman::Plugin.register :foreman_ansible do
     permission :edit_hostgroups,
                { :'api/v2/hostgroups' => [:assign_ansible_roles] },
                :resource_type => 'Hostgroup'
+    permission :generate_report_templates,
+               { :'api/v2/ansible_inventories' => [:schedule] },
+               :resource_type => 'ReportTemplate'
   end
 
   role 'Ansible Roles Manager',
@@ -82,7 +85,7 @@ Foreman::Plugin.register :foreman_ansible do
         :edit_ansible_variables, :destroy_ansible_variables]
 
   role 'Ansible Tower Inventory Reader',
-       [:view_hosts, :view_hostgroups, :view_facts],
+       [:view_hosts, :view_hostgroups, :view_facts, :generate_report_templates],
        'Permissions required for the user which is used by Ansible Tower Dynamic Inventory Item'
 
   add_all_permissions_to_default_roles
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -80,6 +80,13 @@ class Setting
               false,
               format(N_('%{cfgmgmt} out of sync disabled'),
                      :cfgmgmt => 'Ansible')
+            ),
+            set(
+              'ansible_inventory_template',
+              N_('Foreman will use this template to schedule the report '\
+                 'with Ansible inventory'),
+              'Ansible Inventory',
+              N_('Default Ansible inventory report template')
             )
           ].compact.each do |s|
             create(s.update(:category => 'Setting::Ansible'))
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -99,6 +99,7 @@ Rails.application.routes.draw do
             get :hosts
             post :hostgroups
             get :hostgroups
+            post :schedule
           end
         end
       end


sjha4:
We would probably also need the below fields ( or decide which ones we need from the below response) from the earlier setup:

```
"foreman": {
          "architecture_id": 1,
          "architecture_name": "x86_64",
          "build": "False",
          "capabilities": [
            "build"
          ],
          "certname": "samir",
          "comment": "None",
          "compute_profile_id": "None",
          "compute_profile_name": "None",
          "compute_resource_id": "None",
          "compute_resource_name": "None",
          "content_facet_attributes": {
            "applicable_module_stream_count": 0,
            "applicable_package_count": 0,
            "content_source": "None",
            "content_source_id": "None",
            "content_source_name": "None",
            "content_view": {
              "id": 1,
              "name": "Default Organization View"
            },
            "content_view_id": 1,
            "content_view_name": "Default Organization View",
            "errata_counts": {
              "bugfix": 0,
              "enhancement": 0,
              "security": 0,
              "total": 0
            },
            "id": 1,
            "kickstart_repository": "None",
            "kickstart_repository_id": "None",
            "kickstart_repository_name": "None",
            "lifecycle_environment": {
              "id": 1,
              "name": "Library"
            },
            "lifecycle_environment_id": 1,
            "lifecycle_environment_name": "Library",
            "upgradable_module_stream_count": 0,
            "upgradable_package_count": 0,
            "uuid": "None"
          },
          "created_at": "2019-09-10 18:56:49 UTC",
          "disk": "None",
          "domain_id": "None",
          "domain_name": "None",
          "enabled": "True",
          "environment_id": "None",
          "environment_name": "None",
          "global_status": 1,
          "global_status_label": "OK",
          "hostgroup_id": "None",
          "hostgroup_name": "None",
          "hostgroup_title": "None",
          "id": 1,
          "image_file": "",
          "image_id": "None",
          "image_name": "None",
          "installed_at": "None",
          "ip": "10.11.216.233",
          "ip6": "None",
          "last_compile": "2019-09-10 18:56:51 UTC",
          "last_report": "None",
          "location_id": 2,
          "location_name": "Default Location",
          "mac": "e4:70:b8:22:1d:66",
          "managed": "False",
          "medium_id": "None",
          "medium_name": "None",
          "model_id": "None",
          "model_name": "None",
          "name": "samir",
          "operatingsystem_id": 1,
          "operatingsystem_name": "CentOS 7",
          "organization_id": 1,
          "organization_name": "Default Organization",
          "owner_id": 4,
          "owner_name": "Admin User",
          "owner_type": "User",
          "provision_method": "build",
          "ptable_id": "None",
          "ptable_name": "None",
          "puppet_ca_proxy": "None",
          "puppet_ca_proxy_id": "None",
          "puppet_ca_proxy_name": "None",
          "puppet_proxy": "None",
          "puppet_proxy_id": "None",
          "puppet_proxy_name": "None",
          "puppet_status": 0,
          "pxe_loader": "None",
          "realm_id": "None",
          "realm_name": "None",
          "sp_ip": "None",
          "sp_mac": "None",
          "sp_name": "None",
          "sp_subnet_id": "None",
          "subnet6_id": "None",
          "subnet6_name": "None",
          "subnet_id": "None",
          "subnet_name": "None",
          "updated_at": "2019-09-10 18:56:54 UTC",
          "uptime_seconds": "None",
          "use_image": "None",
          "uuid": "None"
        }
```

sjha4:
@ofedoren : I am not sure if I am using the required param correctly, but with something like this: {'required':'all'} I get an error. Do you have an example curl with what the hash expects?

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,30 @@
...
+- plugin: foreman_ansible
+-%>
+<%- load_hosts.each_record do |host| -%>
+<%- inventory_data = {
+      'Name': host.name
+    }
+    required = input('required').split(',').map { |r| r.strip.downcase }
+    inventory_data.merge!('Organization': host.organization) if (['all', 'organization'] & required).any?
+    inventory_data.merge!('Location': host.location) if (['all', 'location'] & required).any?
+    inventory_data.merge!('Hostgroup': host.hostgroup.name) if (['all', 'hostgroup'] & required).any?
```
sjha4:
```suggestion
    inventory_data.merge!('Hostgroup': host.hostgroup.name) if (['all', 'hostgroup'] & required).any? && host.hostgroup
```
to avoid nil reference.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,30 @@
...
+- name: required
+  required: false
+  input_type: user
+  description: Comma-separated values that should be returned.
+  advanced: false
+  value_type: plain
+
+model: ReportTemplate
+require:
+- plugin: foreman_ansible
```
ares: we'll also need to add version restriction here, probably that will be >= 3.1.0

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,30 @@
+<%#
+name: Ansible Inventory
+snippet: false
+template_inputs:
+- name: required
+  required: false
+  input_type: user
+  description: Comma-separated values that should be returned.
```
ares: I think it would be nicer user experience if we define separate input for every optional part, it would have restricted values yes no and would be required. That in UI should generate select boxes to chose from yes/no.

We could also use this chance to add default values to inputs (that would be PR to Foreman core)

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,30 @@
+<%#
+name: Ansible Inventory
+snippet: false
```
ares: Ideally we should introduce description field to templates and put some description to this template saying this template has a special meaning and should not be renamed. This is not the first template that would benefit from it. This would mean a PR in Foreman core but would be a nice addition.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
ares: also we should call it 'Host group' or 'HostGroup' perhaps, this is probably a user facing string

sjha4:
Below is the list of fields available on the host model so they can be pulled into the report:
https://gist.github.com/sjha4/eee846297dcf051e9a4f14198b444309
I am looking at the Content Facet Attributes data that is needed to create the content view and lifecycle groups for hosts.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
sjha4: So based on the new [fields](https://gist.github.com/sjha4/eee846297dcf051e9a4f14198b444309) we want to add, I tried this:
`['architecture_id','architecture_name'].each do |field|
inventory_data.merge!(field=>host[field])`

nor can I do something like `inventory_data.merge!(field=>host.send(field))`

It seems the` Host::Managed::Jail` does not like it.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
ofedoren: Done: https://github.com/theforeman/foreman/pull/7050

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
ofedoren: Done: https://github.com/theforeman/foreman/pull/7051

ofedoren:
The third commit requires https://github.com/Katello/katello/pull/8356.

ofedoren:
The [fourth](https://github.com/theforeman/foreman_ansible/pull/299/commits/1775f9154bfd0e2a893b43b1b71218bc1a555df1) commit requires https://github.com/theforeman/foreman/pull/7052.

ofedoren:
> Below is the list of fields available on the host model so they can be pulled into the report:
> https://gist.github.com/sjha4/eee846297dcf051e9a4f14198b444309
> I am looking at the Content Facet Attributes data that is needed to create the content view and lifecycle groups for hosts.

@sjha4, that's quite a big list and I'm not sure if we should add a separate template input for each field.

@ares, how can we choose what should be in the report (or at least as template inputs) and what should be not?

sjha4:
@ofedoren : Right..It's >70 fields in the intial API call I was testng with. I am not sure we need all the details on the report.
We do need the content_facet_attributes. I have a PR in katello https://github.com/Katello/katello/pull/8357 and a dummy report here: http://pastebin.test.redhat.com/798999 with content_facety_attributes that we would need to add to this report.

app/controllers/api/v2/ansible_inventories_controller.rb
```diff
@@ -31,9 +31,34 @@ def hostgroups
         show_inventory :hostgroup_ids, :hostgroup_id
       end
 
+      api :POST, '/ansible_inventories/schedule_report',
```
ezr-ondrej: This sounds like a odd name for potential user, I would rather name it according to what it does, instead of how.
Maybe just `/ansible_inventories/schedule` would do?
Or `/ansible_inventories/schedule_generate` ?

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
ezr-ondrej: Code there looks good, but the use-case seems bit odd to me. @ares, how another editable field is going to stop me from making a mess, if I am a messy user? Can't we have lockable report_templates instead?

ezr-ondrej: if not, `Ansible Inventory` is suggestive enough for me to get an idea what that might be used for and why I shall not mess with it IMHO.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
ezr-ondrej: Code there looks good, but the use-case seems bit odd to me. @ares, how another editable field is going to stop me from making a mess, if I am a messy user? Can't we have lockable report_templates instead?

ezr-ondrej: if not, `Ansible Inventory` is suggestive enough for me to get an idea what that might be used for and why I shall not mess with it IMHO.

app/controllers/api/v2/ansible_inventories_controller.rb
ofedoren: @ezr-ondrej, thanks, I think you're right, so I've changed it to `/ansible_inventories/schedule`.

ofedoren:
@sjha4, thanks for your input! I've added Content Facet Attributes to the template.

ares:
You can use https://projects.theforeman.org/issues/27937 for tracking this

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,135 @@
...
+      'IPv4': input('IPv4') == 'yes' ? host.ip : nil,
+      'Subnet': input('Subnet') == 'yes' ? host.subnet : nil,
+      'Smart Proxies': input('Smart Proxies') == 'yes' && host.smart_proxies ? host.smart_proxies.map { |p| p.name } : nil,
+      'Content Facet Attributes': content_facet_attribute_data,
+      'Facts': input('Facts') == 'yes' ? host.facts : nil
+    }
+    report_row(inventory_data)
+-%>
+<%- end -%>
+<%= report_render(format: :json) -%>
```
ares: Can we just keep report_render here and specify the format in the ansible inventory endpoint when we call renderer? That way people could use the same template to get the same inventory in different format.

app/views/foreman_ansible/job_templates/ansible_-_inventory.erb
```diff
@@ -0,0 +1,135 @@
...
+  options: "yes\r\nno"
+  default: no
+model: ReportTemplate
+require:
+- plugin: foreman_ansible
+  version: '>= 3.1.0'
+- plugin: katello
+￼  version: '>= 3.9.0'
+-%>
+<%- content_facet_attr = input('Content Facet Attributes') == 'yes' ? true : false -%>
```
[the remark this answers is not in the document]
ares: And since this is user facing, I'd call it 'Content Attributes', user does not know, what is facet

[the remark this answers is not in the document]
ares: oh I confused it with `host_param_true?`, nevermind. let's keep yes/no only, that's fine

ares:
looks well now, safe mode PRs seems to be in, this is now blocked on default values, description and format PRs in core

ofedoren:
The template has been updated:
  - By default the following data are in the report: Host name, Organization, Location, Host Group, IPv4. Other can be configured via template inputs.
  - Host group is shown with nesting included (as in UI, e.g. Base/Katello CentOS 7). This requires https://github.com/theforeman/foreman/pull/7075

mmoll:
please rebase

ofedoren:
@mmoll, rebased.

mmoll:
now it fails on rubocop

ofedoren:
@mmoll, thanks :) fixed.

mmoll:
almost there, permission missing :)

lib/foreman_ansible/register.rb
```diff
@@ -82,7 +82,7 @@
         :edit_ansible_variables, :destroy_ansible_variables]
 
   role 'Ansible Tower Inventory Reader',
-       [:view_hosts, :view_hostgroups, :view_facts],
+       [:view_hosts, :view_hostgroups, :view_facts, :view_schedule],
```
ares: Do we need separate view_schedule permission? I think this should be the same permission as for genertimg regular report

ofedoren: I've reused `generate_report_tempaltes` permission if it makes sense.

ares:
[test foreman_ansible]

lib/foreman_ansible/register.rb
ares: yes, that's exactly what I had in mind

ares:
Thanks @ofedoren, merging!

PR #301, opened 2019-09-12
Fixes #27834 - Expect runner json may miss event_data key 

adamruzicka:
Sometimes ansible-runner gives hints not related to a specific host, but these messages have a different file format than the regular output files.

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -41,7 +41,7 @@ module ForemanAnsibleCore
         logger.debug("[foreman_ansible] - parsing event file #{event_file}")
         begin
           event = JSON.parse(File.read(event_file))
-          if (hostname = event['event_data']['host'])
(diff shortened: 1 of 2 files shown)


lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -41,7 +41,7 @@ def handle_event_file(event_file)
         logger.debug("[foreman_ansible] - parsing event file #{event_file}")
         begin
           event = JSON.parse(File.read(event_file))
-          if (hostname = event['event_data']['host'])
+          if (hostname = event.dig('event_data', 'host'))
```
ares: this is added in 2.3, do we have such ruby everywhere when this code runs?

adamruzicka: I think so, It only runs in smart proxy dynflow core, which runs in SCL and on debs there is ruby 2.5 iirc

PR #309, opened 2019-10-23
Fixes #28113 - Replace deprecated db_facts method

xprazak2:
Removed in: https://github.com/theforeman/foreman/commit/a105d06bd0f3daff804a492cb3c28781fc441b4d#diff-671e5385f8ec83d391c87d7c292604ed

### DIFF
--- a/app/services/foreman_ansible/fact_importer.rb
+++ b/app/services/foreman_ansible/fact_importer.rb
@@ -55,16 +55,13 @@ module ForemanAnsible
     end
 
     def missing_facts
-      db_fact_names = if db_facts.is_a? Hash
-                        db_facts.keys
-                      else
-                        db_facts
-                      end
-      # In Foreman versions prior to 1.14, the db_facts key
-      # used to be a hash. Now it's an ActiveRecord::AssociationRelation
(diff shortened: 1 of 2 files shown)


app/services/foreman_ansible/fact_importer.rb
```diff
@@ -55,10 +55,10 @@ def add_compose_fact(imported_values, fact_name, fact_fqn)
     end
 
     def missing_facts
-      db_fact_names = if db_facts.is_a? Hash
-                        db_facts.keys
+      db_fact_names = if host.fact_values.is_a? Hash
```
tbrisker: it isn't a hash anymore so this condition and the following comment can be removed

ezr-ondrej:
[test foreman_ansible]

test/unit/services/fact_importer_test.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'test_plugin_helper'
```
ezr-ondrej: :policeman: wants an empty line before :disappointed:

PR #310, opened 2019-11-12
Fixes #28245 - Improve textation

domitea:


app/helpers/foreman_ansible/hosts_helper_extensions.rb
```diff
@@ -10,7 +10,7 @@ def multiple_actions
         actions = super
         if User.current.can?(:create_job_invocations) &&
            User.current.can?(:play_roles_on_host)
-          actions += [[_('Play Ansible roles'),
+          actions += [[_('Play all Ansible roles from hostgroup'),
```
ares: s/hostgroup/host group/ :-) it's wrong model name, let's not propagate that mistake to human language :-)

I'd also not even mention host group here, roles can be assigned to host directly. This run will include all roles from host group, its parent, its parent parent, ... and roles assigned to the host directly

we also have other place in host detail, which already says "Run Ansible roles", perhaps that would be better than Play? So I'd vote for either "Run all Ansible roles" or simply "Run Ansible roles"

And another places where this shoud probably changed are
* https://github.com/theforeman/foreman_ansible/blob/master/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb#L20
https://github.com/theforeman/foreman_ansible/blob/master/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb#L30
* https://github.com/theforeman/foreman_ansible/blob/master/app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb#L26
* https://github.com/theforeman/foreman_ansible/blob/master/app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb#L35
* https://github.com/theforeman/foreman_ansible/blob/master/app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb#L3
* https://github.com/theforeman/foreman_ansible/blob/master/app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb#L5

domitea: I think it will be better to use "Run all Ansible roles" rather than "Run Ansible roles" beacuse it removes question "What Ansible roles?"

ezr-ondrej: well there can still raise a question what `all` really means, but I would agree it's a bit more suggestive :)

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
```diff
@@ -1,8 +1,8 @@
 <%=
   play_roles = if hostgroup.all_ansible_roles.empty?
-    "<a title='No Roles assigned' href=\"#\">#{_('Play Roles')}</a>".html_safe
+    "<a title='No Roles assigned' href=\"#\">#{_('Run all Ansible roles')}</a>".html_safe
   else
-    display_link_if_authorized(_('Play Roles'),   hash_for_play_roles_hostgroup_path(:id => hostgroup),   :'data-no-turbolink' => true)
+    display_link_if_authorized(_('Run all Ansible roles'),   hash_for_play_roles_hostgroup_path(:id => hostgroup),   :'data-no-turbolink' => true)
```
ezr-ondrej: Could you please fix the spacing as we are touching this line already?

ares:
I think test failures are unrelated, though pretty consistent. @ezr-ondrej anything else to add or ready to merge?

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -17,7 +17,7 @@ module HostgroupsControllerExtensions
           before_action :find_ansible_roles, :only => [:assign_ansible_roles]
 
           api :POST, '/hostgroups/:id/play_roles',
-              N_('Plays Ansible roles on a hostgroup')
+              N_('Runs all Ansible roles')
```
ezr-ondrej: Here the `on a hostgroup` should stay, as it refers to a host group (we could leave out all here).

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -27,7 +27,7 @@ def play_roles
           end
 
           api :POST, '/hostgroups/multiple_play_roles',
-              N_('Plays Ansible roles on hostgroups')
+              N_('Runs all Ansible roles')
```
ezr-ondrej: The same in here, `on hostgroups` should stay.

app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb
```diff
@@ -23,7 +23,7 @@ def find_resource
           end
 
           api :POST, '/hosts/:id/play_roles',
-              N_('Plays Ansible roles on a host')
+              N_('Runs all Ansible roles')
```
ezr-ondrej: Here as well ^^ `on a host` should stay

app/controllers/foreman_ansible/api/v2/hosts_controller_extensions.rb
```diff
@@ -32,7 +32,7 @@ def play_roles
           end
 
           api :POST, '/hosts/multiple_play_roles',
-              N_('Plays Ansible roles on hosts')
+              N_('Runs all Ansible roles')
```
ezr-ondrej: ^^

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
```diff
@@ -1,8 +1,8 @@
 <%=
   play_roles = if hostgroup.all_ansible_roles.empty?
-    "<a title='No Roles assigned' href=\"#\">#{_('Play Roles')}</a>".html_safe
+    "<a title='No Roles assigned' href=\"#\">#{_('Run all Ansible roles')}</a>".html_safe
```
ezr-ondrej: Not your fault here, but we have a button to `run ansible roles` which does nothing, shouldn't it be at least disabled? Or completely disappear - with just a placeholder message to inform user there are no roles? This seems bit bad UX. @ares, @domitea ?
The `disabled="disabled"` would be sufficient for me (we do it everywhere like that, but I think right UX is the button dissapears with an info)

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
```diff
@@ -1,8 +1,8 @@
...
   else
-    display_link_if_authorized(_('Play Roles'),   hash_for_play_roles_hostgroup_path(:id => hostgroup),   :'data-no-turbolink' => true)
+    display_link_if_authorized(_('Run all Ansible roles'), hash_for_play_roles_hostgroup_path(:id => hostgroup), :'data-no-turbolink' => true)
```
ezr-ondrej: We could frase it `on a hostgroup` here as we know we are on a hostgroup :)

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
ares: I think the button should stay so people don't get confused why sometimes it's missing. I agree disabling would be nice. If we decide to go that way, it should have title (appear on hover) explaining why it's disabled. @domitea is it ok to include this to this PR or should we track it separately? I believe it's a good improvement though and since we're touching this, it makes sense to do it.

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
ares: that seems very long, perhaps add some hover title? That should say something like "Run all Ansible roles on hosts belonging to this host group"

ezr-ondrej: Yeah and is not necessary as we are on the hostgroup, so don't worry about this one, title would be nice, but not needed.

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
```diff
@@ -1,8 +1,9 @@
 <%=
   play_roles = if hostgroup.all_ansible_roles.empty?
-    "<a title='No Roles assigned' href=\"#\">#{_('Play Roles')}</a>".html_safe
+    "<a title='No Roles assigned' href=\"#\" disabled='disabled'>#{_('Run all Ansible roles')}</a>".html_safe
+    link_to _('Run all Ansible roles'), 'javascript:void(0);', disabled: true, title: 'No Roles assigned'
```
ares: don't we have 2 times the same link here now? while only the second will be in fact printed...

domitea:
Comment applied, commits has been squashed... Can anybody review it again? @ezr-ondrej @ares

ezr-ondrej:
This is how it behaves on hostgroup dropdown now if hostgroup has no roles:
run_ar_hostgroup
@domitea, @ares I believe we can't do better, but it feels missleading. I would feel better if it's not there. It's no blocker for me though. I'm ok to merge, @ares ?

ares:
Yeah it's not ideal but not worse than it was before. We display this always without any indication of no roles assigned. Hence I'd say :+1: for merging and bonus point for creating a new redmine issue to allow disabled links in this menu.

ezr-ondrej:
Thanks @domitea and @ares! :+1: 
I've created the ticket [#28374](https://projects.theforeman.org/issues/28374) to track the disabling of the dropdown menu and merging.

PR #311, opened 2019-11-12
Fixes #28253 - upgrade vendor to v3

sharvit:
The main upgrades are `redux-hooks` and the switch to `babel` v7


sharvit:
Updated to use `@theforeman/vendor`

adamruzicka:
[test foreman_ansible]

ares:
core PR is in also rubocop is in, let's retry [test foreman_ansible], if jenkins confirms, I'll merge and build a new version

ares:
oh, this would have to be rebased now in order to make rubocop pass

sharvit:
Thanks @ares, rebased

ares:
tests are green now, mental note for 2020: did we drop patternfly dep and do we consume that from vendor now? (we've seen this with tasks 1.0.0)

sharvit:
**`patternfly-react`**, is a part of the vendor.

As for **`patternfly`**:

The foreman downloads patternfly styles as a gem and build it into the assets pipeline, see:
- https://github.com/theforeman/foreman/blob/develop/bundler.d/assets.rb#L4
- https://github.com/theforeman/foreman/blob/develop/app/assets/stylesheets/application.scss#L9

But, the `vendor` also bring it from `npm` and only include a small partial from `pattenfly`, see:
https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/scss/vendor-core.scss#L15

**IMO it should all be moved to the `vendor` and all the s/css should be moved to webpack.**

sharvit:
After I ran `npm test`, this appears in my `root` folder, guess it should be in `./coverage/*`?
```
lcov-report/
lcov.info
```

sharvit:
Updated to support vendor v4.
Can we merge it?

adamruzicka:
First thing tomorrow morning

package.json
```diff
@@ -6,18 +6,19 @@
...
-    "@theforeman/vendor-dev": "^1.7.0",
-    "babel-eslint": "^8.2.1",
-    "babel-plugin-transform-class-properties": "^6.24.1",
-    "babel-plugin-transform-object-assign": "^6.22.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-preset-env": "^1.6.0",
-    "babel-preset-react": "^6.24.1",
+    "@babel/core": "^7.7.0",
+    "@theforeman/env": "^3.3.2",
+    "@theforeman/builder": "^3.3.2",
```
tbrisker: This will still fail to build since builder is 4 in the packaging already. Please update @sharvit

sharvit:
Thanks @tbrisker, made some small changes so it would work with builder v4

sharvit:
Also updated `eslint` and `prettier` to fix Travis

PR #313, opened 2019-11-25
Fixes #28344 - Add endpoint for variable overrides

xprazak2:
This adds an endpoint that serves ansible variables and overrides, which will be needed for displaying them in host(group) form. 

Includes changes in AnsibleRolesSwitcher to query the endpoint. There are no changes from user's perspective, but you can notice the request and new data in store.

I intentionally did not include any visible UI changes to keep this as small as possible.

### DIFF
--- a/app/controllers/ui_ansible_roles_controller.rb
+++ b/app/controllers/ui_ansible_roles_controller.rb
@@ -3,6 +3,8 @@ class UiAnsibleRolesController < ::Api::V2::BaseController
     super resource
   end
 
+  layout 'api/v2/layouts/index_layout', :only => [:index]
+
   def index
     @ui_ansible_roles = resource_scope_for_index(:permission => :view_ansible_roles)
   end
--- a/app/services/foreman_ansible/override_resolver.rb
+++ b/app/services/foreman_ansible/override_resolver.rb
@@ -1,0 +1,56 @@
+module ForemanAnsible
+  # Service which resolves override values for hosts and hostgroups
+  class OverrideResolver
+    attr_reader :overrides, :ansible_variables
+
+    def initialize(host_or_hg)
+      @overrides = calculate_variable_overrides(host_or_hg)
+    end
+
+    def resolve(ansible_variable)
+      override = @overrides[ansible_variable.id]
+      return unless override
+      override[ansible_variable.key]
+    end
+
+    private
+
+    def calculate_variable_overrides(resource)
+      if resource.is_a? Hostgroup
+        @ansible_variables = AnsibleVariable.where(:ansible_role_id => resource.inherited_and_own_ansible_roles, :override => true)
+        hostgroup_values_hash resource
+      else
+        @ansible_variables = AnsibleVariable.where(:ansible_role_id => resource.all_ansible_roles, :override => true)
+        @ansible_variables.values_hash(resource).raw
+      end
+    end
+
+    def hostgroup_values_hash(hostgroup)
+      @ansible_variables.each_with_object({}) do |ansvar, memo|
+        next memo unless ansvar.override
+        hash_result = inherited_hostgroup_value(hostgroup, ansvar)
+        next memo unless hash_result
+        memo[ansvar.id] = {
+          ansvar.key => {
+            :value => hash_result[:lookup_value].value,
+            :element => 'hostgroup',
+            :element_name => hash_result[:hostgroup].name
+          }
+        }
+        memo
+      end
+    end
+
+    def inherited_hostgroup_value(hostgroup, ansvar)
+      return if !ansvar.path_elements.flatten.include?('hostgroup') && !Setting['matchers_inheritance']
+
+      value = nil
+      matched_hg = hostgroup.path.reverse.find do |hg|
+        value = LookupValue.find_by(:lookup_key_id => ansvar.id, :id => hg.lookup_values)
+      end
+
+      return unless value
+      { :lookup_value => value, :hostgroup => matched_hg }
+    end
+  end
+end
--- a/lib/foreman_ansible/engine.rb
+++ b/lib/foreman_ansible/engine.rb
@@ -18,6 +18,7 @@ module ForemanAnsible
     config.autoload_paths += Dir["#{config.root}/app/services"]
     config.autoload_paths += Dir["#{config.root}/app/views"]
     config.autoload_paths += Dir["#{config.root}/app/lib"]
+    config.autoload_paths += Dir["#{config.root}/test/"]
 
     initializer 'foreman_ansible.load_default_settings',
                 :before => :load_config_initializers do
--- a/app/controllers/ansible_variables_controller.rb
+++ b/app/controllers/ansible_variables_controller.rb
@@ -4,6 +4,7 @@
 class AnsibleVariablesController < ::LookupKeysController
   include Foreman::Controller::AutoCompleteSearch
   include ForemanAnsible::Concerns::ImportControllerHelper
+  include ForemanAnsible::Concerns::AnsibleVariableOverrides
   include Foreman::Controller::Parameters::AnsibleVariable
 
   before_action :import_new_roles, :only => [:confirm_import]
@@ -55,6 +56,22 @@ class AnsibleVariablesController < ::LookupKeysController
     redirect_to ansible_variables_path
   end
 
+  def overrides
+    if params[:hostgroup]
+      process_hostgroup
+      @resolver = ForemanAnsible::OverrideResolver.new(@hostgroup)
+    end
+
+    if params[:host]
+      refresh_host
+      @resolver = ForemanAnsible::OverrideResolver.new(@host)
+    end
+
+    return render :json => [] unless @resolver
+
+    render :json => to_response(@resolver)
+  end
+
   private
 
   def default_order; end
@@ -103,5 +120,14 @@ class AnsibleVariablesController < ::LookupKeysController
                   'smart proxy has the Ansible feature enabled.')
     end
     @smart_proxy
+  end
+
+  def action_permission
+    case params[:action]
+    when 'variables', 'overrides'
+      :view
+    else
+      super
+    end
   end
 end
--- a/app/controllers/foreman_ansible/concerns/ansible_variable_overrides.rb
+++ b/app/controllers/foreman_ansible/concerns/ansible_variable_overrides.rb
@@ -1,0 +1,91 @@
+module ForemanAnsible
+  module Concerns
+    module AnsibleVariableOverrides
+      extend ActiveSupport::Concern
+
+      include Foreman::Controller::Parameters::Hostgroup
+      include Foreman::Controller::Parameters::Host
+
+      def to_response(resolver)
+        resolver.ansible_variables.group_by(&:ansible_role).each_with_object([]) do |(role, variables), memo|
+          var_hash = variables.map do |var|
+            {
+              :id => var.id,
+              :parameter => var.key,
+              :override => var.override?,
+              :description => var.description,
+              :parameter_type => var.parameter_type,
+              :hidden_value? => var.hidden_value?,
+              :omit => var.omit,
+              :required => var.required,
+              :validator_type => var.validator_type,
+              :validator_rule => var.validator_rule,
+              :default_value => var.default_value,
+              :override_values => var.lookup_values.map { |lv| { :id => lv.id, :match => lv.match, :value => lv.value, :omit => lv.omit } },
+              :current_value => resolver.resolve(var)
+            }
+          end
+
+          memo.push(:id => role.id, :name => role.name, :ansible_variables => var_hash)
+          memo
+        end
+      end
+
+      def refresh_host
+        @host = Host::Base.authorized(:view_hosts, Host).find_by(:id => params[:host_id] || params.dig(:host, :id))
+        @host ||= Host.new(host_params)
+
+        unless @host.is_a?(Host::Managed)
+          @host      = @host.becomes(Host::Managed)
+          @host.type = 'Host::Managed'
+        end
+        @host.attributes = host_params unless @host.new_record?
+
+        @host.lookup_values.each(&:validate_value)
+        @host
+      end
+
+      def hostgroup_params
+        super 'hostgroup'
+      end
+
+      def host_params
+        super 'host'
+      end
+
+      def process_hostgroup
+        define_parent
+        refresh_hostgroup
+        inherit_parent_attributes
+      end
+
+      def define_parent
+        return unless params[:hostgroup][:parent_id]
+        @parent = Hostgroup.authorized(:view_hostgroups).find(params[:hostgroup][:parent_id])
+      end
+
+      def refresh_hostgroup
+        if params[:hostgroup][:id].present?
+          @hostgroup = Hostgroup.authorized(:view_hostgroups).find(params[:hostgroup][:id])
+          @hostgroup.attributes = hostgroup_params
+        else
+          @hostgroup = Hostgroup.new(hostgroup_params)
+        end
+
+        @hostgroup.lookup_values.each(&:validate_value)
+        @hostgroup
+      end
+
+      def inherit_parent_attributes
+        return unless @parent
+
+        @hostgroup.architecture       ||= @parent.architecture
+        @hostgroup.operatingsystem    ||= @parent.operatingsystem
+        @hostgroup.domain             ||= @parent.domain
+        @hostgroup.subnet             ||= @parent.subnet
+        @hostgroup.realm              ||= @parent.realm
+        @hostgroup.environment        ||= @parent.environment
+      end
+    end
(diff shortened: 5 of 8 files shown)


webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcherConstants.js
```diff
@@ -5,3 +5,8 @@ export const ANSIBLE_ROLES_ADD = 'ANSIBLE_ROLES_ADD';
 export const ANSIBLE_ROLES_REMOVE = 'ANSIBLE_ROLES_REMOVE';
 export const ANSIBLE_ROLES_ASSIGNED_PAGE_CHANGE =
   'ANSIBLE_ROLES_ASSIGNED_PAGE_CHANGE';
+export const ANSIBLE_ROLES_FORM_OBJECT = 'ANSIBLE_ROLES_FORM_OBJECT';
+export const ANSIBLE_VARIABLES_REQUEST = 'ANSIBLE_VARIABLES_REQUEST';
```
ezr-ondrej: There has been a change merged in core recently in API fetches.
see https://github.com/theforeman/foreman/pull/6739 or core storybook `APIMiddleware` documentation on how to use it.

lib/foreman_ansible/engine.rb
```diff
@@ -18,6 +18,7 @@ class Engine < ::Rails::Engine
     config.autoload_paths += Dir["#{config.root}/app/services"]
     config.autoload_paths += Dir["#{config.root}/app/views"]
     config.autoload_paths += Dir["#{config.root}/app/lib"]
+    config.autoload_paths += Dir["#{config.root}/test"]
```
ezr-ondrej: Why do we need this? :(

app/controllers/ui_ansible_roles_controller.rb
```diff
@@ -3,12 +3,34 @@ def resource_name(resource = 'AnsibleRole')
...
   end
 
+  layout 'api/v2/layouts/index_layout', :only => [:index, :variables]
+
   def index
     @ui_ansible_roles = resource_scope_for_index(:permission => :view_ansible_roles)
   end
 
+  def variables
+    @ui_ansible_roles = resource_scope(:permission => :view_ansible_roles).search_for(*search_options).includes(:ansible_variables)
```
ezr-ondrej: I prefer `preload` if you wouldn't mind - it is more clear what we want to achieve. Or is the assoc needed in the search?

ezr-ondrej:
Btw there are some simple fixes in codeclimate and there is too much of test failures (unfortunatelly doesn't seem related - unfortunatelly as we have seriously broken test suite).

adamruzicka:
[test foreman_ansible]

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,67 @@
...
+            :element => 'hostgroup',
+            :element_name =>  hash_result[:hostgroup].name
+          }
+        }
+        memo
+      end
+    end
+
+    def inherited_hostgroup_value(hostgroup, ansvar)
+      return unless ansvar.path_elements.flatten.include?("hostgroup") || Setting["host_group_matchers_inheritance"]
```
ezr-ondrej, line 58: this Setting was renamed to `matchers_inheritance` in theforeman/foreman#7037

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,67 @@
...
+      hostgroup.path.reverse.each_with_object(nil) do |hg, memo|
+        break memo if memo
+        value = LookupValue.find_by(:lookup_key_id => ansvar.id, :id => hg.lookup_values)
+        next memo unless value
+        { :lookup_value => value, :hostgroup => hg }
```
ezr-ondrej, line 63: If I read this correctly, we can use `each` and return here instead of `each_with_object`, rigth?

adamruzicka:
@xprazak2 ping

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,68 @@
+module ForemanAnsible
+  # Service which resolves override values for hosts and hostgroups
+  class OverrideResolver
+    attr_reader :overrides
+
+    def initialize(resource_id, resource_name, parent_id, variables)
```
ezr-ondrej, line 6: why do we initialize with id and resource name, can't we get the model instead?
We could even have a inner factory for Host/Hostgroup, that would take care of the actual resolving.

ezr-ondrej (commented):
Commenting just on the `OverrideResolver`.
I believe we should come up with generic resolver for all the LookupValues.
As now it is quite DIY for plugins, generic resolver would help a lot.

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,68 @@
...
+          }
+        }
+        memo
+      end
+    end
+
+    def inherited_hostgroup_value(hostgroup, ansvar)
+      return unless ansvar.path_elements.flatten.include?('hostgroup') || Setting['matchers_inheritance']
+
+      hostgroup.path.reverse_each do |hg|
```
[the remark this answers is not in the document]
xprazak2: It is possible I missed an edge case and I am not 100% sure since it has been a while since I wrote this but I believe the intent here is to get the value for hostgroup regardless of whether it is inherited or not. That does not mean we cannot have the 'preview' in UI, because we keep track of the hostgroup name this was resolved for, so it should not be difficult to decide if the value is inherited or not.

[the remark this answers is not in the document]
ezr-ondrej: I mean: you have overriden value on hostgroup, in puppet parameters you see in tooltip what is the value of parent. And you know right away what value to show once you disable the override on the hostgroup. with this, we can't show the tooltip and we need API request in case of override get's disabled. IMHO it can be usefull to have both values inherited and the direct override.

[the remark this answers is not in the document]
xprazak2: I agree that we should not do additional requests if they can be avoided, but right now we do not do any. I think this discussion can be had on the follow-up PR if we face that problem.

What are the changes you would like me to do here?

app/services/foreman_ansible/override_resolver.rb
xprazak2: We have [a method](https://github.com/theforeman/foreman_ansible/pull/313/files#diff-58ce55128ecc828dc6de7ef3ee290f1c107020c00d6ec20137386a51cb948070R26) that resolves it. Would you like me to move it to a different place?

xprazak2:
Rubocop is wrong about using alias - it crashes the whole plugin because there is a subtle difference in how alias and alias_method work.

app/services/foreman_ansible/override_resolver.rb
ezr-ondrej: Well the whole premise of resolving variables only based on resource_name and parent is IMHO wrong, so I'd like to have directly the object, I'm explaining why in the review comment.

ezr-ondrej (commented):
Well the whole premise of resolving variables only based on resource_name and parent is wrong.

What we would be saying by it is, that host can only inherit value from hostgroup and that we don't care wheter the value is inherited or not.

BUT we can have custom matchers so the value can actually depend on any Host parameter.
The reason we have https://github.com/theforeman/foreman/blob/7e9415db73841e05cbcf2fc6746806a6341464f8/app/models/hostgroup.rb#L175 is that hostgroup can only inherit from hostgroups, because their attributes are not complete, so we don't try to resolve the actual value the LookupKey would have at the moment the hostgroup is assigned to host and all the attributes are inherited on host (what's arguably wrong if you think about it :shrug:)

Let me add an example how it works with PuppetClassLookupKeys as this is quite complicated and it did take me quite a while to understand.

So we can have a puppet lookup with two matchers `os,comment = CentOS7,ahoj` and `os=CentOS7` with values `a` and `b` respectively.
We have an Hostgroup `parent` with os=`CentOS7`.
Now when we go and create Hostgroup nested under `parent` we will get no inherited value and we can only override it on the hostgroup level (we may argue this is wrong, because the os matcher would actually give host with this hg value `b`).
But when we go to the new host form and select the `parent` hg, we read the inherited value is `b`, because it already set the os from the hg. And when we fill in a comment with `ahoj`, the inherited value gets refreshed to `a` (well would after the values are refetched and that only happens after Hg change, PuppetClass change, Env change, os change and domain change, but when the update is done, the inherited value will be `a`).

So puppet deals with this by sending the whole host form, construct the Host object as it would look like after the host is send (without saving ofc) and then resolves the lookup values.

So **TL;DR the parent concept is useless to determine the actuall correct value** in all cases. It probably covers the most used use-cases, but the matchers would not work as expected.

xprazak2:
> Well the whole premise of resolving variables only based on resource_name and parent is wrong.
> 
> What we would be saying by it is, that host can only inherit value from hostgroup and that we don't care wheter the value is inherited or not.

You may be right, but note I use the existing [method](https://github.com/theforeman/foreman_ansible/pull/313/files#diff-58ce55128ecc828dc6de7ef3ee290f1c107020c00d6ec20137386a51cb948070R36) for [resolving](https://github.com/theforeman/foreman/blob/develop/app/services/classification/values_hash_query.rb#L17) override values, which I believe is still used for resolving Puppet overrides. If you think that resolution is wrong, then we need to fix it, which is out of scope for this PR.

> 
> BUT we can have custom matchers so the value can actually depend on any Host parameter.
> The reason we have https://github.com/theforeman/foreman/blob/7e9415db73841e05cbcf2fc6746806a6341464f8/app/models/hostgroup.rb#L175 is that hostgroup can only inherit from hostgroups, because their attributes are not complete, so we don't try to resolve the actual value the LookupKey would have at the moment the hostgroup is assigned to host and all the attributes are inherited on host (what's arguably wrong if you think about it shrug)

I am pretty sure [inherited_hostgroup_value](https://github.com/theforeman/foreman_ansible/pull/313/files#diff-58ce55128ecc828dc6de7ef3ee290f1c107020c00d6ec20137386a51cb948070R58) is inspired by that.


@ezr-ondrej I have 3 questions:

Does the newly added endpoint serve incorrect override values?
If so, could you share reproducer steps? 
What do I need to do to move this forward?

ezr-ondrej:
> Does the newly added endpoint serve incorrect override values?
> If so, could you share reproducer steps?

Yes it does. In case we have some specific matchers as described in the example, we will not get the correct inherited value in the form (meaning the value that will be actually set in the inventory).

> What do I need to do to move this forward?

1. Using method `vaues_hash` is correct and for edit form this should work fine. The issue is with a new host form.
For a **new host** form we now try to use it's hostgroup for resolving variables and if there is no hostgroup, we don't resolve them.
1. **MAIN** Puppet does that by calling [refresh_host](https://github.com/theforeman/foreman/blob/ac16209da31ea412e64c53767c7a7acbea728790/app/controllers/hosts_controller.rb#L662) with values filled by user in params prior passing this host to the `values_hash` method. This is my main concern.
1. **minor** Resolving for puppet is actually done by the `values_hash` only for the inherited value, the value filled in form is found by `lookup_key.overridden_value(resource)` (because the filled in value isn't saved yet and can't by the `value_hash`) and only if this value doesn't exist, we say the value is inherited and we use the one comming from `values_hash`. But we pass them both to UI so we know what value would the variable have if it would not get overriden from host form (user unchecks the override checkbox). If we would follow this approach and have both values, for hostgroup we could use [`inherited_lookup_value`](https://github.com/theforeman/foreman/blob/ac16209da31ea412e64c53767c7a7acbea728790/app/models/hostgroup.rb#L175) and we wouldn't need our own method.

The last point is probably the one that is causing the most headache and might be irrelevant, as we may not actually need to get the value filled in on the form, because we have that from other sources (original value comes from the Slot and then we can keep the user input in redux store). If this is the case, we would be free to use the `inherited_lookup_value` method for hostgroup anyway though.

app/controllers/ui_ansible_roles_controller.rb
```diff
@@ -3,12 +3,33 @@ def resource_name(resource = 'AnsibleRole')
...
   def index
     @ui_ansible_roles = resource_scope_for_index(:permission => :view_ansible_roles)
   end
 
+  def variables
+    @ui_ansible_roles = resource_scope(:permission => :view_ansible_roles).search_for(*search_options).preload(:ansible_variables)
+    @override_resolver = ForemanAnsible::OverrideResolver.new(params[:resource_id],
+                                                              params[:resource_name],
+                                                              params[:parent_id],
+                                                              AnsibleVariable.where(:ansible_role_id => @ui_ansible_roles.pluck(:id)))
```
ezr-ondrej, line 17: In addition this is IMHO also wrong. We resolve this for all variables, regardles if they are used or not. We should probably resolve only for host(group)s roles. That's just one more reason to use the `refresh_host`/`refresh_hostgroup` methods. Then the `@resource.all_ansible_roles` will give us the list of roles we want to resolve the variables for.

Loading the variables only once and then show only the ones selected in UI would probably make sense, but we need to reload after almost every change to the host anyway, so it IMHO doesn't justify loading all the variables. Just refetch them also after change in role assignment :shrug:

ezr-ondrej:
:police_officer: is unhappy...

xprazak2:
Updated, main changes:

* resolver now expects host or hostgroup object, which gives it access to all attributes and the resolution should work correctly for persisted as well as new (not yet saved) objects.

* I shamelessly copied the 'refresh' methods that set host(group) attributes from core controllers, it is a mess.

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -17,6 +17,10 @@ def self.prepended(base)
 
         before_provision :play_ansible_roles
         audit_associations :ansible_roles
+
+        # rubocop:disable Style/Alias
+        alias_method :parent, :hostgroup
+        # rubocop:enable Style/Alias
```
ezr-ondrej, line 23: I believe this is not used now, right?

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,56 @@
...
+            :element => 'hostgroup',
+            :element_name => hash_result[:hostgroup].name
+          }
+        }
+        memo
+      end
+    end
+
+    def inherited_hostgroup_value(hostgroup, ansvar)
+      return unless ansvar.path_elements.flatten.include?('hostgroup') || Setting['matchers_inheritance']
```
ezr-ondrej, line 45:
```suggestion
      return unless ansvar.path_elements.flatten.include?('hostgroup') && Setting['matchers_inheritance']
```
But I believe we can just use the `Hostgroup#inherited_lookup_value`

ezr-ondrej (commented):
Awesome @xprazak2 !
I'd not worry about copying the methods, they are as you say quite a mess and we should extract them to separate module in core afterwards.

My last point (**3.**) is still not resolved, and I believe we have two ways to go about it:
1. don't care about the actual object values and expect the UI (redux) have *them* values which is easier and cleaner and we get the initial values through the Slot. 
It would mean using [`inherited_values`](https://github.com/theforeman/foreman/blob/ed70741facd8c57b76eb42be350173560b75b086/app/services/classification/values_hash_query.rb#L7) instead of `values_hash` and `Hostgroup#inherited_lookup_value` instead of our method for resolving hg values.
It is not straight forward solution, but works best IMHO, so I would prefer that, with some comment documenting this on the controller action level.
2. care about the object values, in which case we would need to resolve the actual values through `lookup_key.overridden_value(resource)` that takes in account unsaved values, but we would need to combine inherited values and the object values and that might be more difficult then neccessary.
If you want to go this path, take a look at method [`PuppetLookupKeyHelper#lookup_key_with_diagnostic`](https://github.com/theforeman/foreman/blob/ed70741facd8c57b76eb42be350173560b75b086/app/helpers/lookup_keys_helper.rb#L54) that is doing exactly what we would need to do. But be aware this rabbit hole is QUITE deep :laughing:

app/services/foreman_ansible/override_resolver.rb
xprazak2: > My last point (**3.**) is still not resolved, and I believe we have two ways to go about it:

Valid point, it will be resolved once we add the actual changes in UI. I am looking forward to discussing it in https://github.com/theforeman/foreman_ansible/pull/376 once I rebase on these changes.

xprazak2:
> My last point (**3.**) is still not resolved, and I believe we have two ways to go about it:

Valid point, it will be resolved once we add the changes for UI. I am looking forward to discussing it in https://github.com/theforeman/foreman_ansible/pull/376 once we get this in.

ezr-ondrej (approved):
Agreed, this lays solid ground for the UI PR, let's make necessary changes in UI PR.
Thanks @xprazak2 ! :+1:

PR #314, opened 2019-12-02
Fixes #28388 - Provide a feature for enabling web console

adamruzicka:
Companion to:
- [x] https://github.com/theforeman/community-templates/pull/668

### DIFF
--- a/lib/foreman_ansible/remote_execution.rb
+++ b/lib/foreman_ansible/remote_execution.rb
@@ -36,6 +36,12 @@ module ForemanAnsible
         :description => N_('Run an Ansible playbook against given hosts'),
         :provided_inputs => %w[playbook]
       )
+      RemoteExecutionFeature.register(
+        :ansible_enable_web_console,
+        N_('Enable web console'),
+        :description => N_('Run an Ansible playbook to enable web console on given hosts'),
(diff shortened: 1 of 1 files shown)


lzap:
Let's put this one on hold, I haven been informed in the YT video comment (https://www.youtube.com/watch?v=Ua3b79UG9JM) that Cockpit does NOT need to be enabled on the target host actually. Let's confirm that.

ezr-ondrej:
I believe the `cockpit-system` needs to be installed. That runs the socket. I believe that the comment might have refered to the cockpit UI, that indeed doesn't need to be enabled on the target server, but the communication socket must be installed I believe.

lzap:
Ok so the template has been merged, therefore we want this one as well? How does this work actually, the button calls the SSH action?

lzap:
I have tested the template itself, but how does this match the template with the button? Using a template name? That's actually "Service Action - Enable Web Console".

adamruzicka:
The button is for a specific feature, feature has a template assigned.

In the template it uses the feature property of the template https://github.com/theforeman/community-templates/pull/668/files#diff-490661b6bb91c2a4d49d91d72b46ebfbR8 and matches it against the feature name.

PR #317, opened 2019-12-11
Fix rubocop

ares:


ares:
needed to fix one more cop

adamruzicka:
Were templates synced to core recently?

ares:
Yeah these tests are broken for a while, we can ignore them right now. Thanks for looking, I'll merge this one to unblock rubocop.

PR #318, opened 2019-12-12
Fixes #28463 - Fix missing roles in the assigned roles list

amasolov:
Needed to populate directlyAssignedRoles with allAssignedRoles instead of assignedRoles to get the whole list of roles not just visible ones.

### DIFF
--- a/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js
+++ b/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js
@@ -89,6 +89,7 @@ class AnsibleRolesSwitcher extends React.Component {
             </div>
             <AssignedRolesList
               assignedRoles={assignedRoles}
+              allAssignedRoles={allAssignedRoles}
               pagination={assignedPagination}
               itemCount={assignedRolesCount}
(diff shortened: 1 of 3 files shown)


amasolov:
Fixed my own mess with commits. Should be fine now.

amasolov:
Looks like this PR also fixes [#27678](https://projects.theforeman.org/issues/27678). Can't reproduce the described problem after applying this change.

xprazak2:
Good catch. You need to [pass](https://github.com/theforeman/foreman_ansible/blob/master/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js#L90) `allAssignedRoles` as a prop to the `AssignedRolesList` to propagate it from a parent component though:

```js
<AssignedRolesList
  allAssignedRoles={allAssignedRoles}
//other props
/>
```

amasolov:
> You need to [pass](https://github.com/theforeman/foreman_ansible/blob/master/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js#L90) `allAssignedRoles` as a prop to the `AssignedRolesList`

Yup, somehow missed that in my commit.

xprazak2:
Thanks!

PR #321, opened 2019-12-19
Drop tasks dependency

ares:
With foreman-tasks 1.0.0, this would no longer be compatible. However 1.0.0 bump only means, tasks now use vendor 3. Also I think we don't need to depend on this anymore since we depend on REX.

### DIFF
--- a/foreman_ansible.gemspec
+++ b/foreman_ansible.gemspec
@@ -22,7 +22,6 @@ Gem::Specification.new do |s|
   # Kept as a dev dependency so tests can run together
   s.add_development_dependency 'foreman_ansible_core', '~> 3.0'
   s.add_dependency 'deface', '< 2.0'
-  s.add_dependency 'foreman-tasks', '~> 0.8'
-  s.add_dependency 'foreman_remote_execution', '>= 2.0.0', '< 3.0'
+  s.add_dependency 'foreman_remote_execution', '>= 2.0.0'
   s.add_dependency 'ipaddress', '>= 0.8.0', '< 1.0'
 end


foreman_ansible.gemspec
```diff
@@ -22,7 +22,6 @@ Gem::Specification.new do |s|
   # Kept as a dev dependency so tests can run together
   s.add_development_dependency 'foreman_ansible_core', '~> 3.0'
   s.add_dependency 'deface', '< 2.0'
-  s.add_dependency 'foreman-tasks', '~> 0.8'
   s.add_dependency 'foreman_remote_execution', '>= 2.0.0', '< 3.0'
```
mmoll: This will need an adjustment too, no?

adamruzicka: It will

ares:
Yeah, updated. Given REX doesn't change inside dramatically, I dropped the '< 3.0' part, let me know what you think

ezr-ondrej:
> Yeah, updated. Given REX doesn't change inside dramatically, I dropped the '< 3.0' part, let me know what you think

I find the open dependencies confusing. It can bite your tale, i.e. when you introduce breaking change, it can have impact on debian systems, right?

ares:
It can have have impact elsewhere too. Most of the time when we bump rex major version, it's due to we bump Foreman version dependency. And we then need to update all rex dependencies, like here now. I don't insist, but this tends to break nightlies and I don't recall when we last time caught any regression caused by incompatibility with rex.

PR #323, opened 2020-01-02
Fixes #28612 - Fix runs when proxy batch triggering is disabled

adamruzicka:
Before this change, the playbook was interpreted as a shell script on the proxy side because a default action class from REX was being used.

requires:
- [x] https://github.com/theforeman/foreman_remote_execution/pull/457

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -62,6 +62,10 @@ if defined? ForemanRemoteExecution
           'ansible-runner'
         end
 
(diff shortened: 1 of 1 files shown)


adamruzicka:
[test foreman_ansible]

adamruzicka:
[test foreman_ansible]

adamruzicka:
The tests should be fixed by https://github.com/theforeman/foreman_ansible/pull/360 , could we get this in?

ezr-ondrej:
Thanks @adamruzicka !

PR #324, opened 2020-01-07
Fixes #28662 - Remove VariableLookupKey reference

xprazak2:


app/controllers/api/v2/ansible_variables_controller.rb
```diff
@@ -5,7 +5,6 @@ module V2
     # API controller for Ansible Variables
     class AnsibleVariablesController < ::Api::V2::BaseController
       include ::Api::Version2
-      include Foreman::Controller::Parameters::VariableLookupKey
```
tbrisker: `variable_lookup_key_params` from this concern is used on line 67

xprazak2: Good catch, updated.

tbrisker: the fact this didn't cause tests to fail makes me wonder if we actually have any test covering this endpoint, and looks like we don't - can you please add a simple test to make sure the update works after the change in the strong params filtering?

PR #325, opened 2020-02-05
Fixes #28937 - ansible variable type form field

ezr-ondrej:
On foreman 2.0 we go to the else branch and render old method.


ares:
Thanks @ezr-ondrej, merging!

PR #330, opened 2020-02-24
Fixes #29138 - Fix rendering overriden False variables in YAML output

amasolov:
"if value" checks if value is either true or not nil. So the test is false if the value is false.
Whereas "unless value.nil?" checks if the value not nil.


ezr-ondrej:
ok to test

xprazak2:
Looks good. @amasolov could you rebase on the latest master please? It should make the test failures go away.

amasolov:
Rebased on the latest master but it's still failing

xprazak2:
[test foreman_ansible]

xprazak2:
Thanks!

PR #335, opened 2020-03-12
Fixes #28859, #26485 - ansible import variables uniquely per role

ezr-ondrej:
- Makes imported variables unique per role, not globally
- Adds override variable operation
- Changes meaning of Update operation

For disscusion please followup [discourse thread](https://community.theforeman.org/t/ansible-variables-update-on-import/17753).

Add:
- [x] Tests
- ~~[ ] API~~

### DIFF
--- a/app/services/foreman_ansible/variables_importer.rb
+++ b/app/services/foreman_ansible/variables_importer.rb
@@ -48,22 +48,21 @@ module ForemanAnsible
     def initialize_variables(variables, role)
       variables.map do |variable_name, variable_default|
         variable = AnsibleVariable.find_or_initialize_by(
-          :key => variable_name
+          :key => variable_name,
+          :ansible_role_id => role.id
         )
-        if variable.new_record?
-          variable.assign_attributes(:default_value => variable_default,
-                                     :key_type => infer_key_type(variable_default),
-                                     :imported => true)
-        end
-        variable.ansible_role = role
+        variable.assign_attributes(:default_value => variable_default,
+                                   :key_type => infer_key_type(variable_default))
+        variable.imported = true if variable.new_record?
         variable.valid? ? variable : nil
       end
     end
 
     def detect_changes(imported)
       changes = {}.with_indifferent_access
-      persisted, changes[:new] = imported.partition { |role| role.id.present? }
-      changes[:update], _old = persisted.partition(&:changed?)
+      persisted, changes[:new] = imported.partition { |var| var.id.present? }
+      changed, _old = persisted.partition(&:changed?)
+      _overriden, changes[:update] = changed.partition(&:override?)
       changes[:obsolete] = AnsibleVariable.where.not(:id => persisted.pluck(:id), :imported => false)
       changes
     end
--- a/test/unit/services/ansible_variables_importer_test.rb
+++ b/test/unit/services/ansible_variables_importer_test.rb
@@ -5,52 +5,77 @@ require 'test_plugin_helper'
 # This service is meant to take in essentially a string coming from
 # the proxy API, and parse that into AnsibleVariables.
 class AnsibleVariablesImporterTest < ActiveSupport::TestCase
-  setup do
-    @importer = ForemanAnsible::VariablesImporter.new
-  end
-  test 'does not reimport already existing variables' do
-    already_existing = FactoryBot.create(:ansible_variable)
-    new_role = FactoryBot.create(:ansible_role)
-    api_response = {
-      new_role.name => { 'new_var' => 'new value' },
-      already_existing.ansible_role.name => { already_existing.key => already_existing.default_value }
-    }
-    changes = @importer.import_variables(api_response, [new_role.name])
-    assert_not_empty changes['new']
-    assert_equal 'new_var', changes['new'].first.key
-    assert_equal new_role, changes['new'].first.ansible_role
-  end
+  let(:importer) { ForemanAnsible::VariablesImporter.new }
 
-  test "variables attempts to remove variables that don't exist anymore" do
-    obsolete_variable = FactoryBot.create(:ansible_variable)
-    changes = @importer.import_variables({}, [])
-    assert_not_empty changes['obsolete']
-    assert_equal obsolete_variable.key, changes['obsolete'].first.key
-    assert_equal(
-      obsolete_variable.ansible_role,
-      changes['obsolete'].first.ansible_role
-    )
-  end
+  describe '#import_variables' do
+    test 'does not reimport already existing variables' do
+      already_existing = FactoryBot.create(:ansible_variable)
+      new_role = FactoryBot.create(:ansible_role)
+      api_response = {
+        new_role.name => { 'new_var' => 'new value' },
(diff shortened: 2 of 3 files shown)


app/services/foreman_ansible/variables_importer.rb
```diff
@@ -87,6 +87,10 @@ def create_new_variables(variables)
       end
     end
 
+    def override_variables(variables)
+      update_variables(variables).each { |var| var.update(:override => false) }
```
xprazak2: You will need to move `memo` from l.131 to l.133 for this to work.

More importantly, I do not think we want to undo the overrides, because that will disable matchers users may have set up. Maybe https://projects.theforeman.org/issues/26485 could be a way forward?

ezr-ondrej: @xprazak2, this works I believe. It iterates over the updated results.
It really does, that's what it should do. It's called override, so it override the variable by fresh import :upside_down_face:
But as I'm saying in the post, I'm open to drop this concept :+1: 
As of the issue that's how update behaves now :) so I accidentally addressed the issue.

xprazak2: I run into, it does not seem like my Rails server is convinced :slightly_smiling_face:

import-override

ezr-ondrej:
Ok, I'll address the override as followup, let's first decide about the update and adding uniquelly per roles :)

test/unit/services/ansible_variables_importer_test.rb
```diff
@@ -5,52 +5,64 @@
...
-    changes = @importer.import_variables(api_response, [])
-    assert_empty changes['update']
-    assert_equal 'new_variable', changes['new'].first.key
-  end
+    test 'does not do anything if response is empty' do
+      changes = importer.import_variables({}, [])
+      assert_equal({ 'new' => [], 'obsolete' => [], 'update' => [] }, changes)
+    end
+
+    test 'update changed defaults' do
```
ezr-ondrej, line 41: ^ this

test/unit/services/ansible_variables_importer_test.rb
```diff
@@ -5,52 +5,64 @@
...
+      variable = FactoryBot.create(:ansible_variable, :default_value => 'default value', :ansible_role => role)
+      api_response = {
+        role.name => { variable.key => 'changed value' }
+      }
+      changes = importer.import_variables(api_response, [])
+      assert_not_empty changes['update']
+      assert_equal variable.key, changes['update'].first.key
+    end
+
+    test 'does not update overriden defaults' do
```
ezr-ondrej, line 52: and ^ this

test/unit/services/ansible_variables_importer_test.rb
```diff
@@ -5,52 +5,77 @@
...
+      role = FactoryBot.create(:ansible_role)
+      variable = FactoryBot.create(:ansible_variable, :default_value => 'default value', override: true, :ansible_role => role)
+      api_response = {
+        role.name => { variable.key => 'changed value' }
+      }
+      changes = importer.import_variables(api_response, [])
+      assert_empty changes['update']
+    end
+
+    test 'reimports variable with same key for different role' do
```
ezr-ondrej, line 62: and this ^

xprazak2:
[test foreman_ansible]

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -48,22 +48,21 @@ def import_new_role(role_name, new_roles)
...
       end
     end
 
     def detect_changes(imported)
       changes = {}.with_indifferent_access
-      persisted, changes[:new] = imported.partition { |role| role.id.present? }
-      changes[:update], _old = persisted.partition(&:changed?)
+      persisted, changes[:new] = imported.partition { |var| var.id.present? }
+      changed, _old = persisted.partition(&:changed?)
+      _overriden, changes[:update] = changed.partition(&:override?)
```
xprazak2, line 65: I am not opposed to it, but why not to update the default value for overriden variables? Should this fix [#26485](https://projects.theforeman.org/issues/26485)?

xprazak2:
Nice progress! Variables with the same names are now imported as expected.

I just do not see the part that should fix [#26485](https://projects.theforeman.org/issues/26485)...

app/services/foreman_ansible/variables_importer.rb
ezr-ondrej: Well, because if the value has `override` flag, the `default_value` is set by user :thinking:
So I left those updates for a followup. Such update would effectively mean `reimport the variable and delete my changes to it` I just thought it's too dangerous :)

ezr-ondrej: So let's do it in next PR.

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -48,22 +48,21 @@ def import_new_role(role_name, new_roles)
...
+          :key => variable_name,
+          :ansible_role_id => role.id
         )
-        if variable.new_record?
-          variable.assign_attributes(:default_value => variable_default,
-                                     :key_type => infer_key_type(variable_default),
-                                     :imported => true)
-        end
-        variable.ansible_role = role
+        variable.assign_attributes(:default_value => variable_default,
```
ezr-ondrej, line 54: @xprazak2 notice this line. Previously, the assignment was happening only `if new_record?` now it's happening every time, so it solves [#26485](https://projects.theforeman.org/issues/26485) as it reimports `default_value`

ezr-ondrej: (then the change assesment has totally different meaning obviously, as it asses change of value instead of change of `role_id`)

xprazak2: [#26485](https://projects.theforeman.org/issues/26485) describes a selective import based whether the default value has been changed in foreman or not, that is why I was confused and do not think this change actually implements the behaviour that is described there. 

I am willing to merge as is, because it fixes the import problems for vars with the same name.

xprazak2:
Robocop is failing...

app/services/foreman_ansible/variables_importer.rb
ezr-ondrej: It does solve the issue I believe. This line imports to existing variable and https://github.com/theforeman/foreman_ansible/pull/335/files#diff-e2db0db7ab2083dffdd6509a1e79702fR64 throw away the unchanged ones. So it imports all variables, but doesn't show up in importer if there is no change. That is exactly what the issue is suggesting, right?

ezr-ondrej:
> Robocop is failing...

Hehe, right. I really hate our Hash :rocket:ing, rebased on the fixed rubocop, should pass now.

ezr-ondrej:
it's :green_heart:ish xD

app/services/foreman_ansible/variables_importer.rb
xprazak2: Oh, I see it now, thanks for bearing with me.

xprazak2:
Thanks!

PR #336, opened 2020-03-12
Fixes #29336 - Handle non-string formatting characters in playbook

adamruzicka:


ezr-ondrej:
Sorry for being late for the party :)

PR #337, opened 2020-03-13
Fixes #29343 - update Ansible Inventory template name

ares:
Foreman 2.1 will probably contain a change that renames the inventory
report template, thanks to the mass renaming in https://github.com/theforeman/community-templates/pull/680

We have the name set as the default value in Settings. For people who
didn't change it, changing the default will do. For people who changed
the value to some other template, their value remains unchanged, which
is expected. No more actions are needed therefore.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -78,7 +78,7 @@ class Setting
             'ansible_inventory_template',
             N_('Foreman will use this template to schedule the report '\
                'with Ansible inventory'),
-            'Ansible Inventory',
+            'Ansible - Ansible Inventory',
             N_('Default Ansible inventory report template')
           )
         ]


ares:
This should only be merged when https://github.com/theforeman/community-templates/pull/680 gets synced to Foreman core. We'll need to bump foreman min version accordingly, most likely to `2.1`

ezr-ondrej:
I guess we wait for theforeman/foreman#7532 as well?

adamruzicka:
> I guess we wait for theforeman/foreman#7532 as well?

This is now in.

[test foreman_ansible]

ezr-ondrej (approved):
So get this in?

ezr-ondrej:
This didn't get into the repository for some reason and only got fixed in #342

PR #341, opened 2020-04-14
Fixes #29528 - rubocop task definition

ezr-ondrej:


lib/tasks/foreman_ansible_tasks.rake
```diff
@@ -13,19 +13,15 @@ namespace :test do
...
-                         "#{ForemanAnsible::Engine.root}/test/**/*.rb"]
-      end
-    rescue StandardError
-      puts 'Rubocop not loaded.'
+  begin
+    require 'rubocop/rake_task'
+    RuboCop::RakeTask.new(:rubocop) do |task|
+      task.patterns = ["#{ForemanAnsible::Engine.root}/app/**/*.rb",
+                       "#{ForemanAnsible::Engine.root}/lib/**/*.rb",
+                       "#{ForemanAnsible::Engine.root}/test/**/*.rb"]
```
ezr-ondrej, line 21: shouldn't this be `Include: ...` in `.rubocop.yml`?

xprazak2: It might be worth moving it to `.rubocop.yml` so that we are consistent across plugins...

ezr-ondrej: I've tried that, but it seems, that it doesn't work for the rake task actually :disappointed:

xprazak2:
codeclimate is failing, should be an easy fix

ezr-ondrej:
> codeclimate is failing, should be an easy fix - funny thing is that rubocop on jenkins is passing...

This was introduced in ruby 2.5. I didn't want to go 2.4 incompatible just to make codeclimate happy :thinking: but I guess with Rails 6 now we are ok to do that :)

xprazak2:
Thanks!

PR #342, opened 2020-04-21
Fixes #29587 - Update default template name for Ansible inventory

ofedoren:
@ares, I've tried to use default (changed of course) value without migration and it just worked on my setup. So, not sure if that's needed.

### DIFF
--- a/app/models/setting/ansible.rb
+++ b/app/models/setting/ansible.rb
@@ -78,7 +78,7 @@ class Setting
             'ansible_inventory_template',
             N_('Foreman will use this template to schedule the report '\
                'with Ansible inventory'),
-            'Ansible Inventory',
+            'Ansible - Ansible Inventory',
             N_('Default Ansible inventory report template')
           )
         ]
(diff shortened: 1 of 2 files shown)


ares:
Can you try one thing pls? Set it to some other template and then swtich it back to the original template, then change the default. If that stores the value (instead of the setting being `nil`) I think then the migration would be needed. But you're right, if user didn't change it at all, just changing default should be enough.

ofedoren:
@ares. Well, if someone explicitly saved the old default (Ansible Inventory) and then updated foreman_ansible with this patch, then the migration is needed to update the saved value to the new one. I've wrote the migration the way that it changes only the old default. If user has other template assigned, then it should remain untouched.

ares:
I'm sorry I misunderstood your comment and didn't look at the code, yeah what you have here should be enough. I just need to test. Thanks so far!

ares:
Thanks @ofedoren, merging!

PR #343, opened 2020-04-27
Fixes #29627 - fixed ansible-runner verbosity in plays

snagoor:
With this patch, Ansible execution through foreman shows debug/verbose output as per the `ansible_verbosity` setting.

~~~
 1: ansible-playbook 2.8.10
   2:   config file = /usr/share/foreman-proxy/.ansible.cfg
   3:   configured module search path = [u'/usr/share/foreman-proxy/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
   4:   ansible python module location = /usr/lib/python2.7/site-packages/ansible
   5:   executable location = /usr/bin/ansible-playbook
   6:   python version = 2.7.5 (default, Sep 26 2019, 13:23:47) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
   7: Using /usr/share/foreman-proxy/.ansible.cfg as config file
   8: host_list declined parsing /tmp/d20200427-20155-1ddcn1c/inventory/hosts as it did not pass it's verify_file() method
   9: Parsed /tmp/d20200427-20155-1ddcn1c/inventory/hosts inventory source with script plugin
  10:

  11: PLAYBOOK: playbook.yml *********************************************************
  12: 1 plays in playbook.yml
  13:

  14: PLAY [Patching] ****************************************************************
  15:

  16: TASK [Gathering Facts] *********************************************************
  17: task path: /tmp/d20200427-20155-1ddcn1c/project/playbook.yml:2
  18: <client.example.com> ESTABLISH SSH CONNECTION FOR USER: root
  19: <client.example.com> SSH: EXEC ssh -o ProxyCommand=none -o StrictHostKeyChecking=no -o Port=22 -o 'IdentityFile="/usr/share/foreman-proxy/.ssh/id_rsa_foreman_proxy"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="root"' -o ConnectTimeout=10 client.example.com '/bin/sh -c '"'"'echo ~root && sleep 0'"'"''
~~~

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -10,6 +10,7 @@ module ForemanAnsibleCore
         @inventory = rebuild_secrets(rebuild_inventory(input), input)
         @playbook = input.values.first[:input][:action_input][:script]
         @root = working_dir
+        @verbosity_level = input.values.first[:input][:action_input][:verbosity_level]
       end
 
       def start
@@ -99,8 +100,17 @@ module ForemanAnsibleCore
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        command << verbosity if verbose?
         initialize_command(*command)
         logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
+      end
+
+      def verbosity
+        '-' + 'v' * @verbosity_level.to_i
+      end
+
+      def verbose?
+        @verbosity_level.to_i.positive?
       end
 
       def prepare_directory_structure


lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,28 @@ def write_playbook
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        setup_verbosity(command) if verbose?
         initialize_command(*command)
         logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
       end
 
+      def setup_verbosity(command)
+        verbosity_level = @verbosity_level.to_i
```
ezr-ondrej, line 109:
```suggestion
        verbosity_level = [@verbosity_level.to_i, 4].min
```

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,28 @@ def write_playbook
...
+        verbosity = '-'
```
ezr-ondrej, line 110:
```suggestion
        verbosity = '-' + verbosity_level * 'v'
```

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,28 @@ def write_playbook
...
+          verbosity += 'v'
+        end
+        command << verbosity
+      end
+
+      def verbose?
+        verbosity_level = @verbosity_level
+        # rubocop:disable Rails/Present
+        !verbosity_level.nil? && !verbosity_level.empty? &&
+          verbosity_level.to_i.positive?
```
ezr-ondrej, line 121: If we dont care about explicit 0 being `== nil == ''` what we, as I read, don't, we could simplify and use only one liner method (as `''.to_i == nil.to_i == 0`):
```suggestion
          @verbosity_level.to_i.positive?
```

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,28 @@ def write_playbook
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        setup_verbosity(command) if verbose?
         initialize_command(*command)
         logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
       end
 
+      def setup_verbosity(command)
```
ezr-ondrej, line 108: what about `def verbosity_option`

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,28 @@ def write_playbook
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        setup_verbosity(command) if verbose?
```
ezr-ondrej, line 103: I do not like the name `setup` too much xD WDYT about this:
```suggestion
        command << verbosity_option if verbose?
```

snagoor: @ezr-ondrej thank you so much for the overall feedback. With all your suggestions embedded, my commit looks more clean and tweaked. :green_heart: 

I tested this on a Satellite 6.7 server and it works :thumbsup:

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -99,10 +100,19 @@ def write_playbook
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        command << verbosity_option if verbose?
         initialize_command(*command)
         logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
       end
 
+      def verbosity_option
+        '-' + 'v' * @verbosity_level.to_i
```
ezr-ondrej, line 109: what happens if there is `@verbosity_level > 5`? can ansible deal with `-vvvvv`? won't it fail? Wouldn't it be better to limit it?
```suggestion
        '-' + 'v' * [@verbosity_level.to_i,4].min
```

snagoor: @ezr-ondrej verbosity_level is the value derived from Ansible Settings https://github.com/theforeman/foreman_ansible/blob/master/app/models/setting/ansible.rb#L39-L52 which can't be more than 4. This value is exposed during the Ansible run https://github.com/theforeman/foreman_ansible/blob/1e040a065a97daa84d1a70cb90c396ec02f43e80/app/models/foreman_ansible/ansible_provider.rb#L26 So, its safe to assume that `verbosity_level` would always in the range of 0-4 (where 0 means disabled).

ezr-ondrej: :+1: I'm ok with that, just it's always good to be on the safe side, in case we allow some other places for the same input, but no biggie :)

ezr-ondrej:
Works smoothly, thanks @snagoor !

PR #344, opened 2020-05-07
Refs #29587 - Correct setting migration

tbrisker:
In case the setting isn't present when running the migration, migration
will fail. Also no need to update the value if it is the default - it
would be saved as nil in the database and automatically update to the
new default value.
Added down migration as well for returning it to the previous default.

### DIFF
--- a/db/migrate/20200421201839_update_ansible_inv_template_name.rb
+++ b/db/migrate/20200421201839_update_ansible_inv_template_name.rb
@@ -1,9 +1,9 @@
 class UpdateAnsibleInvTemplateName < ActiveRecord::Migration[5.2]
   def up
-    setting = Setting.find_by(:name => 'ansible_inventory_template')
-    setting.update(:default => 'Ansible - Ansible Inventory')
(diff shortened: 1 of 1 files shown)


tbrisker:
@ofedoren please take a look

tbrisker:
found this thanks to failing build tests: https://ci.theforeman.org/job/test_plugin_matrix/2008/database=postgresql,ruby=2.5,slave=fast/console

ezr-ondrej:
What happens in case user saves the value with the default? Will it still be `nil`?

tbrisker:
> What happens in case user saves the value with the default? Will it still be `nil`?

yes, we have a callback that reset it:

https://github.com/theforeman/foreman/blob/eab834375cc319e46729311e89b6bec95a7918a7/app/models/setting.rb#L367-L371

ezr-ondrej:
Jenkins failed on 2.6, but IMO can't be related to this PR.
Thanks @tbrisker and @ofedoren !

PR #346, opened 2020-05-13
Fixes #29784 - Use InternalProxy selector for capsule upgrade

adamruzicka:
Requires
- [x] https://github.com/theforeman/foreman_remote_execution/pull/490
- [x] https://github.com/theforeman/community-templates/pull/731

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -62,6 +62,14 @@ if defined? ForemanRemoteExecution
           'ansible-runner'
         end
 
+        def required_proxy_selector_for(template)
+          if template.remote_execution_features.where(:label => 'ansible_run_capsule_upgrade').any?
+            ::InternalProxyProxySelector.new
+          else
+            super
+          end
+        end
+
         private
 
         def ansible_command?(template)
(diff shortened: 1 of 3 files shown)


adamruzicka:
Note to myself: needs a version bump on REX once the rex PR gets merged

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -62,6 +62,14 @@ def proxy_operation_name
           'ansible-runner'
         end
 
+        def required_proxy_selector_for(template)
+          if template.remote_execution_features.where(:label => 'ansible_run_capsule_upgrade').any?
+            ::InternalProxyProxySelector.new
```
ehelms, line 67: Where is this class defined?

adamruzicka: And of course I forgot to add the new file to the commit. It is defined in the REX PR this one depends on.

https://github.com/theforeman/foreman_remote_execution/pull/490/files#diff-00044ee009a2a6f328c8251ee0dc386d

lib/foreman_ansible/remote_execution.rb
```diff
@@ -36,6 +36,11 @@ def self.register_rex_feature
         :description => N_('Run an Ansible playbook against given hosts'),
         :provided_inputs => %w[playbook]
       )
+      RemoteExecutionFeature.register(
+        :ansible_run_capsule_upgrade,
+        N_('Upgrade capsules on given hosts'),
+        :description => N_('Upgrade capsules on given capsule server hosts')
```
ehelms, line 42: I think in both cases Capsules should be uppercase since we treat it as the proper name for a thing.

PR #350, opened 2020-07-13
Fixes #30384 - Fix ansible secrets to support effective_user_password

patilsuraj767:


patilsuraj767:
This change is required because of - https://github.com/theforeman/foreman_remote_execution/pull/508

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -35,19 +35,14 @@ def secrets(host)
...
-                'ansible_sudo_pass' => rex_sudo_password(host)
+                'ansible_become_password' => rex_effective_user_password(host)
               }
             }
           }
         end
 
-        def rex_ssh_password(host)
-          host_setting(host, 'remote_execution_ssh_password')
-        end
```
adamruzicka, line 47: Do we really want to do this? The ssh password can still be different from the effective user one

patilsuraj767: Yeah, I should only change sudo_pass. Got confused in ansible_ssh_pass and ansible_su_pass. :P 
But I think we should change ansible variable from "ansible_sudo_pass" to "ansible_become_password", it will help to deal with both sudo and su user method.
I will add the ansible_ssh_pass again.

adamruzicka:
The changes in rex will be released in 4.0.0, could you bump the versions in gemspec here?

test/unit/ansible_provider_test.rb
```diff
@@ -35,13 +35,13 @@ class AnsibleProviderTest < ActiveSupport::TestCase
       it 'generates secrets properly' do
         params = {
           'remote_execution_ssh_password' => 'password',
-          'remote_execution_sudo_password' => 'letmein'
+          'ansible_become_password' => 'password'
```
ezr-ondrej, line 38: why it matches the other password, it would rather not, for the test to catch if we are assigning them incorrectly.

patilsuraj767: Didn't get your point, but do you mean that it should be ` 'remote_execution_effective_user_password'` instead of `'ansible_become_password'`?

ezr-ondrej: no, that's not what I mean.
We've kept the `remote_execution_ssh_password => 'password'` and we use `password` for `ansible_become_password` as well, we should be able to distinquish each of the value in asserts and now we are not.
But as I said, it's a nitpick :)

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -46,8 +46,8 @@ def rex_ssh_password(host)
           host_setting(host, 'remote_execution_ssh_password')
         end
 
-        def rex_sudo_password(host)
-          host_setting(host, 'remote_execution_sudo_password')
+        def rex_effective_user_password(host)
+          host_setting(host, 'remote_execution_effective_user_password')
```
ezr-ondrej, line 50: Doesn't this imply a migration of old settings is necessary?
We need migration to rename setting like this:
```
Setting.where(name: 'remote_execution_sudo_password').update_all(name: 'remote_execution_effective_user_password')
```

patilsuraj767: @ezr-ondrej Migration is written in this PR - https://github.com/theforeman/foreman_remote_execution/pull/508

ezr-ondrej: Right :+1: Thanks for the pointer, missed that out. In that case no more picks.

ezr-ondrej:
@adamruzicka this should get in before release for foreman 2.2, right?

adamruzicka:
Ideally, yes. The changes in rex will go out in 4.0.0, which will be the first >=2.2 release

tbrisker:
Looks like one of these tests is now failing: https://ci.theforeman.org/job/test_plugin_matrix/2441/

adamruzicka:
Being handled here https://github.com/theforeman/foreman_ansible/pull/360

PR #351, opened 2020-07-14
Fixes #30398 - Add DSL docs for ansible

ofedoren:


app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -67,6 +67,11 @@ def import_host(*args)
 
 module Host
   class Managed
+    apipie :class do
```
xprazak2, line 70: Curiously enough, the following three do not show up. Is it just me? Changes to ansible_role and renderer_methods work as expected.

ofedoren: I do see them.. 
ScreenShot-1594815388668

They are added to `Host Managed` class as properties. Have you re-cached docs?

ares: I see them too

PR #352, opened 2020-07-20
Refs #30398 - fix rubocop

ares:


ezr-ondrej:
Don't we want to change the rule instead?
The :rocket: style is making me sad :panda_face: and core is accepting both already.

ares:
@adamruzicka opinions? I don't care that much and I'm all for keeping pandas happy :-)

adamruzicka:
I used to heavily favor rockets, but not so much nowadays. How about sticking with core and allowing both?

ezr-ondrej:
I'd be for allowing both :+1:
What about [no_mixed_keys](https://www.rubydoc.info/gems/rubocop/0.46.0/RuboCop/Cop/Style/HashSyntax)?

ezr-ondrej:
Thank you @ares ! I've opened #356 to change the required Hash style according to the discussion.

PR #355, opened 2020-08-03
Fixes #30545 - Pass whitelist options only if provided by the user

adamruzicka:


ezr-ondrej:
Only the rubocop, that's addressed in #356 is failing.

PR #358, opened 2020-08-10
Fixes #30616 - remove callback

ezr-ondrej:
Favor `foreman-ansible-modules` callback, it can be downloaded even for older installations.

The callback here is only to support ansible <= 2.1
This version should not be supported any longer and this callback is not working on python >= 2.4, so it only confuses users.


adamruzicka (approved):
Let's get rid of it

adamruzicka:
[test foreman_ansible] just to be sure, even though I don't really see how this could break anything

adamruzicka:
The broken test should be fixed by https://github.com/theforeman/foreman_ansible/pull/360, merging.

adamruzicka:
Thank you @ezr-ondrej !

PR #359, opened 2020-08-17
Fixes #30648 - Make sure cloud connector is configured from internal proxy

adamruzicka:
The playbook does some local lookups so running it from another proxy would fail.

Requires
- [x] https://github.com/theforeman/foreman_remote_execution/pull/530

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -5,6 +5,11 @@ if defined? ForemanRemoteExecution
     # Provider for RemoteExecution that allows to run Ansible playbooks.
     # Read the source of other RemoteExecution providers for more.
     class AnsibleProvider < RemoteExecutionProvider
+      FEATURES_THROUGH_INTERNAL_PROXY = %w(
+        ansible_run_capsule_upgrade
+        ansible_configure_cloud_connector
+      ).freeze
+
       class << self
         def ssh_password(host)
           host_setting(host, :remote_execution_ssh_password)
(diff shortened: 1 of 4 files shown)


app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -5,6 +5,11 @@ module ForemanAnsible
     # Provider for RemoteExecution that allows to run Ansible playbooks.
     # Read the source of other RemoteExecution providers for more.
     class AnsibleProvider < RemoteExecutionProvider
+      FEATURES_THROUGH_INTERNAL_PROXY = %w(
+        ansible_run_capsule_upgrade
+        ansible_configure_cloud_connector
```
ares, line 10: Would it make sense to have this as a flag set on feature definition? I don't like hardcoding names here too much. I thought we already have some flags for features.

adamruzicka: Do we want users to have control over this?

adamruzicka:
Now requires https://github.com/theforeman/foreman_remote_execution/pull/530

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -67,7 +67,7 @@ def proxy_action_class
         end
 
         def required_proxy_selector_for(template)
-          if template.remote_execution_features.where(:label => 'ansible_run_capsule_upgrade').any?
+          if template.remote_execution_features.where(:label => FEATURES_THROUGH_INTERNAL_PROXY).any?
```
ezr-ondrej, line 70: This shouldn't be here now, right?

adamruzicka: Removed the entire method since it can use what we have in REX as-is

adamruzicka:
Required change in REX was merged, still needs REX dependency version bump

ezr-ondrej (approved):
:+1:

ezr-ondrej:
Thanks @adamruzicka.

PR #364, opened 2020-10-28
Fixes #31192 - use correct status when report scheduling fails

evgeni:


evgeni:
[test foreman_ansible]

evgeni:
FWIW: this code was copied from core (both the previous state, and now the fixed one: https://github.com/theforeman/foreman/pull/8111)

ares (approved):

ares:
Thanks @evgeni, mergnig!

PR #365, opened 2020-11-04
Fixes #31260 - add host group methods to the safemode

ares:


ofedoren:
Thanks, @ares!

Since codeclimate mentions two the same block with documentation, I'd suggest you to try `prop_group` similar to these: https://github.com/ofedoren/katello/blob/dfc2745080810e967a69ba311125b0a335ba2417/app/models/katello/model.rb#L13

The logic is the same as with `param_group`: define and then include in `apipie :class do ... end` block via `prop_group :name`: https://github.com/ofedoren/katello/blob/6f83ca3f57172e6280b41970e3e28be8a5e4fe4a/app/models/katello/pool.rb#L119

Also, rubocop isn't happy :/

ares:
I don't think the prop group makes sense, the help differs a bit. I made the difference a bit bigger now, so that the meaning of the methos is clearer for the user. The rubocop issue is unrelated, however I've fix that in a separate commit in this PR, it needs to be merged anyway. The issue is `Error: unrecognized cop Rails/ContentTag found in /home/jenkins/workspace/foreman_ansible-pull-request/database/postgresql/label/fast/ruby/2.7/plugin/.rubocop.yml`

Ready for rereview I think

ares:
btw this would be great to get to the release for 2.3 :-)

ofedoren:
ok to test

ofedoren (commented):
LGTM, @ares, but few checks are failing.

BTW, not insisting at all, but `prop_group` has a little bit of sense since the props are almost the same (the name, ret value) from the library perspective and probably rubocop's. The only difference is the descriptions, which can be adjusted semi-automatically.

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -68,9 +68,9 @@ def import_host(*args)
 module Host
   class Managed
     apipie :class do
-      property :all_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns all ansible roles assigned to the host, both its own and inherited'
-      property :ansible_roles, array_of: 'AnsibleRole', desc: 'Returns ansible roles assigned to the host'
-      property :inherited_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns inherited ansible roles assigned to the host'
+      property :all_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns all ansible roles assigned to the host, both its own and inherited from the host group and its parents'
+      property :ansible_roles, array_of: 'AnsibleRole', desc: 'Returns ansible roles directly assigned to the host'
+      property :inherited_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns only ansible roles assigned to the host host group and its parents'
```
ezr-ondrej, line 73:
```suggestion
      property :inherited_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns only ansible roles assigned to the Host\'s Hostgroup and its parents'
```

ezr-ondrej (commented):
We have a lot of unrelated changes here :thinking:

ares:
We have because of rubocop failures, hence 2 commits... the first one adds safe mode definition and related templates dsl documentation and now also contains your suggestion :-)

app/models/concerns/foreman_ansible/hostgroup_extensions.rb
```diff
@@ -36,3 +36,15 @@ def all_ansible_roles
     end
   end
 end
+
+class Hostgroup
+  apipie :class do
+    property :all_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns all ansible roles assigned to the host group, both its own and inherited from parent host groups'
+    property :ansible_roles, array_of: 'AnsibleRole', desc: 'Returns ansible roles directly assigned to the host group'
+    property :inherited_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns only the inherited ansible roles assigned to the host group parents'
```
ezr-ondrej, line 44:
```suggestion
    property :inherited_ansible_roles, array_of: 'AnsibleRole', desc: 'Returns only the inherited ansible roles assigned to Hostgroup\'s parents'
```

ezr-ondrej (commented):
Now we should do the same for Hostgroup and I believe we are good to go

app/models/concerns/foreman_ansible/hostgroup_extensions.rb
ares: I actually checked the form we use in apipie documentation and for consistency, I kept it lowcase and two words form (in both docs). Just added `'s` as suggested.

ares:
Updated, please take another look

ezr-ondrej (approved):
LGTM :+1:

PR #367, opened 2020-11-16
Fixes #30142 - Add support for Ansible Forks

snagoor:


lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -118,6 +120,10 @@ def verbose?
         @verbosity_level.to_i.positive?
       end
 
+      def forks
+        '--forks ' + (@ansible_forks.to_i.positive? ? @ansible_forks.to_s : '5')
```
adamruzicka, line 124: I'd say we should not enforce a default value here. If an override is not provided by the user, let's let ansible decide on its own

snagoor: Agreed @adamruzicka please take another look when you can.

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -118,6 +120,14 @@ def verbose?
         @verbosity_level.to_i.positive?
       end
 
+      def forks_defined?
+        @ansible_forks.to_i.positive?
+      end
+
+      def forks
+        '--forks ' + @ansible_forks.to_s
```
adamruzicka, line 128: Because of how arguments are processed, you either need to insert `--forks` and the number as two separate elements into the `command` array or replace `--forks ` with `--forks=`

adamruzicka (changes requested):

snagoor:
@adamruzicka Apologies for the delay. 

If I don't set the default value for forks, all `dynflow-sidekiq@` services are failing. So, I did set the default value matching with the Ansible forks default value i.e. `5`.

snagoor:
Here is the error from the from `dynflow-sidekiq@orchestrator.service` https://gist.github.com/snagoor/5bc0a9137d9bf6353237c67bd2aa02f8

adamruzicka:
You would need to push the setting name into `Setting::BLANK_ATTRS`, similarly as we do here https://github.com/theforeman/foreman_ansible/blob/master/app/models/setting/ansible.rb#L88

ares:
Should this perhaps use the new provider specific fields? Global setting is nice, but I assume I want to tweak this per invocation (based on how many hosts I run this on)

stejskalleos:
I'm closing the PR, it's been over year since the last update, feel free to reopen new PR if you wish to continue.

PR #368, opened 2020-11-26
Update Github Actions to use bundler-cache

ekohl:
This does the right thing when it comes to caching and reduces the code needed.

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -1,6 +1,7 @@
 name: Ruby Testing
 on: [pull_request]
 env:
+  BUNDLE_WITHOUT: journald:development:console:libvirt
   RAILS_ENV: test
   DATABASE_URL: postgresql://postgres:@localhost/test
   DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
@@ -13,10 +14,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 2.6
-      - name: Setup
-        run: |
-          gem install bundler
-          bundle install --jobs=3 --retry=3
+          bundler-cache: true
       - name: Run rubocop
         run: bundle exec rubocop
   test_ruby:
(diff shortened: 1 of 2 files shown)


ekohl:
Rubocop is not happy.

.github/workflows/ruby_tests.yml
```diff
@@ -13,10 +13,7 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 2.6
-      - name: Setup
-        run: |
-          gem install bundler
-          bundle install --jobs=3 --retry=3
+          bundler-cache: true
```
ezr-ondrej, line 16: `bundler-cache` is installing it in the vendor subfolder, so we would need to exclude it from Rubocop.

I'd be for not enabling it for RuboCop in this PR.
We don't need the caching so prudently here (especially once we don't know if we can have two gem caches for project). Ideally RuboCop job should install just the RuboCop strict dependencies, that should be quite minimal.

ezr-ondrej (changes requested):
Awesome find! I had no idea about this to exist, I'd still like to know if they've solved some issues in the caching or if we still need to use the tricks, but it's definitelly better to use the `setup-ruby` to solve the cache :+1: I'd be for just using it in the main test jobs for now though (regardless the error, that can be fixed by excluding the `vendor` dir IMO)

.github/workflows/ruby_tests.yml
ekohl: Still, ignoring `vendor/bundle` is a good thing. There are other benefits to using `ruby/setup-ruby` instead of `actions/setup-ruby`, like that it's maintained. Having the action also perform the install means they will ensure it's set up correctly. Caching is IMHO just a nice side effect.

ekohl: Also updated. Now `BUNDLER_WITHOUT` is a global variable which means the cache of the rubocop step should also be used in the actual test run.

ekohl: Oh, of course that doesn't work. Different git repos.

ezr-ondrej: Yeah, it is creating the Gemfile.lock prior to fetching the cache, so it needs to match and those two jobs don't match for obvious resons, but it works for both great!

ezr-ondrej (approved):
I'm still not sure about the scope of the cache if that will work but I'm gonna try in followup PR, this is definitelly an improvement :+1: 
Thanks @ekohl for finding it out and making it work!

ezr-ondrej:
This still doesn't create the cache for use on all PRs, we need following to work properly: #369 OR #370

PR #369, opened 2020-11-29
Write cache on push to master

ezr-ondrej:
In order to have the cache accessible from PR source branches, we need to create it on target one.
Cache created from different PR is not accessible to others.
We need to run the tests on master.

~~Alternative to #370~~.
Runs all test suite after merge - maybe an overkill for just creating the cache, but could be usefull for additional confidence in time overlaping merges (previous merge broke following one, but we haven't re-run the tests on the later).
EDIT: using the `event_name`, we can filter out the push triggers.

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -1,5 +1,9 @@
 name: Ruby Testing
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+    - master
 env:
   BUNDLE_WITHOUT: journald:development:console:libvirt
   RAILS_ENV: test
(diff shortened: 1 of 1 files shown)


ekohl:
I think you can also skip stages depending on the event. In https://github.com/theforeman/foreman-installer-modulesync/pull/136 I intend to find out if it does. Crucial line is `if: github.event_name != 'schedule' || github.repository == 'theforeman/puppet-foreman'` to only run the schedule on our version rather than also on all forks. I think you can also exclude running tests on pushes that way. Perhaps that's a nice compromise in maintainability?

ezr-ondrej:
> Perhaps that's a nice compromise in maintainability?

Sounds a good compromise, let's try :+1:

ekohl (approved):

ezr-ondrej:
@xprazak2 could you take a look, this is needed for the cache to start doing its thing (without it, we don't create cache in this repo, only in the forks)

xprazak2:
Thanks!

PR #370, opened 2020-11-29
Write cache on push to master

ezr-ondrej:
In order to have the cache accessible from PR source branches, we need to create it on target one.
Cache created from different PR is not accessible to others.
We need to run the tests on master.

Alternative to #369 
Harder to maintain, but doesn't run unecessary (is it really unecessary? :)) actions

### DIFF
--- a/.github/workflows/write_cache.yml
+++ b/.github/workflows/write_cache.yml
@@ -1,13 +1,11 @@
-name: Ruby Testing
-on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


ekohl:
I don't think this is too pretty. See https://github.com/theforeman/foreman_ansible/pull/369#issuecomment-735813324 for my proposed solution.

ezr-ondrej:
Closing for #369

PR #372, opened 2020-12-01
Fixes #31419 - Update f-m pkg if already installed

snagoor:
foreman-maintain package needs to be at the latest version when running the Capsule upgrade playbook otherwise the newer versions wouldn't be detected.


ezr-ondrej (approved):
Thanks @snagoor, makes sense! :+1:

PR #373, opened 2020-12-10
Fixes #31488, #23288 - Galaxy install location

ezr-ondrej:
We didn't output from the ruby helper, so the option was never used.
Also the default for ansible-galaxy has changed and this is a quick fix to use our default.
This should be overpassed by local configuration file in working directory we run ansible from.


tbrisker (approved):
Thanks @ezr-ondrej !

PR #375, opened 2021-01-05
Add job template - Ansible Roles with tags support

dejongm:
Create job template for running Ansible roles with --tags and --skip-tags support.


ares:
This is to some degree implemting https://projects.theforeman.org/issues/25459, however we should be able to add the tag selection to the job invocation form I think. Note that this PR basically runs ansible from ansible process itself.

dejongm:
Sorry.. I was half asleep this morning and pasted the incomplete code. The code should now reflect the input fields for Tags and Skip-Tags

ezr-ondrej:
Thanks a lot for the idea @dejongm ! We've reimplemented this and finally merged in #395.
This will be available in the next release.

PR #377, opened 2021-01-06
Fix running js tests

xprazak2:
Core is not needed to run plugin
js tests. It was not even being used
since it was not linked to plugin
and was causing failures.


ezr-ondrej (approved):
Make sense, I've merged the fix for rubocop, but it is irrelevant to this, let's get this in :+1:

PR #378, opened 2021-01-15
Fixes #31651 - show the value of imported values

sbernhard:


sbernhard:
Hmm, looks like it doesn't work. Need to have a look at it again.

sbernhard:
[test foreman_ansible]

xprazak2:
I think the test is failing because variables importer now always imports all new variables with `:hidden_value => false`

sbernhard:
Hmm, unfortunately, it doesn't work right now on a installed foreman and I can not find the method "hidden_value?" It looks like, that hidden_value? is false when the ansible_variable is initialized but then "somehow" hidden_value is still set...

sbernhard:
OK, I found out why hidden_value is always "true" right now. The process is like this:
- read all roles and initialize new variables. At this time, hidden_value is set to false
- UI shows which new/updated/deleted parameters exists
- finish_import is called which will then call e.g.
- 
https://github.com/theforeman/foreman_ansible/blob/cb2ffa3fe5fd311a1d413960e09177b7377e6115/app/services/foreman_ansible/variables_importer.rb#L80

At this stage, "hidden_value" is (as always) something like "****" => not false => true is saved to hidden_value instead of false.
How to fix this?

Change the method "create_new_variables" and always set hidden_value = false
Change the method "update_variables" to let hidden_value untouched

Ideas?

xprazak2:
I'd say `hidden_value` should not be a string if we expect it to be a boolean...

@shiramax I know you are currently looking into this area, so I am mentioning you for visibility.

sbernhard:
hidden_value is a string - and, somehow a boolean which can be used by hidden_value? to get if its true or false. That mechanism is implemented in foreman core (lookup_keys). The issue just came around in foreman_ansible because how it is used:
- initiate the ansible variable (hidden_value = false)
- on the UI show the section on which a user can select if a parameter should be updated/added
- submit on this UI will send the data again and again ansible_variables are initiated based on the previous values - and on this stage hidden_value is "****" => which will result in hidden_value = true.

So, a change in foreman core is necessary OR a workaround need to be used on foreman_ansible.

sbernhard:
hidden_value is a string - and, somehow a boolean which can be used by hidden_value? to get if its true or false. That mechanism is implemented in foreman core (lookup_keys). The issue just came around in foreman_ansible because how it is used:
- initiate the ansible variable (hidden_value = false)
- on the UI show the section on which a user can select if a parameter should be updated/added
- submit on this UI will send the data again and again ansible_variables are initiated based on the previous values - and on this stage hidden_value is "****" => which will result in hidden_value = true.

So, a change in foreman core is necessary OR a workaround need to be used on foreman_ansible.

xprazak2:
I think it is ok to fix it here for now.

sbernhard:
works now. have a look at it @xprazak2 :)

xprazak2:
Thanks!

PR #379, opened 2021-01-19
Fixes #28287 - Import Ansible roles and variables in one step

shiramax:
This PR Includes: 
- The ability to sync roles and their variables from Capsule in one step
- React table to present all the roles and variables that need to be synced.
- Removing the 'import variables' view from the index variables page.


webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,223 @@
...
+  const columns = columnsData.map(col => ({
+    ...col,
+    transforms: [cellWidth(10)],
+  }));
+  const [rows, setRows] = useState(rowsData);
+
+  const [isChecked, setIsChecked] = useState(false);
+  const [selectedRowsCount, setSelectRowsCount] = useState(0);
+
+  const dispatch = useDispatch();
```
xprazak2, line 47: redux wiring could be moved to a wrapper component and actions into a separate file, [a rough sketch](https://github.com/xprazak2/foreman_ansible/commit/d0b08e03f501b1dccae2a6c6fb47956bfa84021e)

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,223 @@
...
+  const [page, setPage] =useState();
+  const [perPage, setPerPage]= useState(DEFAULT_PER_PAGE);
+  const [paginatedRows, setPaginatedRows]= useState(rows.slice(0, perPage));
+
+  debugger;
```
xprazak2, line 52: :slightly_smiling_face:

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,223 @@
...
+        rows={paginatedRows}
+        cells={columns}
+        canSelectAll={false}
+      >
+        <TableHeader />
+        <TableBody />
+      </Table>
+      {renderNotFound()}
+      {renderPagination()}
+      {renderSubmitAndCancel()}
```
xprazak2, line 205: You can turn these into separate components to make this component a bit more slim.

package.json
```diff
@@ -10,17 +10,26 @@
...
+    "react-json-tree": "^0.11.2"
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
+    "@patternfly/react-core": "^4.84.4",
+    "@patternfly/react-table": "^4.19.45",
     "@theforeman/builder": "^6.0.0",
     "@theforeman/eslint-plugin-foreman": "6.0.0",
     "@theforeman/test": "^6.0.0",
+    "@theforeman/vendor": "^8.1.2",
```
xprazak2, line 24: Foreman is currently on ^6.0.0, it should be ok but I am not sure if plugins should be ahead...

xprazak2:
Nice work, I added a couple of comments.

app/views/ansible_roles/import.html.erb
```diff
@@ -1,51 +1,53 @@
...
-          </tr>
+<%= webpacked_plugins_js_for :foreman_ansible %>
+
+<% my_rows = [] %>
+<% changed.each do |kind, roles| %>
+  <% roles.each do |role| %>
+
+    <% action =  { "new" => _("Import Role "), "obsolete" => _("Remove Role"), "old" => _("Update Role Variables")}[kind] %>
+    <% if kind == "new" %>
+      <% variables = {"Add" => @importer.get_variables_names(role)[role.name]} %>
```
xprazak2, line 10: This is a lot of logic inside an erb template. Maybe calling proxy api would be better to have in controller rather than in view?

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,176 @@
...
+  rowsData,
+  proxy,
+  onSubmit,
+  onCancel,
+}) => {
+  const columns = columnsData.map(col => ({
+    ...col,
+    transforms: [cellWidth(10)],
+  }));
+  const [rows, setRows] = useState(rowsData);
```
xprazak2, line 32: Not sure if you are aware that initializing state from props has pitfalls - `useState` does not update the state when prop value changes. No problems are manifested here since props are passed from erb on mount, but they would be apparent when loading data asynchronously.

package.json
```diff
@@ -14,13 +14,19 @@
...
     "@theforeman/test": "^6.0.0",
     "@theforeman/vendor-dev": "^6.0.0",
     "babel-eslint": "^10.0.3",
     "eslint": "^6.7.2",
-    "prettier": "^1.13.5"
+    "prettier": "^1.13.5",
+    "react-redux": "^5.0.7",
+    "react-redux-test-utils": "^0.1.1",
+    "redux": "^3.7.2",
+    "redux-thunk": "^2.3.0"
```
xprazak2, line 29: I do not think these need to be added - the tests should work without them now.

app/controllers/ansible_roles_controller.rb
```diff
@@ -22,18 +22,71 @@ def destroy
...
+    rows = []
+    changed.each do |kind, roles|
+      roles.each do |role|
+        role_action = get_role_action(kind)
+        if kind == 'new'
+          variables = { 'Add' => @importer.get_variables_names(role)[role.name] }
+        elsif kind == 'obsolete'
+          variables = { 'Remove' => role.ansible_variables.map(&:key) }
+        elsif kind == 'old'
+          variables = get_old_roles_variables(roles, role)
```
xprazak2, line 76: The proxy API is called for each existing role with a request for variable names of all existing roles, but all variables that do not belong to `role` role are discarded. It should be enough to make just one call to get everything that is needed, can you think of a way?

app/controllers/ansible_roles_controller.rb
```diff
@@ -22,18 +22,71 @@ def destroy
...
+    variables.each do |action, temp_variables|
+      str += "#{action}: #{temp_variables.size} " unless temp_variables.empty?
+    end
+  end
+
+  def get_role_action(kind)
+    { 'new' => _('Import Role '), 'obsolete' => _('Remove Role'), 'old' => _('Update Role Variables') }[kind]
+  end
+
+  def prepare_rows(changed)
```
xprazak2, line 66: I like how this has been split up into multiple methods, it might be worth to consider moving them into a concern or a separate service.

xprazak2:
Nice progress, the pagination does not quite work for me:

roles-page

nil: it could be nice to add a bit of space between the buttons

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,176 @@
...
+    </Toolbar>
+  );
+
+  const renderSubmitAndCancel = () => (
+    <div>
+      <br />
+      <br />
+      <Button variant="primary" onClick={() => onSubmit(rows, proxy)}>
+        {' '}
+        Submit{' '}
```
xprazak2, line 132: What effect should spaces have?

shiramax:
@xprazak2 I've addressed all your comments, let me know if you have more comments. thanks!

app/controllers/foreman_ansible/concerns/ansible_roles_controller_extensions.rb
```diff
@@ -0,0 +1,62 @@
...
+            variables[get_variable_action(kind)].append(temp_variable.key) if temp_variable.ansible_role_id == role.id
+          end
+        end
+        variables
+      end
+
+      def variables_to_s(variables)
+        str = ''
+        variables.each do |action, temp_variables|
+          str += "#{action}: #{temp_variables.size} " unless temp_variables.empty?
```
xprazak2, line 25: nit: You may need to add a separator when there are multiple actions so the final text could be: `Add: 4, Remove 2` instead of `Add: 4 Remove: 2`

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,178 @@
...
+  const renderPagination = (variant = 'top') => (
+    <Pagination
+      isCompact
+      itemCount={rows.length}
+      page={page}
+      perPage={perPage}
+      defaultToFullPage
+      onSetPage={handleSetPage}
+      onPerPageSelect={handlePerPageSelect}
+      perPageOptions={[
```
xprazak2, line 99: Should we rather use [common values](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/Pagination/Pagination.js#L27-L28) for the whole application? Same goes for `perPage`.

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,178 @@
...
+  );
+
+  const renderSubmitAndCancel = () => (
+    <div>
+      <br />
+      <br />
+      <Button variant="primary" onClick={() => onSubmit(rows, proxy)}>
+        Submit
+      </Button>
+      &nbsp;
```
xprazak2, line 136: What do you think about using [css](https://github.com/theforeman/foreman_ansible/commit/b723d344e568af5afbde9e1b217194814f61cdc9) rather than adding nodes into page to achieve the desired look?

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariablesHelpers.js
```diff
@@ -0,0 +1,9 @@
+export const prepareResult = rows => {
```
xprazak2, line 1: Could you add a test please? [Tests for plain js functions](https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/SettingsTable/__tests__/SettingsTableHelpers.test.js#L49-L70) are quite easy to write and do not require snapshots.

xprazak2:
I found a small problem when importing multiple/all roles, it seems only one of them seems to be actually imported:

1)  Select all roles for import and import
2) Try to import again, roles that were supposed to be imported in previous step are shown

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariablesHelpers.js
shiramax: done

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
shiramax: done

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
shiramax: I think the values I need are a bit different.

app/controllers/foreman_ansible/concerns/ansible_roles_controller_extensions.rb
shiramax: done

shiramax:
@xprazak2 I believe now everything works as expected.

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -37,6 +37,34 @@ def import_variables(role_variables, new_roles)
...
+          values.each do |value|
+            variables_to_import[role][value.key] = value.attributes.to_json
+          end
+        end
+        create_new_variables(variables_to_import)
+      end
+      return if roles['old'].empty?
+      variables_to_update = { 'new' => {}, 'obsolete' => {}, 'update' => {} }
+      import_variable_names([]).each do |kind, variables|
+        variables.each do |variable|
```
xprazak2, line 58: This iterates over each variable in every `kind`, but `new` variables are already created above and `update` variables are never handled, so changes for updated variables are never imported.

webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -0,0 +1,178 @@
...
+          </ToolbarItem>
+        </ToolbarGroup>
+      </ToolbarContent>
+    </Toolbar>
+  );
+
+  const renderSubmitAndCancel = () => (
+    <div className="submit-cancel-btns">
+      <br />
+      <br />
```
xprazak2, line 133: nit: I think you can remove the `<br />` tags, the space between table and buttons is determined by css and you can change padding to get the desired look.

ares:
Nice work! Few notes from user view testing:
1) If I remove a variable from Foreman and rerun the import, I correctly see `Add 1` in Variables column. If I also comment some variables in dafault.yml, I see the action as `Remove 8` the number is weird and in fact tells me the number of current existing variables. The operation at the end is the removal of one (first) variable and nothing is added. I'd expect to see somethiling like `Add 1, Remove 1` and do this in a single operation. Would that be possible? Other option is to list this as two operations and I could pick any subset of that. Also while playing with this, I saw `Add 6` but it only adds one.

It seems that the initial role import work reliably though.

2) When I go to Variables list from the role, I also see Import button there, which leads to the old import. Could that also use the new import? The old combines variables from all roles anyway. I think we could drop the old way entirely.

ares:
One more though (I haven't tested how it behaves), variables can be imported or created manally, the import process should never suggest removing variables created manually. It should suggest removal only of those that were imported. (I'll check that on next time testing)

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -37,6 +37,34 @@ def import_variables(role_variables, new_roles)
...
+          if roles['old'].values.map { |role| role['id'] }.include?(variable.ansible_role_id)
+            role = AnsibleRole.where(:id => variable.ansible_role_id)[0]
+            variables_to_update[kind][role.name] = { variable.key => variable.attributes.to_json }
```
xprazak2, line 61: This assigns variable into hash under `role.name`, but what if there are multiple variables that pass the condition above?

app/services/foreman_ansible/roles_importer.rb
```diff
@@ -26,11 +27,15 @@ def import_roles(roles)
...
     def detect_changes(imported)
       changes = {}.with_indifferent_access
-      old, changes[:new] = imported.partition { |role| role.id.present? }
-      changes[:obsolete] = ::AnsibleRole.where.not(:id => old.map(&:id))
+      changes[:old], changes[:new] = imported.partition { |role| role.id.present? }
+      changes[:obsolete] = ::AnsibleRole.where.not(:id => changes[:old].map(&:id))
       changes
     end
 
+    def get_variables_names(role)
```
ezr-ondrej, line 35: Wouldn't this be better of in the `VariablesImporter`?
And it would probably be better to load variables for all the roles at once, as like this if we are importing higher number of roles, we make proxy API call per role. It is not very likely to be an issue, but there might be such cases.

ezr-ondrej (commented):
I've tried to go through the ruby code and made commets to improve the code, it's mostly ideas, if you feel you're way is better, feel free to disagree with any or all of them :)

app/views/ansible_roles/import.html.erb
```diff
@@ -1,51 +1,18 @@
...
+<%= webpacked_plugins_js_for :foreman_ansible %>
+<%= webpacked_plugins_css_for :foreman_ansible %>
+
+<% cols =  [
+    {title: 'Name'},
+    {title:'Operation'},
+    {title:'Variables'},
+    {title:'Hosts Count'},
+    {title:'Hostgroups count'},
+] %>
```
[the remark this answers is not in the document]
shiramax: just because it feels like this data is only for this table, and maybe later own, I will choose to use the react component in different places (for example the roles index or the variables index)

shiramax:
Thanks all for the feedback, 

@xprazak2 
I still wasn't able to reproduce it, and I've added some tests with variable with :override => false, :hidden_value => true, and it seems to be working fine now, I did change some things so please let me know if you are still able to reproduce this issue. 

@ares 
1. I've tested it multiple times, and I wasn't able to reproduce this issue as well I'm able to see 
Add: 1, Remove: 1 .  I've added some tests to double-check it, let me know if you still experiencing this issue. 
2. Sure, no problem, I can also delete importing variables completely, I even think it will be less confusing. let me know what do you think. (right now I removed it completely)

@ezr-ondrej I've addressed all your comments. 
I've changed the extensions to a helper, but now I have rubocop failures 
`
app/helpers/foreman_ansible/ansible_roles_data_preparations.rb:48:30: C: Rails/HelperInstanceVariable: Do not use instance variables in helpers.
        imported_variables = @variables_importer.import_variable_names(roles)
                             ^^^^^^^^^^^^^^^^^^^
`
I'm not sure what to do here, what do you think?

app/helpers/foreman_ansible/ansible_roles_data_preparations.rb
```diff
@@ -0,0 +1,65 @@
...
+        variables = { 'Add' => @variables_importer.get_variables_names(role.name) }
+      elsif kind == 'obsolete'
+        variables = { 'Remove' => role.ansible_variables.map(&:key) }
+      elsif kind == 'old'
+        variables = get_old_roles_variables(imported_variables, role)
+      end
+      variables_to_s(variables)
+    end
+
+    def prepare_rows(changed)
```
ezr-ondrej, line 45: Please keep in mind, helper methods should have quite specific names because the scope can be very global, so we should avoid generic names that could override one another :)

Also to address the instance variables, we just need to pass the instance variable to the helper method.
```suggestion
    def prepare_ansible_import_rows(changed, variables_importer)
```
and then we just call it as `prepare_ansible_import_rows(changed, @variables_importer)`

app/views/ansible_roles/import.html.erb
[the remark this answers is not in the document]
ezr-ondrej: You sure? The component has Import in its name, sure you can have subcomponents in it, and reuse those, but the main component will hardly be reused for another purpose with such name, right?

You can have specific component, that is just passing these options to a generic component.

app/helpers/foreman_ansible/ansible_roles_data_preparations.rb
shiramax: done.

app/views/ansible_roles/import.html.erb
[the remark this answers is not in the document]
shiramax: ok I agree, done. thanks @ezr-ondrej

xprazak2:
I still see roles/variables repeatedly offered for update. Reproducer:

1. Create a fake role with variable inside `roles_path` so that proxy is able to detect it:

```
mkdir -p roles_path/another.role/defaults
echo "another_role_variable: foo" > roles_path/another.role/defaults/main.yml
```

2. Create a role and variable via console to simulate that records already exist in database, use the same names as what is in `roles_path`:

```ruby
role = AnsibleRole.create(:name => "another.role")
AnsibleVariable.create(
  :key => "another_role_variable",
  :value => "foo",
  :hidden_value => true,
  :imported => true,
  :ansilbe_role_id => role.id
)
```

3. Import will show `another.role` as a role to update, select and confirm. Then try to import again and `another.role` will be shown for update again.

app/services/foreman_ansible/variables_importer.rb
```diff
@@ -37,6 +41,34 @@ def import_variables(role_variables, new_roles)
...
+          end
+        end
+        create_new_variables(variables_to_import)
+      end
+      return if roles['old'].empty?
+      variables_to_update = { 'new' => {}, 'obsolete' => {}, 'update' => {} }
+      import_variable_names([]).each do |kind, variables|
+        variables.each do |variable|
+          next unless roles['old'].values.map { |role| role['id'] }.include?(variable.ansible_role_id)
+          role = AnsibleRole.where(:id => variable.ansible_role_id)[0]
```
xprazak2, line 64: nit: `role = variable.ansible_role` ?

package.json
```diff
@@ -14,6 +14,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
+    "@patternfly/react-core": "^4.84.4",
+    "@patternfly/react-table": "^4.19.45",
```
ezr-ondrej, line 18: Are these packaged?

shiramax: I think so :
https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/package.json#L32 
https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/package.json#L29

app/services/foreman_ansible/variables_importer.rb
shiramax: done.

shiramax:
@xprazak2 thanks now I'm able to reproduce it, it was fixed.

shiramax:
[test foreman_ansible]

package.json
xprazak2: You can remove both lines, those packages will be available via vendor which is already in peerDependencies.

xprazak2:
Works as expected, patternfly packages should not be in devDependencies, but I have no additional comments other than that.

package.json
shiramax: @xprazak2  done. thanks!

xprazak2 (approved):
Does anyone have any additional comments or concerns?

ezr-ondrej:
I didn't tested it, but code looks reasonable, if you want mi to take it for a spin, give me a shout

ezr-ondrej:
Maybe it would deserve some commit description? :shrug:

app/controllers/ansible_roles_controller.rb
```diff
@@ -22,18 +22,19 @@ def destroy
 
   def import
     changed = @importer.import!
-    if changed.values.all?(&:empty?)
+    @rows = prepare_ansible_import_rows(changed, @variables_importer)
+    if @rows.empty?
       success no_changed_roles_message
```
ezr-ondrej, line 27: this message could be tailored into "No added or removed roles nor variables" I'm ok to do so in another PR though :)

shiramax: done

ares (approved):
All of my comments are addressed and all works as expected. I'm not merging since I see red tests, but if that's unrelated, anyone please go ahead and click the green button. Thanks @shiramax this is really a nice improvement! Looking forward to see it on the next community demo :-)

xprazak2:
Thanks everyone!

PR #380, opened 2021-01-22
Fixes #27273 - Specify order for ansible roles

xprazak2:
Packaging for `acts_as_list`:
- [x] rpm: https://github.com/theforeman/foreman-packaging/pull/6312
- [ ] deb: https://github.com/theforeman/foreman-packaging/pull/6331 (that takes over for previous ~~https://github.com/theforeman/foreman-packaging/pull/6313~~)

### DIFF
--- a/app/helpers/foreman_ansible/ansible_roles_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_roles_helper.rb
@@ -35,8 +35,29 @@ module ForemanAnsible
       _('%s ago') % time_ago_in_words(role.updated_at)
     end
 
+    def role_attributes_for_roles_switcher(form_object)
+      if form_object.is_a?(Hostgroup)
+        model_roles_attrs(form_object, HostgroupAnsibleRole, :hostgroup_id, :hostgroup_ansible_role_id)
+      else
+        model_roles_attrs(form_object, HostAnsibleRole, :host_id, :host_ansible_role_id)
+      end
+    end
+
     def roles_attrs(roles)
       roles.map { |item| { :id => item.id, :name => item.name } }
+    end
+
+    def model_roles_attrs(form_object, assoc_class, foreign_key, assoc_key)
+      inherited_attrs = roles_attrs form_object.inherited_ansible_roles_ordered
+
+      own_attrs = assoc_class.includes(:ansible_role).
+                  where(foreign_key => form_object.id).
+                  where.not(:ansible_roles => { :id => form_object.inherited_ansible_roles.pluck(:id) }).
+                  order(:position).
+                  map do |join_record|
+        { :id => join_record.ansible_role_id, assoc_key => join_record.id, :name => join_record.ansible_role.name, :position => join_record.position }
+      end
+      inherited_attrs + own_attrs
     end
   end
 end
--- a/app/models/concerns/foreman_ansible/host_managed_extensions.rb
+++ b/app/models/concerns/foreman_ansible/host_managed_extensions.rb
@@ -9,8 +9,11 @@ module ForemanAnsible
         include ::ForemanAnsible::Concerns::JobInvocationHelper
 
         has_many :host_ansible_roles, :foreign_key => :host_id
-        has_many :ansible_roles, :through => :host_ansible_roles,
-                                 :dependent => :destroy
+        accepts_nested_attributes_for :host_ansible_roles, :allow_destroy => true
+        has_many :ansible_roles,
+                 -> { order('host_ansible_roles.position ASC') },
+                 :through => :host_ansible_roles,
+                 :dependent => :destroy
         scoped_search :relation => :ansible_roles, :on => :name,
                       :complete_value => true, :rename => :ansible_role,
                       :only_explicit => true
@@ -47,7 +50,7 @@ module ForemanAnsible
     end
 
     def all_ansible_roles
-      (ansible_roles + inherited_ansible_roles).uniq
+      (inherited_ansible_roles + ansible_roles).uniq
     end
 
     # Class methods we may need to override or add
--- a/app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
+++ b/app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
@@ -2,10 +2,9 @@
 <%= webpacked_plugins_css_for :foreman_ansible %>
 
 <div class='tab-pane' id='ansible_roles'>
-  <% roles = f.object.is_a?(Hostgroup) ? roles_attrs(f.object.inherited_and_own_ansible_roles) : roles_attrs(f.object.all_ansible_roles) %>
   <% class_name = f.object.is_a?(Hostgroup) ? 'Hostgroup' : 'Host' %>
   <%= react_component('AnsibleRolesSwitcher', {
-                                                :initialAssignedRoles => roles,
+                                                :initialAssignedRoles => role_attributes_for_roles_switcher(f.object),
                                                 :inheritedRoleIds => f.object.inherited_ansible_roles.map(&:id),
                                                 :availableRolesUrl => ui_ansible_roles_path,
                                                 :resourceId => f.object.id,
--- a/db/migrate/20210120150019_add_position_to_ansible_role.rb
+++ b/db/migrate/20210120150019_add_position_to_ansible_role.rb
@@ -1,0 +1,27 @@
+class AddPositionToAnsibleRole < ActiveRecord::Migration[6.0]
+  def change
+    add_column :host_ansible_roles, :position, :integer
+    add_column :hostgroup_ansible_roles, :position, :integer
+
+    update_hostgroup_ansible_roles
+    update_host_ansible_roles
+    change_column_null :host_ansible_roles, :position, false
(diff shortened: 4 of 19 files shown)


xprazak2:
Needs https://github.com/theforeman/foreman/pull/8268 for drag& drop to work properly

test/factories/ansible_roles.rb
```diff
@@ -4,4 +4,16 @@
   factory :ansible_role do
     sequence(:name) { |n| "ansible_role_#{n}" }
   end
+
+  factory :host_ansible_role do
+    sequence(:position)
+    host_id { nil }
+    ansible_role_id { nil }
```
ezr-ondrej, line 11: Why don't we use `host` and `ansible_role`, it would have the benefit of autocreating those associated records if you don't specify them and its more natural to FactoryBot, so should be easier to read and use.

app/controllers/foreman_ansible/concerns/api_common.rb
```diff
@@ -5,13 +5,44 @@ module Concerns
...
-        @ansible_roles = AnsibleRole.authorized(:view_ansible_roles).find(ids)
+        instance = instance_variable_get("@#{controller_name.singularize}")
+
+        attrs = find_ansible_roles_attributes params, join_model, foreign_key, instance
+        instance.update join_attrs => attrs
       rescue ActiveRecord::RecordNotFound => e
         not_found(e.message)
       end
+
+      def find_ansible_roles_attributes(params, join_model, foreign_key, instance)
```
ezr-ondrej, line 27: Wouldn't it be better to have this method on the host(group) as `ansible_role_ids=` I'd expect it to be the case.
As such setter make sense to me and given it takes Array, we can have the possition by idx as you have here, but I believe it could be easier if we don't need to transform it to nested parameters and you wouldn't need to transform params itself, what feels bit wrong to me.

ezr-ondrej (changes requested):
In general it is definitelly the right direction, but the implementation feels bit convoluted to me.

Main question: why don't we use `acts_as_list`? I would strongly suggest to use it, as it keeps the positions correct all the time (eg. if you delete posittion 2, it repositions all the following records.). If we decide not to use it, we would need to take care of it ourselves. Meaning we would need to write all the `after_update` `after_destroy`, `after_create` hooks and that feels like duplicated effort - the gem is nice and works well.

I've left few other notes in the comments.

xprazak2:
I switched to using `acts_as_list`, it does make the code simpler in some areas, unfortunately not everywhere.

db/migrate/20210120150019_add_position_to_ansible_role.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def update_host_ansible_roles
+    hosts = Host.unscoped.where(:id => HostAnsibleRole.distinct.pluck(:host_id).uniq)
+    hosts.each do |host|
+      host.host_ansible_roles.each_with_index do |har, idx|
+        har.position = idx
+        har.save!
+      end
+    end
```
ezr-ondrej, line 19: This is highly inefficient :worried: What about something like: ?
```suggestion
    HostAnsibleRole.all.pluck(:host_id, :id).group_by(&:first) do |host_id, role_ids|
      role_ids.each_with_index do |(host_id, host_role_id), idx|
        HostAnsibleRole.where(id: host_role_id).update_all(position: idx)
      end
    end
```

db/migrate/20210120150019_add_position_to_ansible_role.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def update_hostgroup_ansible_roles
```
ezr-ondrej, line 22: the same inefficiency is here, so we should do the same, although here it's not such a deal breaker as users probably don't have thousands of roles (but you never know).

test/factories/ansible_roles.rb
```diff
@@ -4,4 +4,16 @@
...
   end
+
+  factory :host_ansible_role do
+    sequence(:position)
+    host
+    ansible_role
+  end
+
+  factory :hostgroup_ansible_role do
+    sequence(:position)
```
ezr-ondrej, line 15: If I'm correct, if you omit possition, acts_as_list will assign it automatically as the highest in given scope.
This would assign weirdly high possitions as it is rising with every usage of the factory.
```suggestion
    position { nil }
```

app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
```diff
@@ -2,7 +2,10 @@
 <%= webpacked_plugins_css_for :foreman_ansible %>
 
 <div class='tab-pane' id='ansible_roles'>
-  <% roles = f.object.is_a?(Hostgroup) ? roles_attrs(f.object.inherited_and_own_ansible_roles) : roles_attrs(f.object.all_ansible_roles) %>
+  <% roles = f.object.is_a?(Hostgroup) ?
+     model_roles_attrs(f.object, HostgroupAnsibleRole, :hostgroup_id, :hostgroup_ansible_role_id) :
+     model_roles_attrs(f.object, HostAnsibleRole, :host_id, :host_ansible_role_id) %>
```
ezr-ondrej, line 7: Wouldn't it be more readable to hide the implementation detail to the method?
```suggestion
  <% roles = attributes_for_assigned_ansible_roles(f.object) %>
```

and then in the method we coul have at the beggining something like:
```ruby
 if model.is_a?(Hostgroup)
  assoc_class, foreign_key, assoc_key = HostgroupAnsibleRole, :hostgroup_id, :hostgroup_ansible_role_id
 else # only other option is Host
   assoc_class, foreign_key, assoc_key = HostAnsibleRole, :host_id, :host_ansible_role_id
 end
```

Or we could implement that at the model level, what would make sense to me.

app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -9,6 +9,7 @@ def self.prepended(base)
         include ::ForemanAnsible::Concerns::JobInvocationHelper
 
         has_many :host_ansible_roles, :foreign_key => :host_id
+        accepts_nested_attributes_for :host_ansible_roles, :allow_destroy => true
         has_many :ansible_roles, :through => :host_ansible_roles,
```
ezr-ondrej, line 13: Could we order by default? :thinking: 
```suggestion
        has_many :ansible_roles, -> { order('host_ansible_roles.position ASC') }, :through => :host_ansible_roles,
```

app/helpers/foreman_ansible/ansible_roles_helper.rb
```diff
@@ -38,5 +38,18 @@ def import_time(role)
     def roles_attrs(roles)
       roles.map { |item| { :id => item.id, :name => item.name } }
     end
+
+    def model_roles_attrs(form_object, assoc_class, foreign_key, assoc_key)
+      inherited_attrs = roles_attrs form_object.inherited_ansible_roles_ordered
+
+      own_attrs = assoc_class.includes(:ansible_role).
```
ezr-ondrej, line 45: in Puppet we have similar issue and it's solved by both Host and Hostgroup to respond to `all_ansible_roles_ordered` and then we just iterate over those. Would something like that be possible, or we need the distinction where the roles came from in the result?

ezr-ondrej (changes requested):
Left few comments, some are just my ADHD, but some (mainly the migration inefficiency) would be nice to implement.

ezr-ondrej:
> I switched to using `acts_as_list`, it does make the code simpler in some areas, unfortunately not everywhere.

The main benefit is IMHO the assurance the possition is kept correct after deletes, creates, insterts at possitions atp...

app/helpers/foreman_ansible/ansible_roles_helper.rb
xprazak2: We need to know if the role is inherited or not.

ezr-ondrej:
The only failure is in `test/unit/concerns/hostgroup_extensions_test.rb` where we expect the old order :)

app/views/foreman_ansible/ansible_roles/_select_tab_content.html.erb
```diff
@@ -2,7 +2,7 @@
 <%= webpacked_plugins_css_for :foreman_ansible %>
 
 <div class='tab-pane' id='ansible_roles'>
-  <% roles = f.object.is_a?(Hostgroup) ? roles_attrs(f.object.inherited_and_own_ansible_roles) : roles_attrs(f.object.all_ansible_roles) %>
+  <% roles = role_attributes_for_roles_switcher(f.object) %>
```
ezr-ondrej, line 5: given we use it just once, it probably make sense to use it directly, instead of assigning the variable and using the variable. WDYT?

ezr-ondrej (commented):
Packaging is in for RPM and IMO we don't need DEB.
One nit, and the test and I believe we are good to go :+1:

ezr-ondrej (approved):
Thanks for the hard work @xprazak2 !
Looks great, we need to wait what happens with the DEB packaging though, but seems we need to get it in :(

ezr-ondrej:
The packaging should get resolved once we bump the plugin version.

PR #381, opened 2021-01-23
Fix #31709 - prefer lsb facts for OS

ares:
LSB facts contain more specific information about OS, e.g. Raspbian vs
Debian. They shoud have higher priority for the OS name.


ares:
it seems rubocop wasn't happy but is unrelated

xprazak2:
Thanks!

PR #382, opened 2021-01-27
Fixes #31672 - define ansible roles to ignore

shiramax:


app/controllers/foreman_ansible/concerns/ansible_roles_controller_extensions.rb
```diff
@@ -38,11 +38,22 @@ def get_roles_variables(imported_variables, kind, role)
         variables_to_s(variables)
       end
 
+      def excluded_roles
+        @excluded_roles ||= Setting.convert_array_to_regexp(
+          Setting[:ansible_roles_to_ignore],
+          {
+            :prefix => '(\A|.*::|.*_)',
```
xprazak2, line 45: Why `::` and `_`? I can't figure that out...

xprazak2:
Works, #382 should go in first. It would be nice to have a test we can correctly exclude roles.

app/controllers/foreman_ansible/concerns/ansible_roles_controller_extensions.rb
shiramax: that's not necessary. I removed it, thanks.

app/models/setting/ansible.rb
```diff
@@ -80,6 +81,12 @@ def default_settings
                'with Ansible inventory'),
             'Ansible - Ansible Inventory',
             N_('Default Ansible inventory report template')
+          ),
+          set(
+            'ansible_roles_to_ignore',
+            N_('Those Roles will be excluded when importing roles from smart proxy, you can use you can use * wildcard metacharacters'),
```
ares, line 87: It should probably mention how to specify multiple values. E.g. comma separated values wrapped in quotes or similar. An example wouldn't hurt :-) feel free to use `theforeman.foreman_scap_client` and `theforeman.foreman.*` for that example

xprazak2:
Needs a rebase now.

shiramax:
@xprazak2 done.

ares (changes requested):
It seems to work fine except for one edge case. When I set some roles to be ignored, they are correctly ignored, the wildcards also work fine. But when I reset this to empty value again, all roles seems to be ignored. It seems like `[]` is considered as all. Setting it back to something that doesn't exist (e.g. "a") helps.

The problem is that `Setting.convert_array_to_regexp(Setting[:ansible_roles_to_ignore])` responds with //, which matches everything. The `excluded_roles` should probably have a check for this case. Or we may want to modify the convert_array_to_regexp behavior in core.

shiramax:
@ares thanks for the review, I believe I fixed it now.

ares (approved):
Nice, it seems tests are now failing due to rubocop issue in REX, but I think they were green before. It may not be a bad idea to add a test for empty string and nil being a setting, but leaving that for another PR. Merging this now, thanks @shiramax!

PR #383, opened 2021-01-28
Fixes #31710 - Add check mode option for running playbook

yifatmakias:
Add the check mode option to the foreman when running a playbook.

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -13,6 +13,7 @@ module ForemanAnsibleCore
         @root = working_dir
         @verbosity_level = action_input[:verbosity_level]
         @rex_command = action_input[:remote_execution_command]
+        @check_mode = action_input[:check_mode]
       end
 
       def start
@@ -109,6 +110,7 @@ module ForemanAnsibleCore
         env = {}
         env['FOREMAN_CALLBACK_DISABLE'] = '1' if @rex_command
         command = [env, 'ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        command << '--cmdline' << '"--check"' if check_mode?
         command << verbosity if verbose?
         initialize_command(*command)
         logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
@@ -120,6 +122,10 @@ module ForemanAnsibleCore
 
       def verbose?
         @verbosity_level.to_i.positive?
+      end
+
+      def check_mode?
+        @check_mode == true
       end
 
       def prepare_directory_structure
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -27,7 +27,8 @@ if defined? ForemanRemoteExecution
             :remote_execution_command => ansible_command?(
               template_invocation.template
             ),
-            :name => host.name
+            :name => host.name,
+            :check_mode => host.host_param('ansible_roles_check_mode')
           )
         end
 


lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -104,7 +105,8 @@ def write_playbook
       end
 
       def start_ansible_runner
-        command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        logger.debug(@check_mode)
```
xprazak2, line 108: Debugging leftovers?

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -11,6 +11,7 @@ def initialize(input, suspended_action:)
         @playbook = input.values.first[:input][:action_input][:script]
         @root = working_dir
         @verbosity_level = input.values.first[:input][:action_input][:verbosity_level]
+        @check_mode = input.values.first[:input][:action_input][:ansible_inventory][:all][:vars][:check_mode]
```
xprazak2, line 14: Just out of curiosity, why is check mode nested so deep when compared with the playbook script?

xprazak2:
Works nicely, just a couple of minor things. Would it be possible to have some sort of humanized label in the form?

form_check_mode

lib/foreman_ansible_core/runner/ansible_runner.rb
yifatmakias: Because the script is a general action_input that is defined in the remote_execution project: specific in foreman_remote_execution/app/lib/actions/remote_execution/run_host_job.rb

yifatmakias:
@xprazak2 What do you mean by humanized label? I improved the description is it better now?

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -104,7 +105,7 @@ def write_playbook
       end
 
       def start_ansible_runner
-        command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        command = check_mode? ? ['ansible-runner', '--cmdline', '"--check"', 'run', @root, '-p', 'playbook.yml'] : ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
```
adamruzicka, line 108: Could we keep the original command and if check mode is enabled, just append `--cmdline` and `"--check"` to it instead of having arrays which are mostly the same?

xprazak2:
> @xprazak2 What do you mean by humanized label? I improved the description is it better now?

I meant having `Check mode` instead of `check_mode` as a label for the field. Is that possible?

lib/foreman_ansible_core/runner/ansible_runner.rb
yifatmakias: Yes, I agree I will fix this.

yifatmakias:
> > What do you mean by humanized label? I improved the description is it better now?
> 
> I meant having `Check mode` instead of `check_mode` as a label for the field. Is that possible?

I think not, it takes the name from the template input name. you can see it also in the Ansible - Run insights maintaince plan:
image

xprazak2 (approved):

app/views/foreman_ansible/job_templates/run_playbook-ansible_default.erb
```diff
@@ -12,6 +12,14 @@ template_inputs:
   description: The playbook to run against given hosts
   advanced: false
   value_type: plain
+- name: check_mode
```
ares, line 15: If the name is "Check mode", we can still use it in the template by `input('Check mode')` but it looks nicer in the form

app/views/foreman_ansible/job_templates/run_playbook-ansible_default.erb
```diff
@@ -12,6 +12,14 @@ template_inputs:
   description: The playbook to run against given hosts
```
ares, line 12: I don't think this is the right job to implement this for. This one is to run arbitrary playbook. While it may be handy there too, I think we need it mainly for https://github.com/theforeman/foreman_ansible/blob/master/app/views/foreman_ansible/job_templates/ansible_roles_-_ansible_default.erb

That job template is used to run all assigned roles of the host/host group.

app/views/foreman_ansible/job_templates/run_playbook-ansible_default.erb
```diff
@@ -12,6 +12,14 @@ template_inputs:
...
   advanced: false
   value_type: plain
+- name: check_mode
+  required: true
+  input_type: user
```
ares, line 17: If this is a user input, user must set/override that per run. I think we should instead make it a parameter type, so that it can be customized through parameters (e.g. on host level, host group level or set a global value). That way I can easily say all hosts should be running in the check mode until I explicitly allow the run for real perhaps on a specific host only.

We should also seed such parameter as a global parameter, the name could be `ansible_roles_check_mode` and the parameter should be of a boolean type.

ares (changes requested):
Few comments, please don't hesitate to ask for more details if something is unclear.

app/views/foreman_ansible/job_templates/run_playbook-ansible_default.erb
yifatmakias: I have a few questions regarding this:
1. I managed to find only user type inputs, is there an input type parameter, should it be added or should it be implemented in another way?
2. Do you think we should also let the user set the check mode for every run or only as a host/host_group parameter? and if so what will be prioritized first when running the host roles for example?

ezr-ondrej: host parameter doesn't need to be specified in the template metadata, you get it by `host_param` from the template.
https://github.com/theforeman/foreman/blob/ccbc96d8f2aa962a72b9b1ede33a657504723c2c/app/models/concerns/host_params.rb#L52

If we seed it as global parameter, we may even rely on it being present.

Being able to override this on a template level may be useful in some situations as well.
In my opininion if we choose to have such template input, it should not be required, but if filled, it should take precedence.

yifatmakias: @ezr-ondrej So just to make sure I understood, we want the templates to have a user input that is not required and also we want to have a global parameter to a host or host group. If the user filled the option in the job it should get precedence and if not it will get the check mode value from the global parameter. Is this correct?

ezr-ondrej: Yes this is correct :+1:
The template option is not even mandatory, so if you'd struggle with the parameter/option precedence, we can end up just with the parameter and it would be ok. But having both is the best result :+1: 

*Just note (to be sure we are talking about the same parameters):* it is always host parameter, it just can get inherited from Hostgroup, but also OS, org/loc, or just from global values.

yifatmakias: Ok thanks for the clarification, I will fix it.

ares:
Do I understand the intetion correctly that we respect the parameter now and in two jobs, we expose the user input to override this? So e.g. if I set the check mode to be enabled, I can still say "No" to a specific invocation? Thanks for clarification.

yifatmakias:
> Do I understand the intetion correctly that we respect the parameter now and in two jobs, we expose the user input to override this? So e.g. if I set the check mode to be enabled, I can still say "No" to a specific invocation? Thanks for clarification.

@ares Yes exactly.

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -105,6 +107,7 @@ def write_playbook
 
       def start_ansible_runner
         command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        command << '--cmdline' << '"--check"' if check_mode?
```
ezr-ondrej, line 110: What is the `--cmdline` option for? :thinking:

yifatmakias: --cmdline CMDLINE     command line options to pass to ansible-playbook at execution time (default=None)
This is the description from the help.
If we were using the ansible command or ansible-playbook command we could have used --check without --cmdline before but because we are using the ansible-runner this is the way to add the check mode option.

xprazak2:
Current status:

* we can add `ansible_roles_check_mode` param to host to run in check mode
* there is input for 2 job templates (ansible roles - anisble default, run playbook - ansible default) that allows setting check mode per job execution

Any additional changes that should be included?

I am getting error on proxy when selecting hosts on index page and choosing 'Run all ansible roles' from dropdown:

run-hosts
proxy-error


A curious thing is that check mode does not get applied to tasks in roles, only to those directly in playbook - but it is possible I am reading this wrong or my env has a bad case of Monday:

tasks

yifatmakias:
@xprazak2 Thanks for the review!
Regarding the error this is weird because it does not happen when I try to run it this way.
Regarding the other thing you mention I think I do not fully understand what do you mean by check mode does not get applied to tasks in roles, only to those directly in playbook? And how can it be an environment issue?

ezr-ondrej (commented):
Do we need the template inputs, when we have the provider input already?
IMHO it doesn't make sense, it would only confuse users as they would get two inputs for check mode on those templates. I'd just remove them.

yifatmakias:
> Do we need the template inputs, when we have the provider input already?
> IMHO it doesn't make sense, it would only confuse users as they would get two inputs for check mode on those templates. I'd just remove them.

@ezr-ondrej What do you mean by provider input? I can remove them of course but if I understand what you mean I think the goal was to allow users to set the check mode option in a specific job and not only as a host param.

ezr-ondrej:
> @ezr-ondrej What do you mean by provider input? I can remove them of course but if I understand what you mean I think the goal was to allow users to set the check mode option in a specific job and not only as a host param.

Main goal was to allow that globally and then for specific Host(group), the template is nice addition. I need to test it, but I thought provider input adds this to a provider form?

xprazak2:
I looked a bit deeper into the error on proxy. Below is the inventory I am getting when running roles from job invocation form vs hosts/index:

missing-vars

Though I am not sure why `check mode` is completely missing in vars.

Regarding the check mode and tasks in roles, I am getting the following output from running roles on host:

tasks


If you take a look at the output of the first task (Do stuff), it says it is being skipped because we run in check mode, which is expected. I added this task directly into playbook by modifying a template:

```yaml
---
- hosts: all
  pre_tasks:
    - name: Display all parameters known for the Foreman host
      debug:
        var: foreman
  tasks:
    - name: Do stuff
      command: 'touch /tmp/foo.txt'
    - name: Apply roles
      include_role:
        name: "{{ role }}"
      loop: "{{ foreman_ansible_roles }}"
      loop_control:
        loop_var: role
```

However if you take a look at the output of the last task (adriagalin.motd: Add motd), it says "changed" and the play recap below reports one change. This task is part of the adriagalin.motd role, which was assigned to the host and therefore applied but since we are running in check mode, it surprised me that this task is not skipped as the previous one.

So I was merely curious if it is same for you.

yifatmakias:
> I looked a bit deeper into the error on proxy. Below is the inventory I am getting when running roles from job invocation form vs hosts/index:
> 
> missing-vars
> 
> Though I am not sure why `check mode` is completely missing in vars.
> 
> Regarding the check mode and tasks in roles, I am getting the following output from running roles on host:
> 
> tasks
> 
> If you take a look at the output of the first task (Do stuff), it says it is being skipped because we run in check mode, which is expected. I added this task directly into playbook by modifying a template:
> 
> ```yaml
> ---
> - hosts: all
>   pre_tasks:
>     - name: Display all parameters known for the Foreman host
>       debug:
>         var: foreman
>   tasks:
>     - name: Do stuff
>       command: 'touch /tmp/foo.txt'
>     - name: Apply roles
>       include_role:
>         name: "{{ role }}"
>       loop: "{{ foreman_ansible_roles }}"
>       loop_control:
>         loop_var: role
> ```
> 
> However if you take a look at the output of the last task (adriagalin.motd: Add motd), it says "changed" and the play recap below reports one change. This task is part of the adriagalin.motd role, which was assigned to the host and therefore applied but since we are running in check mode, it surprised me that this task is not skipped as the previous one.
> 
> So I was merely curious if it is same for you.

@xprazak2 This is definitely weird because it doesn't happen for me.
When I tried to run a playbook with 2 tasks and check mode applied they were both skipped and when I tried to run them without the check mode option only then the tasks were marked as changed.

yifatmakias:
> > @ezr-ondrej What do you mean by provider input? I can remove them of course but if I understand what you mean I think the goal was to allow users to set the check mode option in a specific job and not only as a host param.
> 
> Main goal was to allow that globally and then for specific Host(group), the template is nice addition. I need to test it, but I thought provider input adds this to a provider form?

I am sorry but I still do not understand what do you mean by provider form?
Could you show me in which form there's an addition that should not be?

As far as I know the changes I made added this to the job creation form according to the relevant template and it could be added as a host/host group parameter called ansible_check_mode

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -11,6 +11,8 @@ def initialize(input, suspended_action:)
         @playbook = input.values.first[:input][:action_input][:script]
         @root = working_dir
         @verbosity_level = input.values.first[:input][:action_input][:verbosity_level]
+        @check_mode = input.values.first[:input][:action_input][:ansible_inventory][:all][:vars]['check mode']
+        @check_mode = input.values.first[:input][:action_input][:check_mode] if @check_mode.empty?
```
ezr-ondrej, line 15: If it's not run from form, but the run roles button, this will fail with `<NoMethodError>: undefined method `empty?' for nil:NilClass`
```suggestion
        @check_mode = input.values.first[:input][:action_input][:check_mode] if @check_mode.blank?
```

ezr-ondrej (commented):
I thought provided_inputs add the input on the form... but it does not, so my comment was irrelevant.
I was thinking we can force it onto every Ansible form, but there doesn't seem to be an easy way.

lib/foreman_ansible_core/runner/ansible_runner.rb
yifatmakias: @ezr-ondrej This is very weird, as I know blank? should work on any object but when I changed it to blank? I received this error: <NoMethodError> undefined method `blank?' for "no":String.
Do you maybe have an idea why this is happening?

yifatmakias: @ezr-ondrej After another look it seems that blank? method works on every object in rails project only.
I used empty? and nil? instead. Let me know if it looks ok for you.

ezr-ondrej: huh, right, I didn't realize we are out of rails (the code mixture in one repo is very confusing...)
`@check_mode.nil? || @check_mode.empty?` should be fine I guess unless the input can be a real boolean though

ezr-ondrej:
Actually seeing how much hustle having the template inputs is, wouldn't it be easier to just keep the parameter and be done with it? The input would be very useful on a new template, that user add, but we can't really do that, so I don't think adding them is much useful.

yifatmakias:
> Actually seeing how much hustle having the template inputs is, wouldn't it be easier to just keep the parameter and be done with it? The input would be very useful on a new template, that user add, but we can't really do that, so I don't think adding them is much useful.

But if we keep only the host parameter and remove the template input it will be impossible to run a specific job with the check mode apply. Don't you think this is a useful action for users?

ezr-ondrej (approved):
Works as expected thanks @yifatmakias !

Although I have two ides for followup Improvements:
- Adding information to the job output that this has been run in check mode.
- Documenting this to users in some way.

PR #384, opened 2021-02-03
Fixes #27273 - Make Ansible Roles Switcher draggable

xprazak2:
Adds ability to reorder roles in AnsibleRolesSwitcher.

Needs https://github.com/theforeman/foreman_ansible/pull/380. Without it, no changes are persisted, not even adding/removing roles as posted data are filtered out by strong params.


webpack/components/AnsibleRolesSwitcher/components/AssignedRolesList.js
```diff
@@ -31,40 +77,25 @@ const AssignedRolesList = ({
             dropdownButtonId="assigned-ansible-roles-pagination-row-dropdown"
           />
         </div>
-        {assignedRoles.map(role => (
-          <AnsibleRole
+        {assignedRoles.map(renderRole)}
```
ezr-ondrej, line 80: shouldn't we order the inherited roles first? :thinking:
They will come first, so we should make sure we display it that way, right?
What about sorting the roles beforehand like (expects we have the position of the roles):
```js
assignedRoles.sort((r1, r2) => { if (r1.inherited && r2.inherited) return 0; if (r1.inherited) return -1; if (r2.inherited) return 1; return r1.position - r2.position; };
```

xprazak2: We do not need to order here, roles already have correct order.

ezr-ondrej (changes requested):
I've not undertake enough tests and now we need this: https://github.com/theforeman/foreman_ansible/pull/386
With that it works reasonably though.
I'm just missing the drag hanler, that would give away that roles are draggable to users: https://www.patternfly.org/v4/components/data-list/html/#draggable

xprazak2:
> I'm just missing the drag hanler

I miss it too, seems like that was never finished in [pf3](https://www.patternfly.org/v3/pattern-library/forms-and-controls/drag-and-drop/index.html#) - at least all I can see is overview and design sections.

ezr-ondrej:
> I miss it too, seems like that was never finished in [pf3](https://www.patternfly.org/v3/pattern-library/forms-and-controls/drag-and-drop/index.html#) - at least all I can see is overview and design sections.

Right, we will need to use PF4 at some point anyway, right? so I guess we can wait with the handler till then :thinking:

xprazak2:
Any additional concerns?

ezr-ondrej:
Can we merge https://github.com/theforeman/foreman_ansible/pull/386 first? it takes care of issue brough by https://github.com/theforeman/foreman_ansible/pull/380 that you're partially fixing here, but not fully, without #386 this wont work for new host (when you submit new host form with errors, you'd need to reassign the roles again what can be quite annoying).

xprazak2:
#386 has been merged and I rebased on top of it.

ezr-ondrej:
Something weird is going on there:
output

xprazak2:
Good catch, fixed.

webpack/components/AnsibleRolesSwitcher/components/AnsibleRoleInputs.js
```diff
@@ -0,0 +1,43 @@
...
+  );
+  return (
+    <React.Fragment>
+      {role[attrName(resourceName)] ? idField : null}
+      <input
+        name={fieldName('ansible_role_id')}
+        value={role.id}
+        type="hidden"
+      />
+      <input name={fieldName('position')} value={idx} type="hidden" />
```
ezr-ondrej, line 27: Position is indexed from 1, without this, first save saves wrong order as 0 is put at the end, because the library assigns first free position to it.
```suggestion
      <input name={fieldName('position')} value={idx+1} type="hidden" />
```
Or we could directly send position to this Component and assign it as `idx + 1` in the parent :thinking:

ezr-ondrej (changes requested):
- The possitioning almost works, but there is quirk with idx vs position.
- The removal was not persisted for me.

*Note* While playing with it, I found the pagination is not 100% working, but I've retested on master and it behaves the same, so it is not caused by these changes.

xprazak2:
Positioning and role removal should be fixed now.

ares:
When testing this, I'm getting the traceback on the host edit page

```
2021-03-09T17:53:32 [I|app|8674c109] Completed 500 Internal Server Error in 603ms (ActiveRecord: 43.2ms | Allocations: 438829)
 8674c109 | 
 8674c109 | 
2021-03-09T17:53:32 [F|app|8674c109]   
 8674c109 | ActionView::Template::Error (PG::UndefinedColumn: ERROR:  column host_ansible_roles.position does not exist
 8674c109 | LINE 1: ...HERE "host_ansible_roles"."host_id" = $1 ORDER BY host_ansib...
 8674c109 |                                                              ^
 8674c109 | ):
 8674c109 |     94: <div class="tab-pane" id="ansible_roles">
 8674c109 |     95:   <% class_name = f.object.is_a?(Hostgroup) ? 'Hostgroup' : 'Host' %>
 8674c109 |     96:   <%= react_component('AnsibleRolesSwitcher', {
 8674c109 |     97:                                                 :initialAssignedRoles => role_attributes_for_roles_switcher(f.object),
 8674c109 |     98:                                                 :inheritedRoleIds => f.object.inherited_ansible_roles.map(&:id),
 8674c109 |     99:                                                 :availableRolesUrl => ui_ansible_roles_path,
 8674c109 |    100:                                                 :resourceId => f.object.id,
 8674c109 |   
 8674c109 | /home/ares/z/foreman_ansible/app/helpers/foreman_ansible/ansible_roles_helper.rb:47:in `reject'
 8674c109 | /home/ares/z/foreman_ansible/app/helpers/foreman_ansible/ansible_roles_helper.rb:47:in `role_attributes_for_roles_switcher'
```

but it can totally be me doing something wrong :-)

ares:
Sweet, this resolves the long term issue for small amount of roles already.

The problems I encountered are related to pagination. When I have 15 roles assigned, the pagination seems to be ignored and I always see all 15 roles. I know we faced some issues with pagination in this component in the past and am wondering, whether we could save us some troubles if we drop pagination for now. If we don't, how will one assign a role to the middle of the list when they have 3 pages of roles?

From visualization perspective, we should probably somehow tell the user that drag and drop is supported here. There's a hand cursor, but it was there before, so people know they can click on the role to move it to the other list. Would it be possible to add an order column before the role name? Or pehaps just changed the "Assigned Ansible Roles" header to "Ordered Assigned Ansible Roles"? Extra points for a question mark icon with tooltip help explaining the behavior with also host group inheritance :-)

I don't think all needs to be addressed in this PR, but should be probably somehow addressed soon.

ezr-ondrej:
The drag handler was [discussed previously](https://github.com/theforeman/foreman_ansible/pull/384#issuecomment-784033766) and I think the component should get a redesign to PF4 and the issues should get addressed there, maybe the renaming could be done, but it sounds weird to me to have it in the title.

The pagination for assigned roles is basically ignored as it always was, so you easily order them in the one page. Should we get rid of the pagination for the assigned roles? probably, but I don't believe it should be done in this PR.

ares:
Just to be clear, I'm not blocking this PR at all, I'm totally fine with addressing these as separate PRs :-)

ezr-ondrej:
Merging this, but we need to address some issues before releasing this.

ezr-ondrej:
Thanks @xprazak2, great improvement ! And thanks @ares for the reviews :)

PR #385, opened 2021-02-16
Fixes #31898 - Add a job for installing ansible collections

shiramax:


xprazak2:
Works for me.

A strange thing was that I got 'No such file or directory' when I targeted a machined without Ansible (and therefore ansible-galaxy command was not available). I would expect a more meaningful message, but that is probably out of scope here.

no-dir

app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb
```diff
@@ -0,0 +1,32 @@
...
+template_inputs:
+- name: ansible_collections_list
+  required: true
+  input_type: user
+  description: "List of collections in Ansible Galaxy to install, separated by commas, e.g:
+    \"mysql, nginx\"\r\n\r\nThe default location is the 'collections_paths' configured on
+    /etc/ansible/ansible.cfg, you may override it by filling the 'location' input.
+    Click on \"Advanced\" to see it."
+  advanced: false
+- name: location
```
ofedoren, line 15: Since we've got loc/org options/things mostly in context of taxonomies, I'd suggest to rename the input to `target` or `dir`, or something similar.

shiramax: done

shiramax:
@xprazak2 verifying if the ansible package exists with gathering the facts wouldn't work if the operating system of the client is not RPM-based OS. 
I think we should keep it as is ATM, and investigate later what is the best approach here.

xprazak2:
LGTM

app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb
```diff
@@ -0,0 +1,32 @@
...
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: ansible-galaxy collection install {{ item }} -p <%= input('collections_path').present? ? input('collections_path') : '/etc/ansible/collections' %>
+      register: out
+      with_items:
+        - <%= input('ansible_collections_list').split(',').join(' ') %>
```
ezr-ondrej, line 31: this doesn't do what you intend it to do.
It actually runs with list including only one item that is a string "<col1> <col2> <col3>" so the loop will only run once.
It works though, because ansible `galaxy collection install <col1> <col2> <col3>` works, but we should eighter run that directly or generate the list properly :)

ares: good catch, I guess this should instead be something like
```erb
      with_items:
<%= input('ansible_collections_list').split(',').map { |col| "        - #{col}"}.join("\n") %>
```

or any other implementation :-)

ezr-ondrej: I think we should be fine with just

``yml
- command: ansible-galaxy collection install <%= input('ansible_collections_list').split(',').join(' ') %> -p ...
```

without the loop :) unless we want to split the outputs to work with them separatelly, but I don't see reason for that

Comment (shiramax) [reply app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb:31]: thanks @ezr-ondrej !
I tried that before, and got this:
Invalid collection name 'mrlesmithjr.system  dsglaser.cis_security', name must be in the format <namespace>.<collection>. Please make sure namespace and collection name contains characters from [a-zA-Z0-9_] only.", "stderr_lines": ["[WARNING]:

which is kinda strange, because it did work before, but maybe doing the loop is better.

Comment (ezr-ondrej) [reply app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb:31]: didn't you wrap it in quotes, so it was treated as one argument?
Do what works for you better, here it doesn't matter much, but loop is usually slower :)

Comment (shiramax) [reply app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb]: you are right, I removed the quotes. thanks.

**Comment by ezr-ondrej** [comment]
We are still using the loop, is it intentional?

**Comment by shiramax** [comment]
@ezr-ondrej yes, now the looping is working

**Comment by ezr-ondrej** [comment]
> @ezr-ondrej yes, now the looping is working

Hmm the result template looks like what makes it work because with items will be [['', '']], but ansible flattens it.
```yml
---
- hosts: all
  tasks:
    - command: ansible-galaxy collection install {{ item }} -p /etc/ansible/collections
      register: out
      with_items:
        - ["crivetimihai.development", "extreme.exos"]
    - debug: var=out
```

But it feels to me uneccessary following works just fine
```yml
---
- hosts: all
  tasks:
    - command: "ansible-galaxy collection install <%= input('ansible_collections_list').split(',').join(' ') %> -p <%= input('collections_path').present? ? input('collections_path') : '/etc/ansible/collections' %>"
      register: out
    - debug: var=out
```

app/views/foreman_ansible/job_templates/ansible_collections_-_install_from_galaxy.erb
```diff
@@ -0,0 +1,32 @@
...
+%>
+
+---
+- hosts: all
+  tasks:
+    - command: ansible-galaxy collection install {{ item }} -p <%= input('collections_path').present? ? input('collections_path') : '/etc/ansible/collections' %>
+      register: out
+      with_items:
+        - <%= input('ansible_collections_list').split(",") %>
+    - debug: var=out
```
ezr-ondrej, line 32: Do we want this? Or is it just develop debug leftover?

ezr-ondrej (commented):
I still believe the loop is not necessary, but I dont mind.
The debug seems uneccessary though. If people want to dig into it, they can rise the verbosity level, but failed / changed is fine for most people IMHO

ezr-ondrej (approved):
Much simplier and straight forward IMHO :)
BTW almost looks like we could have used rex here, but I get the value of forcing the ansible setup prior installing the collections :))

ezr-ondrej (approved):
Thank you @shiramax !

PR #386, opened 2021-02-22
Refs #27273 - improve passing assigned roles

ezr-ondrej:
Gets the assigned roles from the association itself as that preserve unsaved assignment changes.
Also this is much easier to get and should be most definitely leveraged.

### DIFF
--- a/app/helpers/foreman_ansible/ansible_roles_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_roles_helper.rb
@@ -36,28 +36,26 @@ module ForemanAnsible
     end
 
     def role_attributes_for_roles_switcher(form_object)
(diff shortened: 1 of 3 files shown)


xprazak2:
Makes the inherited roles show up multiple times:

multiple-roles

ezr-ondrej:
> Makes the inherited roles show up multiple times:

Right, it's taken care of now :+1:

xprazak2:
Roles no longer show multiple times, tests need a slight change.

ezr-ondrej:
> Roles no longer show multiple times, tests need a slight change.

Actually it was wrong, the tests are there to test exactly that :) ... This version works and I added tests for deduplication of the roles :)

ezr-ondrej:
@xprazak2 we are :green_apple: :)

xprazak2:
Thanks!

PR #387, opened 2021-02-24
Fixes #31952 - Run import roles and variables in the background

shiramax:


xprazak2:
Imports roles and variables as expected. 

I do not get a toast after submission - it should probably say something about async import task being started and that changes will take effect once the task finishes. Or would it make sense to redirect to the task detail?

@adamruzicka, @ezr-ondrej, is `ApplicationJob` the right tool for this?

Needs a fix for rubocop.

shiramax:
@xprazak2 I have added a toast, but if I'm redirecting to the index page, I wouldn't able to read the toast, so I removed the dispatching, 
I can also direct to the task, but not sure it's needed.

ezr-ondrej:
I think the job should have notification after it's done (success or fail) :thinking:

app/jobs/sync_roles_and_variables.rb
```diff
@@ -0,0 +1,8 @@
+class SyncRolesAndVariables < ::ApplicationJob
```
ezr-ondrej, line 1: We should provide `humanized_name` so it appears in the tasks as human understandable task.
But :+1: for `ApplicationJob`

xprazak2:
The toast shows now, but staying on the import page introduces more problems than it solves. So I would redirect as you originally had and add the toast, it will pop up just for a second before redirect to let user know what is happening. I think that is the best we can do right now with our hybrid UI.

```js
handleSuccess: () => {
    dispatch(push(ANSIBLE_ROLES_INDEX));
},
successToast: response => (
    // toast
)
```

ezr-ondrej:
Why is staying on the page an issue? We leave up to the user what he/she wants to do?
Or we can delay the redirect for a second if we insist on it :thinking:  
I dont think relying on our slow app (and thus redirects) is the best solution... what if we accidently speed our app? xD

shiramax:
@ezr-ondrej @xprazak2 I will delay the redirection and add a notification, thanks

xprazak2:
I did not mean delaying the redirect. I did not mean relying on application being slow. I merely meant have acceptable user experience given the constraints of our hybrid UI approach, which does not make it easy in this case. It is also possible to dispatch the message before dispatching redirect if that helps:

```js
handleSuccess: (response) => {
    dispatch(addToast({ type: "success", message: "Import started" }))
    dispatch(push(ANSIBLE_ROLES_INDEX));
},

> Why is staying on the page an issue? 

Because the form has been already submitted and the presented information is no longer relevant.

xprazak2:
Works for me. Any additional comments @ezr-ondrej?

ezr-ondrej (approved):
Would be great if we could disable the submit button after clicking it as there is no response for a bit, so user doesn't have immediate feedback that they really clicked the button, but otherwise works as expected :+1:

xprazak2:
Thanks!

PR #389, opened 2021-03-02
Refs #31980 - improve Ansible as authorized feature

ezr-ondrej:
In 1b90f872697c272511e95818bb986c79796fd367 we allowed Ansible as
authorized feature on smart proxy, but we acutally overriden the Array
so Puppet wouldn't be authorized_feature.

Also adding a test that config report can be created with proxy only
having Ansible feature.

### DIFF
--- a/app/services/foreman_ansible/ansible_report_importer.rb
+++ b/app/services/foreman_ansible/ansible_report_importer.rb
@@ -6,7 +6,12 @@ module ForemanAnsible
   # to the right hostname in Foreman
   module AnsibleReportImporter
     extend ActiveSupport::Concern
+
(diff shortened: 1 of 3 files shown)


domq:
Ah, my bad - Ther's still so much I don't know about Ruby! Thanks for the fix.

ezr-ondrej:
> Ah, my bad - Ther's still so much I don't know about Ruby! Thanks for the fix.

No worries this is quite a strange ruby behaviour.
It's more of a my fault, I should have test it better before merging ;)

ezr-ondrej:
This has been fixed by #491, closing

PR #390, opened 2021-03-05
Fixes #32020 - Allow disabling foreman callback plugin

adamruzicka:
unless applying roles

requires:
- [x] https://github.com/theforeman/foreman-ansible-modules/pull/1161

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -8,9 +8,11 @@ module ForemanAnsibleCore
       def initialize(input, suspended_action:)
         super input, :suspended_action => suspended_action
         @inventory = rebuild_secrets(rebuild_inventory(input), input)
-        @playbook = input.values.first[:input][:action_input][:script]
+        action_input = input.values.first[:input][:action_input]
+        @playbook = action_input[:script]
         @root = working_dir
-        @verbosity_level = input.values.first[:input][:action_input][:verbosity_level]
+        @verbosity_level = action_input[:verbosity_level]
+        @rex_command = action_input[:remote_execution_command]
       end
 
       def start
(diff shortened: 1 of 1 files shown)


lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -104,7 +106,9 @@ def write_playbook
       end
 
       def start_ansible_runner
-        command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        env = {}
+        env['FOREMAN_DISABLE'] = '1' if @rex_command
+        command = [env, 'ansible-runner', 'run', @root, '-p', 'playbook.yml']
```
ezr-ondrej, line 111: Is this a correct way to set env variables? passing the env hash seems to be quite wild.

ezr-ondrej (commented):
This doesn't seem to work, the rex_command is set and I'm quite sure I'm running the right version of callback.

lib/foreman_ansible_core/runner/ansible_runner.rb
adamruzicka: Not sure if it is the correct way, but it is the way we did it for ansible-playbook

ezr-ondrej: Ignore me, I was using wrong callback :) It works... still looks weird, but works xD

ezr-ondrej (approved):
This works great, just small nits in the modules PR after testing :)

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -104,7 +106,9 @@ def write_playbook
       end
 
       def start_ansible_runner
-        command = ['ansible-runner', 'run', @root, '-p', 'playbook.yml']
+        env = {}
+        env['FOREMAN_DISABLE'] = '1' if @rex_command
```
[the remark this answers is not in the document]
adamruzicka: Changed to `FOREMAN_CALLBACK_DISABLE`

ezr-ondrej (approved):
This will work only with next release of `foreman-ansible-modules` but this will not break without is, just won't work, so no need to wait for it :)

PR #391, opened 2021-03-07
Fixes #32030 - Don't crash upon `localhost` Ansible tasks

domq:
When ansible-runner executes a local task (through `delegate_to: localhost` or any other available mechanism in Ansible to do so), it breaks the implicit assumption that all logs are related to a host that we know about (with “we” being Foreman in general, and the `inputs["targets"]` Dynflow structure in this particular case).
- Broadcast log entries for `localhost`, under the assumption that they consist of some kind of prep step that concerns all hosts
- Still fail in case we are presented with a log entry that is for an unknown host

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -63,7 +63,9 @@ module ForemanAnsibleCore
 
       def handle_host_event(hostname, event)
         log_event("for host: #{hostname.inspect}", event)
-        publish_data_for(hostname, event['stdout'] + "\n", 'stdout') if event['stdout']
+        if event['stdout'] && @outputs[hostname] # rubocop:disable Style/IfUnlessModifier
+          publish_data_for(hostname, event['stdout'] + "\n", 'stdout')
+        end
         case event['event']
         when 'runner_on_ok'
           publish_exit_status_for(hostname, 0) if @exit_statuses[hostname].nil?
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -63,8 +63,14 @@ module ForemanAnsibleCore
 
       def handle_host_event(hostname, event)
         log_event("for host: #{hostname.inspect}", event)
-        if event['stdout'] && @outputs[hostname] # rubocop:disable Style/IfUnlessModifier
-          publish_data_for(hostname, event['stdout'] + "\n", 'stdout')
+        if event['stdout']
+          if @outputs[hostname]
+            publish_data_for(hostname, event['stdout'] + "\n", 'stdout')
+          elsif hostname == 'localhost'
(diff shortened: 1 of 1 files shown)


domq:
Fixes [#32030](https://projects.theforeman.org/issues/32030)

adamruzicka:
As you mentioned Foreman assumes it knows about every single host in the job and each event generated by ansible-runner is related either to a specific known host or to all hosts. If the assumption doesn't hold, it crashes. As far as I'm concerned, having an unknown host in the run is an error state since it deviates from the intended use and should be treated accordingly.

If I'm reading this right, with this patch all events which belong to any unexpected host get silently discarded. This isn't really ideal. If the job does something, then the user should be able to find out somewhere.

 It seems what you want to accomplish is more in line with what https://github.com/ATIX-AG/foreman_acd does

domq:
I understand your line of reasoning, however sometimes Ansible needs to prep things on the local host e.g. to generate secrets etc. Would you be OK if I changed the PR to special-case `localhost` only?

adamruzicka:
@ezr-ondrej any opinions?

ezr-ondrej:
I'd agree that localhost preparations is quite a valid use case for ansible and thus we should support it. While playing with this I've come up with a fix for another bug we are facing with this reporting: https://projects.theforeman.org/issues/32188.

I've opened #398 containing both the fixes, just moving the unknown host guard to `handle_event_file`

ezr-ondrej:
We should instead deny running tasks against `localhost`

domq:
> We should instead deny running tasks against `localhost`

Please don't do this. A number of Ansible use cases don't want to be forced to use ssh (e.g. when managing Kubernetes objects).

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -63,7 +63,15 @@ def handle_event_file(event_file)
...
       def handle_host_event(hostname, event)
         log_event("for host: #{hostname.inspect}", event)
-        publish_data_for(hostname, event['stdout'] + "\n", 'stdout') if event['stdout']
+        if event['stdout']
+          if @outputs[hostname]
+            publish_data_for(hostname, event['stdout'] + "\n", 'stdout')
+          elsif hostname == 'localhost'
+            broadcast_data(event['stdout'] + "\n", 'stdout')
+          else
+            raise "handle_host_event: unknown host #{hostname}"
```
adamruzicka, line 72: Since this PR was opened, we moved to processing the artifact files on demand. That means this raise does not really make sense anymore because the events may be processed after the job has already finished.

From where I stand we have essentially three options:
1) Broadcast events from unknown hosts the same way you do for localhost
2) Broadcast events from unknown hosts, but mark them as unexpected by broadcasting them as stderr
3) Just log those as errors

adamruzicka:
The issue this tried to resolve should be fixed by https://github.com/theforeman/smart_proxy_ansible/pull/55 . Closing, thank you for proposing these changes

PR #393, opened 2021-03-09
Fixes #32044 - Do not store per-host inputs in two copies

adamruzicka:
This change wipes unneeded input passed to OutputCollectorActions.

Requires https://github.com/theforeman/foreman-tasks/pull/617 to be able to match actions when retrieving outputs on-demand


ares (approved):
pending the linked PR

ezr-ondrej (approved):
Make sense :+1: 
But the naming is terribly missleading, it took me almost half an hour to figure out why in `runner_input` we are building it, and then in `transform_input` filtering it out again.
My :brain::fire:

ezr-ondrej:
Should we bump the dependencies (`foreman-tasks`)?

adamruzicka:
Bumped the version, tests will most likely be red until new tasks are released.

>  might be worth renaming to transform_child_input

Naming is hard. It would need to happen in tasks and then propagate here. Before we do that we should be 100% sure about the new name.

ezr-ondrej:
> Naming is hard

I know, it wasn't a blocker comment, just a note ;)

ezr-ondrej:
Maybe we can just depend on newer REX, that would already depend on newer tasks?
I'm ok if we just keep it in mind actually...

adamruzicka:
So far there is no real reason why rex should depend on anything added to tasks after 4.0.0 :/

ezr-ondrej:
Rubocop doesn't like the order in gemfile.
Ideally I'd like to avoid depending on tasks directly if possible :(

ezr-ondrej:
@adamruzicka I guess now we should depend on the yet unreleased version of REX? :thinking:

adamruzicka:
> @adamruzicka I guess now we should depend on the yet unreleased version of REX? thinking

Just so we can get a transitive dependency, yes. Can't say we really helped ourselves by not depending on tasks directly

ezr-ondrej:
> Can't say we really helped ourselves by not depending on tasks directly

Well, I don't like double dependencies, but yeah, we probably did not. But I guess I would be fine with us just saying "tasks and rex" are coordinating their releases and have always the same release manager (and it's accidently always named Adam) so we can have just release process, that will always bump version of tasks in  REX and here we just depend on newer rex when releasing if necessary. Or just merge Tasks to core xD

adamruzicka:
[test foreman_ansible]

ezr-ondrej:
I guess we bump the dependency during the release process? if so, I'm ok to just merge as-is and handle dependency elsewhere.

adamruzicka:
Bumped the dependency version so it doesn't get forgotten.

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #395, opened 2021-03-12
Fixes #32080 - Run Ansible with flags for tags

xprazak2:
Adds support for --tags and --skip-tags

Needs https://github.com/theforeman/foreman_remote_execution/pull/569

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -18,6 +18,10 @@ if defined? ForemanRemoteExecution
           'Ansible'
         end
 
+        def provider_input_namespace
+          :ansible
+        end
+
         def proxy_command_options(template_invocation, host)
           super(template_invocation, host).merge(
             'ansible_inventory' => ::ForemanAnsible::InventoryCreator.new(
@@ -57,6 +61,51 @@ if defined? ForemanRemoteExecution
 
         def supports_effective_user?
           true
+        end
+
+        def provider_inputs
+          [
+            ForemanRemoteExecution::ProviderInput.new(
+              name: 'tags',
+              label: _('Tags'),
+              value: '',
+              value_type: 'plain',
+              description: 'Tags used for Ansible execution'
+            ),
+            ForemanRemoteExecution::ProviderInput.new(
+              name: 'tags_flag',
+              label: _('Include/Exclude Tags'),
+              value: 'include',
+              description: 'Option whether to include or exclude tags',
+              options: "include\nexclude"
+            )
+          ]
+        end
+
+        def provider_inputs_doc
+          opts = provider_inputs.find { |input| input.name == 'tags_flag' }.options.split("\n")
+          {
+            :namespace => provider_input_namespace,
+            :opts => { :desc => N_('Ansible provider specific inputs') },
+            :children => [
+              {
+                :name => :tags,
+                :type => Array,
+                :opts => { :required => false, :desc => N_('A comma separated list of tags to use for Ansible run') }
+              },
+              {
+                :name => :tags_flag,
+                :type => opts,
+                :opts => { :required => false, :desc => N_('Include\Exclude tags for Ansible run') }
+              }
+            ]
+          }
+        end
+
+        def proxy_command_provider_inputs(template_invocation)
+          tags = template_invocation.provider_input_values.find_by(:name => 'tags')&.value || ''
+          tags_flag = template_invocation.provider_input_values.find_by(:name => 'tags_flag')&.value || ''
+          { :tags => tags, :tags_flag => tags_flag }
         end
 
         def proxy_operation_name
--- a/app/views/foreman_ansible/job_templates/ansible_roles_-_ansible_default.erb
+++ b/app/views/foreman_ansible/job_templates/ansible_roles_-_ansible_default.erb
@@ -15,10 +15,14 @@ model: JobTemplate
     - name: Display all parameters known for the Foreman host
       debug:
         var: foreman
+      tags:
+        - always
   tasks:
     - name: Apply roles
       include_role:
         name: "{{ role }}"
+      tags:
+        - always
       loop: "{{ foreman_ansible_roles }}"
       loop_control:
         loop_var: role


app/controllers/foreman_ansible/api/v2/job_invocations_controller_extensions.rb
```diff
@@ -0,0 +1,20 @@
+module ForemanAnsible
+  module Api
+    module V2
+      module JobInvocationsControllerExtensions
+        module ApipieExtensions
+          extend Apipie::DSL::Concern
+
+          opts = ::ForemanAnsible::AnsibleProvider.provider_inputs.find { |input| input.name == 'tags_flag' }.options.split("\n")
```
ezr-ondrej, line 8: Would this be worth adding to the REX PR?
We should probably go over all providers and add their respective parameters there, wouldn't it be easier?

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -29,7 +33,7 @@ def proxy_command_options(template_invocation, host)
             ),
             :name => host.name,
             :check_mode => host.host_param('ansible_roles_check_mode')
-          )
+          ).merge(proxy_command_provider_inputs(template_invocation))
```
ezr-ondrej, line 36: This could also probably be generalized and done in the REX PR, right? :thinking:

ezr-ondrej (commented):
Looks good overall, just the two generalizations that could probably be done here.
If you think its better this way, I'm open to it, just suggestions :)

app/controllers/foreman_ansible/api/v2/job_invocations_controller_extensions.rb
ezr-ondrej: We are doing something along the lines for facets (note its not great example, because it doesn't work very well, but as a starting point )
https://github.com/theforeman/foreman/blob/fbf9bfef48fda7404e7b1b59fd0f1094bbd1a3ce/app/controllers/concerns/foreman/controller/parameters/host.rb#L28
https://github.com/theforeman/foreman/blob/fbf9bfef48fda7404e7b1b59fd0f1094bbd1a3ce/app/controllers/api/v2/hosts_controller.rb#L117

xprazak2:
I made some of the things more general and moved them to rex.

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -109,15 +111,29 @@ def write_playbook
...
+      def cmdline
+        cmd = [tags_cmd, check_cmd].reduce("") do |memo, item|
+          item.empty? ? memo : "#{memo} #{item}"
+        end
+        cmd.empty? ? '' : "--cmdline \"#{cmd}\""
+      end
+
+      def tags_cmd
+        flag = @tags_flag == 'include' ? '--tags' : '--skip-tags'
+        @tags.empty? ? '' : "#{flag} #{@tags}"
```
ezr-ondrej, line 128: I think that we have to accept arrays as well as strings for `@tags` (when I pass --tags="tag1,tag2" to hammer, this will be an array) and endup with `--tags ["tag1", "tag2"]` so what about:
```suggestion
        @tags.empty? ? '' : "#{flag} #{Array(@tags).join(',')}"
```

ezr-ondrej (changes requested):
Multiple tags passed from hammer command doesn't work well and I'd love to have documentation for format to pass the tags in in UI, apart of that works well :+1:

xprazak2:
needs https://github.com/theforeman/foreman_remote_execution/pull/590 for tag values to be added into playbook.

lib/foreman_ansible/engine.rb
```diff
@@ -93,6 +93,7 @@ class Engine < ::Rails::Engine
       ::Api::V2::HostgroupsController.include ForemanAnsible::Api::V2::HostgroupsControllerExtensions
       ::Api::V2::HostgroupsController.include ForemanAnsible::Api::V2::HostgroupsParamGroupExtensions
       ::ConfigReportImporter.include ForemanAnsible::AnsibleReportImporter
+      ::ForemanRemoteExecution::Renderer::Scope::Input.prepend ForemanAnsible::Renderer::Scope::AnsibleTagsMacro
```
ezr-ondrej, line 96: Why are we not using the plugin [extension point](https://github.com/theforeman/foreman/blob/e49da5764920b3810e1c33ea2517ec8c870362b8/app/registries/foreman/plugin.rb#L434) for this?

xprazak2: Because we extend the scope from rex here rather than the one from core.

app/views/foreman_ansible/job_templates/ansible_roles_-_ansible_default.erb
```diff
@@ -15,10 +15,12 @@ model: JobTemplate
     - name: Display all parameters known for the Foreman host
       debug:
         var: foreman
+      <%= add_ansible_tags %>
   tasks:
     - name: Apply roles
       include_role:
         name: "{{ role }}"
+      <%= add_ansible_tags %>
```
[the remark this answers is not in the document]
xprazak2: Changed.

app/lib/foreman_ansible/renderer/scope/ansible_tags_macro.rb
```diff
@@ -0,0 +1,22 @@
...
+      module AnsibleTagsMacro
+        extend ApipieDSL::Module
+
+        apipie :method, 'Returns value of provider template input' do
+          returns String, desc: 'The value of template provider input'
+          example 'add_ansible_tags #=> "tags: [foo, bar]"'
+        end
+        def add_ansible_tags
+          return '' if provider_input_values['tags'].blank?
+          "tags: [#{provider_input_values['tags']}]"
```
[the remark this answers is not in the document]
xprazak2: I improved the condition.

lib/foreman_ansible/engine.rb
ezr-ondrej: Right, I'd rather see us having similar way of extension, but you are right, it is not there now.
just we are patching module and that might fail when include of that module happens prior to the patch, so this is bit fragile.

xprazak2: Any suggestions for a better approach? By the way, which module do you mean? `::ForemanRemoteExecution::Renderer::Scope::Input` is a [class](https://github.com/theforeman/foreman_remote_execution/blob/master/app/lib/foreman_remote_execution/renderer/scope/input.rb#L4)...

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -109,15 +111,29 @@ def write_playbook
...
-        logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
+        command = "ansible-runner run #{@root} -p playbook.yml #{cmdline} #{verbosity}"
+        initialize_command(env, command)
+        logger.debug("[foreman_ansible] - Running command '#{command}'")
+      end
+
+      def cmdline
+        cmd = [tags_cmd, check_cmd].reduce("") do |memo, item|
+          item.empty? ? memo : "#{memo} #{item}"
+        end
```
ezr-ondrej, line 122: wouldn't simple `[tags_cmd, check_cmd].reject(&:empty?).join(' ')` work here?

Or maybe even:
```ruby
cmd_args = [tags_cmd, check_cmd].reject(&:empty?)
return unless cmd_args.any?
"--cmdline \"#{cmd_args.joins(' ')}\""
```

app/views/foreman_ansible/job_templates/ansible_roles_-_ansible_default.erb
```diff
@@ -15,10 +15,12 @@ model: JobTemplate
     - name: Display all parameters known for the Foreman host
       debug:
         var: foreman
+      <%= ansible_tags %>
```
ezr-ondrej, line 18: can be quite simplified by usage of special ansible tag:
```suggestion
      tags:
      - always
```

ezr-ondrej (commented):
I think we are overdoing it here, I found special `always` tag, that runs unless explicitly asked not to.

xprazak2:
Using 'always' tag makes it more simple, thanks.

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -109,15 +111,28 @@ def write_playbook
       def start_ansible_runner
         env = {}
         env['FOREMAN_CALLBACK_DISABLE'] = '1' if @rex_command
-        command = [env, 'ansible-runner', 'run', @root, '-p', 'playbook.yml']
-        command << '--cmdline' << '"--check"' if check_mode?
-        command << verbosity if verbose?
-        initialize_command(*command)
-        logger.debug("[foreman_ansible] - Running command '#{command.join(' ')}'")
+        command = "ansible-runner run #{@root} -p playbook.yml #{cmdline} #{verbosity}"
+        initialize_command(env, command)
```
adamruzicka, line 115: Is there a reason why we switch from the array form to an escaped string? It feels less safe, especially if part of the string is unsanitized user input

xprazak2: The quotes were not correctly recognized when using array which was causing malformed commands.

adamruzicka: Which quotes?

ezr-ondrej: Would it work as array if we would pass all `cmdline` as one arg?
```ruby
command << '--cmdline' << cmdline
```

xprazak2: > Which quotes?
Quotes delimiting an argument for runner command line: `--cmdline "--tags foo"`

xprazak2: > Would it work as array if we would pass all `cmdline` as one arg?
> 
> ```ruby
> command << '--cmdline' << cmdline
> ```

Nope, that does not work either.

adamruzicka: > Quotes delimiting an argument for runner command line: `--cmdline "--tags foo"`

Represented in array form as `['--cmdline', '"--tags foo"']`?

xprazak2: > > Quotes delimiting an argument for runner command line: `--cmdline "--tags foo"`
> 
> Represented in array form as `['--cmdline', '"--tags foo"']` or `['--cmdline "--tags foo"']`?

I tried both, neither works as expected.

adamruzicka: Well, neither of those is supposed to work as there are superfluous quotes. `['--cmdline=foo,baz']` or `['--cmdline', 'foo,baz']` should work

ezr-ondrej: we need those quotes tho, it is one argument, that contains arguments for some subprocess

ezr-ondrej: one way around it, would be dumping these args to folder, which ansible-runner reads. I think it is `cmd`

ezr-ondrej:
we've moved the code to the smart_proxy_ansible, so proxy changes will need to go there now.

lib/foreman_ansible_core/runner/ansible_runner.rb
xprazak2: Do you mean [env/cmdline](https://ansible-runner.readthedocs.io/en/stable/intro.html#env-cmdline) or is there something else?

adamruzicka: > we need those quotes tho, it is one argument, that contains arguments for some subprocess

The array form is not processed by a shell, a single entry in the array is a single argument being passed to the child process, I wouldn't say the quotes are necessary.

> ['--cmdline', 'foo,baz'] gives me --cmdline foo,baz while I need --cmdline "--tags foo,baz". The --tags flag together with list of the tags is an arg for the --cmdline.

Right, my bad. I meant `['--cmdline', '--tags=foo,baz']`

xprazak2: I haven't tried with the `=` which might help just for tags but when I add a flag for check mode into the mix, I still need those quotes: `--cmdline "--tags=foo,baz --check"`

ezr-ondrej (approved):
Works nicely :+1:

PR #396, opened 2021-03-17
Fixes #31991 - API for importing roles & variables

shiramax:
- Change fetch roles to show all roles that needs a sync, with more data: the action of the roles, and the amount
  of variables to be synced and more
- Change import roles to sync the seleced roles and their variables
  togther
- Add deprecation warnnings for: obselte roles, obslete variables and
  import variables

### DIFF
--- a/app/controllers/api/v2/ansible_roles_controller.rb
+++ b/app/controllers/api/v2/ansible_roles_controller.rb
@@ -5,6 +5,7 @@ module Api
     # API controller for Ansible Roles
     class AnsibleRolesController < ::Api::V2::BaseController
       include ::Api::Version2
+      include ::ForemanAnsible::AnsibleRolesDataPreparations
 
       resource_description do
         api_version 'v2'
@@ -12,8 +13,8 @@ module Api
       end
 
       before_action :find_resource, :only => [:show, :destroy]
-      before_action :find_proxy, :only => [:import, :obsolete, :fetch]
-      before_action :create_importer, :only => [:import, :obsolete, :fetch]
+      before_action :find_proxy, :only => [:import, :obsolete, :fetch, :sync]
+      before_action :create_importer, :only => [:import, :obsolete, :fetch, :sync]
 
       api :GET, '/ansible_roles/:id', N_('Show role')
       param :id, :identifier, :required => true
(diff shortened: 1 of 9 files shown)


xprazak2:
Works nicely, since we are doing a sync on background now, could we have a new endpoint for that? That way we do not change the response data for the existing endpoint.

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,30 @@ def destroy
...
-        @imported = @importer.import!(role_names)
+        params = @importer.import!(role_names)
+        task = @importer.confirm_import(params)
+        render json: {
+          task: task
+        }, status: :ok
       end
 
-      api :PUT, '/ansible_roles/obsolete', N_('Obsolete Ansible roles')
+      api :PUT, '/ansible_roles/obsolete', N_('DEPRECATED: Obsolete Ansible roles')
```
ofedoren, line 47: Please, use
```ruby
, deprecated: true
```
option when specifying deprecated endpoints.

I've noticed this should be added to few more places.

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,46 @@ def destroy
         process_response @ansible_role.destroy
       end
 
-      api :PUT, '/ansible_roles/import', N_('Import Ansible roles')
+      api :PUT, '/ansible_roles/import', N_('DEPRECATED: Import Ansible roles'), deprecated: true
       param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to import from')
-      param :role_names, Array, N_('Ansible role names to import')
+      param :role_names, Array, N_('Ansible role names to imported')
```
ofedoren, line 38: Is there a reason to change it? Maybe it's `to be imported`?

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,46 @@ def destroy
...
       def import
-        @imported = @importer.import!(role_names)
+        Foreman::Deprecation.api_deprecation_warning(_('Use /ansible_roles/sync to sync role with ansible smart proxy'))
```
ofedoren, line 40: If this is going to be translated then `/ansible_roles/sync` should be extracted from the string. And probably it should be something like `to sync roles from Smart Proxy with Ansible feature enabled` or similar...

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,46 @@ def destroy
...
+        if role_names.present?
+          new_roles.select! do |role|
+            role_names.include?(role.name)
+          end
+        end
+        new_roles.map(&:save)
+        @imported = new_roles
+      end
+
+      api :PUT, '/ansible_roles/sync', N_('DEPRECATED: Sync Ansible roles')
```
ofedoren, line 51: Am I missing something? You've deprecated `/import` to use `/sync` and this is also deprecated?

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,46 @@ def destroy
...
+      param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
+      param :role_names, Array, N_('Ansible role names to synced')
```
ofedoren, line 53: Same as the first comment.

app/controllers/api/v2/ansible_roles_controller.rb
```diff
@@ -32,28 +33,46 @@ def destroy
...
+      def sync
+        params = @importer.import!(role_names)
+        if params['changed'].present?
+          @task = @importer.confirm_sync(params)
+        else
+          process_resource_error({ :message => 'Roles could not be synced' })
```
ofedoren, line 59: Shouldn't it be translated?

lib/foreman_ansible/register.rb
```diff
@@ -24,7 +24,7 @@
                  :'api/v2/ansible_roles' => [:destroy, :obsolete] },
                :resource_type => 'AnsibleRole'
     permission :import_ansible_roles,
-               { :ansible_roles => [:import, :confirm_import],
+               { :ansible_roles => [:import, :confirm_import, :sync],
                  :'api/v2/ansible_roles' => [:import] },
```
xprazak2, line 28: Should `:sync` be added into the array for `api/v2/ansible_roles` instead?

xprazak2:
Works well, test failures seem related.

lib/foreman_ansible/register.rb
shiramax: right, done.

xprazak2:
Only 2 test errors remaining now.

shiramax:
[foreman ansible]

xprazak2:
Thanks!

PR #400, opened 2021-03-26
Fixes #30090 - fix the controller permission mapping

ares:
The AnsibleVariables controller inherits from LookupKeys controller
which means it inherits it's permission name. We have separate
permission for ansible variables so we need to define that explicitly.

To reproduce the issue, just create non-admin user and try to list
ansible variables, no variable is displayed. With this patch, correct
permission name is constructed `view_ansible_variables` instead of
`view_external_variables` and all variables are correctly displayed.

### DIFF
--- a/app/controllers/ansible_variables_controller.rb
+++ b/app/controllers/ansible_variables_controller.rb
@@ -104,4 +104,8 @@ class AnsibleVariablesController < ::LookupKeysController
     end
     @smart_proxy
   end
+
+  def controller_permission
+    'ansible_variables'
+  end
 end


ares:
FTR, api is unaffected

ezr-ondrej:
This is correct with current permission model, but just to inform myself, do we want to have some general permission giving the user to see all the lookup variables, or do we just want to have specific permissions?
If we want to have only the specific permissions, we (while extracting puppet) should mark the controller as abstract and force descendants to override this, now it's easy to forget.

ezr-ondrej (approved):

xprazak2:
Needs a rebase.

ezr-ondrej (approved):
I've rebased and I believe in current permission model this is correct and we should get it in.
Agreed @xprazak2 ?

PR #402, opened 2021-04-04
Fixes #32248 - add check mode detail to ansible report

yifatmakias:
This PR is blocked by:
https://github.com/theforeman/foreman-ansible-modules/pull/1190

### DIFF
--- a/app/helpers/foreman_ansible/ansible_reports_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_reports_helper.rb
@@ -19,6 +19,14 @@ module ForemanAnsible
       name
     end
 
+    def ansible_run_in_check_mode?(log)
+      log.message&.value == 'check_mode_enabled' if check_mode_log?(log)
+    end
+
+    def check_mode_log?(log)
+      log.source&.value == 'check_mode'
+    end
+
     def ansible_module_args(log)
       report_json_viewer module_invocations parsed_message_json(log)
     end
--- a/app/services/foreman_ansible/ansible_report_scanner.rb
+++ b/app/services/foreman_ansible/ansible_report_scanner.rb
@@ -5,7 +5,21 @@ module ForemanAnsible
   # sets the origin of the report to 'Ansible'
   class AnsibleReportScanner
     class << self
-      def add_reporter_data(report, raw); end
+      def add_reporter_data(_report, raw)
+        check_mode_message = raw['check_mode'] ? 'check_mode_enabled' : 'check_mode_disabled'
+        check_mode_log = {
+          'log': {
+            'sources': {
+              'source': 'check_mode'
+            },
+            'messages': {
+              'message': check_mode_message
+            },
+            'level': 'info'
+          }
+        }
+        raw['logs'].insert(0, check_mode_log.deep_stringify_keys)
+      end
 
       def identify_origin(raw)
         'Ansible' if ansible_report?(raw)
--- a/app/views/foreman_ansible/config_reports/_ansible.html.erb
+++ b/app/views/foreman_ansible/config_reports/_ansible.html.erb
@@ -1,6 +1,11 @@
 <%= webpacked_plugins_js_for :foreman_ansible %>
 <%= stylesheet 'foreman_ansible/foreman-ansible' %>
+<% if ansible_run_in_check_mode?(logs.first) %>
+  <%= alert :class => 'alert-info',
+            :header => _('Ansible check mode'),
+            :text => _('Notice that ansible roles run in check mode.')%>
 
+<% end %>
 <table id='report_log' class="<%= table_css_classes %>">
   <thead>
     <tr>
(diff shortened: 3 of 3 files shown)


app/services/foreman_ansible/ansible_report_scanner.rb
```diff
@@ -5,7 +5,13 @@ module ForemanAnsible
   # sets the origin of the report to 'Ansible'
   class AnsibleReportScanner
     class << self
-      def add_reporter_data(report, raw); end
+      def add_reporter_data(report, raw)
+        check_mode = raw['check_mode'] == '1'
+        message = Message.find_or_create check_mode
+        source  = Source.find_or_create 'check_mode'
+        log = Log.create(:message_id => message.id, :source_id => source.id, :report => report, :level => :info)
```
xprazak2, line 12: I am getting errors about log being invalid - the report is not persisted at this point so we do not have a report id to use when saving the log. You could try to add the log data into `raw['logs']` and then let the [core logic](https://github.com/theforeman/foreman/blob/develop/app/services/report_importer.rb#L84-L104) handle the log creation or maybe `report.logs.build` could help.

yifatmakias: Thanks, I will try to do that.

app/helpers/foreman_ansible/ansible_reports_helper.rb
```diff
@@ -19,6 +19,14 @@ def ansible_module_name(log)
       name
     end
 
+    def ansible_run_in_check_mode?(log)
+      log.message&.value == 't' if check_mode_log?(log)
```
xprazak2, line 23: `add_reporter_data` sets message value as boolean, but the db column is a text, so I think that is where the `t` comes from. Can we set the value to something more meaningful?

xprazak2:
Nice, the banner shows the warning:

check-mode

app/helpers/foreman_ansible/ansible_reports_helper.rb
yifatmakias: @xprazak2 What do you mean by more meaningful? a longer message that says the check mode was enabled?

xprazak2: If the message value is a string, I do not think we should assign a boolean. Any reasonable string value would be ok, for example `check_mode_enabled`.

yifatmakias: Sounds good, I will fix this.

xprazak2:
Looks good overall, it should be probably merged at the same time as changes for the callback.

ezr-ondrej (changes requested):
I'm not sure we want to hide the logs completely... am I missing something?

app/helpers/foreman_ansible/ansible_reports_helper.rb
```diff
@@ -19,6 +19,14 @@ def ansible_module_name(log)
       name
     end
 
+    def ansible_run_in_check_mode?(log)
+      log.message&.value == 'check_mode_enabled' if check_mode_log?(log)
```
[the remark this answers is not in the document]
xprazak2: No, it is set to be created by the AnsibleReportScanner as a way of adding information about check mode into the report.

app/views/foreman_ansible/config_reports/_ansible.html.erb
```diff
@@ -1,6 +1,11 @@
 <%= webpacked_plugins_js_for :foreman_ansible %>
 <%= stylesheet 'foreman_ansible/foreman-ansible' %>
+<% if ansible_run_in_check_mode?(logs.first) %>
```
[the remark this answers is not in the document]
xprazak2: Having a banner is one of the requirements, so I would definitely keep it. Storing the information about check mode in log is just a way to have 'meta' information about the report so that we know if the banner should be displayed, so I think it is ok to hide it.

app/views/foreman_ansible/config_reports/_ansible.html.erb
```diff
@@ -12,12 +17,14 @@
   </thead>
   <tbody>
     <% logs.each do |log| %>
-      <tr>
-        <td><span <%= report_tag log.level %>><%= h log.level %></span></td>
-        <td><%= ansible_module_name(log) %></td>
-        <td><%= ansible_module_args(log) %></td>
-        <td><%= ansible_module_message(log) %></td>
-      </tr>
+      <% unless check_mode_log?(log) %>
```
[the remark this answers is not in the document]
xprazak2: Because that log contains information whether the run was executed in check mode and was created by report scanner rather than generated by Ansible.

app/views/foreman_ansible/config_reports/_ansible.html.erb
```diff
@@ -1,6 +1,11 @@
 <%= webpacked_plugins_js_for :foreman_ansible %>
 <%= stylesheet 'foreman_ansible/foreman-ansible' %>
+<% if ansible_run_in_check_mode?(logs.first) %>
+  <%= alert :class => 'alert-info',
+            :header => _('Ansible check mode - '),
```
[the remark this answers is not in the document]
yifatmakias: Removed the dash.

ezr-ondrej (commented):
I think I'm still missing the point of why we throw all the log messages out of the window?
The check mode message is fine workaround and I would not mind to have it as first log message, but why do we throw all the other ones?
It might be quite interesting especially when some failures appear - that may happen in check mode as well and then I want to see the log for it.

yifatmakias:
> I think I'm still missing the point of why we throw all the log messages out of the window?
> The check mode message is fine workaround and I would not mind to have it as first log message, but why do we throw all the other ones?
> It might be quite interesting especially when some failures appear - that may happen in check mode as well and then I want to see the log for it.

I explained in person but I wanted to be here as well. I am not throwing all the log messages I am just throwing the log that contains the check mode information so it will not appear in the logs table.

ezr-ondrej (approved):
Sorry for the mess! I've missread the code. Let it be a lesson for me, always test before checking the code :facepalm:

ezr-ondrej (approved):
Works as expected :+1:

PR #405, opened 2021-04-20
Refs #31991 - Do not respond with error on no changes

xprazak2:
When there are no changes in roles on proxy,
we currently respond with 500 - but there were
actually no errors and everything completed,
only no changes were detected.


ezr-ondrej (approved):
Thanks @xprazak2 ! :+1:

PR #406, opened 2021-04-22
Fixes #32399 - Add Ansible tab to host detail page

xprazak2:
Registeres a new tab on host detail page for Ansible
content.

### DIFF
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -96,6 +96,8 @@ Foreman::Plugin.register :foreman_ansible do
   parameter_filter Host::Managed, base_role_assignment_params.merge(:host_ansible_roles_attributes => {})
   parameter_filter Hostgroup, base_role_assignment_params.merge(:hostgroup_ansible_roles_attributes => {})
 
+  register_global_js_file 'global'
+
   divider :top_menu, :caption => N_('Ansible'), :parent => :configure_menu
   menu :top_menu, :ansible_roles,
        :caption => N_('Roles'),
--- a/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
+++ b/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
@@ -1,0 +1,33 @@
+import React, { useState } from 'react';
+import { Tabs, Tab, TabTitleText, Label } from '@patternfly/react-core';
+import { InfoCircleIcon } from '@patternfly/react-icons';
+
+import { translate as __ } from 'foremanReact/common/I18n';
+
+import './AnsibleHostDetail.scss';
+
+const AnsibleHostDetail = props => {
+  // https://projects.theforeman.org/issues/32398
+  const [activeTab] = useState('variables');
+
+  return (
+    <div className="ansible-host-detail">
+      <Tabs activeKey={activeTab} isSecondary>
+        <Tab
+          eventKey="variables"
+          title={<TabTitleText>{__('Variables')}</TabTitleText>}
+        >
+          <Label
+            color="blue"
+            icon={<InfoCircleIcon />}
+            style={{ marginTop: '1.5rem' }}
+          >
+            Ansible Variables coming soon!
+          </Label>
+        </Tab>
+      </Tabs>
+    </div>
+  );
+};
+
+export default AnsibleHostDetail;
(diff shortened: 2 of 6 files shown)


xprazak2:
Basic layout for Ansible content on host detail page.

webpack/components/AnsibleRolesAndVariables/__test__/__snapshots__/AnsibleRolesAndVariablesImport.test.js.snap
```diff
@@ -74,6 +74,7 @@ exports[`ImportRolesAndVariablesTable should render 1`] = `
     expandId="expandable-toggle"
     gridBreakPoint="grid-md"
     isStickyHeader={false}
+    isTreeTable={false}
```
xprazak2, line 77: Unrelated change but tests fail without it as the prop was added in a newer version of dependency.

package.json
```diff
@@ -14,6 +14,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
+    "@testing-library/jest-dom": "^5.12.0",
+    "@testing-library/react": "^11.2.6",
```
amirfefer, line 18: We already have these libraries in [@theforeman/test](https://github.com/theforeman/foreman-js/blob/3e463939623016543474c93e9591d4129ee80d91/packages/test/package.json#L28)

Is this version bump require?

lib/foreman_ansible/register.rb
```diff
@@ -96,6 +96,8 @@
   parameter_filter Host::Managed, base_role_assignment_params.merge(:host_ansible_roles_attributes => {})
   parameter_filter Hostgroup, base_role_assignment_params.merge(:hostgroup_ansible_roles_attributes => {})
 
+  register_global_js_file 'fills'
```
amirfefer, line 99: We use a different pattern, instead of having a global file just for fills, we have a global index file for any global need (i.e routes registrations).

For further information and some examples, please have a look at the plugin template: https://github.com/theforeman/foreman_plugin_template/blob/master/webpack/global_index.js

webpack/components/AnsibleHostDetail/AnsibleHostDetail.scss
```diff
@@ -0,0 +1,6 @@
+@import '~@patternfly/patternfly/base/patternfly-variables';
```
amirfefer, line 1: doesn't it a list of PF3 variables?

webpack/components/AnsibleHostDetail/AnsibleHostDetail.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import AnsibleHostDetail from './index';
```
amirfefer, line 5: no need to add index name to import
```suggestion
import AnsibleHostDetail from './';
```

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -0,0 +1,33 @@
+import React, { useState } from 'react';
+import { Tabs, Tab, TabTitleText, Label } from '@patternfly/react-core';
+import InfoCircleIcon from '@patternfly/react-icons/dist/js/icons/info-circle-icon';
```
amirfefer, line 3: This importing causes an issue during build, please use:
```suggestion
import { InfoCircleIcon } from '@patternfly/react-icons';
```

amirfefer (changes requested):
Thanks @xprazak2! 👍 
I left a few inline comments

webpack/components/AnsibleHostDetail/AnsibleHostDetail.scss
xprazak2: [foreman-js](https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/scss/variables.scss#L20) says these are pf4...

webpack/components/AnsibleHostDetail/index.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import AnsibleHostDetail from './AnsibleHostDetail';
+
+const WrappedAnsibleHostDetail = props => <AnsibleHostDetail {...props} />;
```
amirfefer, line 4: You can just do 
```js
export {default as AnsibleHostDetail} from './AnsibleHostDetail';
```
though I understand this is just an initial step, this file will be changed

amirfefer (approved):
Thanks @xprazak2 👍 
Looks great for initial step.

PR #407, opened 2021-04-29
Fixes #32451 - adds convert2rhel job

ares:


app/views/foreman_ansible/job_templates/convert_to_rhel.erb
```diff
@@ -0,0 +1,45 @@
...
+        dest: /usr/share/convert2rhel/subscription-manager/katello-ca-consumer-latest.noarch.rpm
+    - name: Start convert2rhel
+      command: convert2rhel -y --activationkey "<%= activation_key %>" --org "<%= @host.organization.label %>" > /root/convert2rhel.log
+<%- if input('Restart') == "yes" -%>
+    - name: Reboot the machine
+      reboot:
+        reboot_timeout: 1800
+<%- end -%>
+    - name: Update system facts
+      command: subscription-manager facts --update
```
ezr-ondrej, line 45: shouldn't this go before the reboot? :thinking:

xprazak2:
It worked pretty well, facts were uploaded after the reboot so ack from me.

app/views/foreman_ansible/job_templates/convert_to_rhel.erb
```diff
@@ -0,0 +1,45 @@
...
+  advanced: false
+  value_type: plain
+  resource_type: Katello::ActivationKey
+  hidden_value: false
+model: JobTemplate
+job_category: Convert 2 RHEL
+provider_type: Ansible
+kind: job_template
+%>
+<%- activation_key = input('Activation Key').blank? ? "rhel#{@host.os.major}" : input('Activation Key') -%>
```
xprazak2, line 25: I am wondering if this line is needed. Activation key is required, so it should never be blank.

app/views/foreman_ansible/job_templates/convert_to_rhel.erb
ares: no, we need the new facts after the machine is converted to RHEL

app/views/foreman_ansible/job_templates/convert_to_rhel.erb
ares: ah good point, I'll drop this and repush

ares:
updated

xprazak2:
Thanks!

PR #410, opened 2021-05-06
Fixes #32551 - Use private key file path from the proxy

adamruzicka:


xprazak2:
Works for me.

ares:
fixed the issue for me

ezr-ondrej (approved):
LGTM :+1:

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -142,7 +142,12 @@ def rebuild_inventory(input)
         action_inputs = input.values.map { |hash| hash[:input][:action_input] }
         hostnames = action_inputs.map { |hash| hash[:name] }
         inventories = action_inputs.map { |hash| hash[:ansible_inventory] }
-        host_vars = inventories.map { |i| i['_meta']['hostvars'] }.reduce(&:merge)
+        host_vars = inventories.map { |i| i['_meta']['hostvars'] }.reduce({}) do |acc, hosts|
+          hosts.reduce(acc) do |inner_acc, (hostname, vars)|
+            vars[:ansible_ssh_private_key_file] ||= ForemanRemoteExecutionCore.settings[:ssh_identity_key_file]
+            inner_acc.merge(hostname => vars)
+          end
+        end
```
ezr-ondrej, line 150: Do we have to fill the default here for every host?
Wouldn't it be better to do this globally for the whole inventory and leave the per host one blank?

ezr-ondrej (commented):
Once already touching this, I'd have a thought.
But I'm ok to create a RM ticket for it and merge as-is.

lib/foreman_ansible_core/runner/ansible_runner.rb
adamruzicka: iirc currently the key is defined for each host, but the value is optionally nil. We'd have to stop sending it over if the value is nil. Not saying that's bad, but it would probably result in a bigger diff

ezr-ondrej: Agreed https://projects.theforeman.org/issues/32595.

PR #412, opened 2021-05-10
Fixes #32189 - Improve ansible config report data

yifatmakias:
Add textual representation of the ansible task based on data provided by the module, if there's no human readable string, add the message "no data available" information.

### DIFF
--- a/app/helpers/foreman_ansible/ansible_reports_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_reports_helper.rb
@@ -4,19 +4,17 @@ module ForemanAnsible
   # This module takes the config reports stored in Foreman for Ansible and
   # modifies them to be properly presented in views
   module AnsibleReportsHelper
-    ANSIBLE_META_KEYS = %w[
-      _ansible_parsed _ansible_no_log _ansible_item_result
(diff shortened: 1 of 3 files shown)


ares (changes requested):
I have created a role, that creates a bit meaningful report, I'd recommend using that for further improving the report output. 
https://github.com/ares/foreman_ansible_test

I tried to apply this PR and have few immediate comments for this PR. The 
report

1) The module column now seems to be in the Task name, I'd drop that one completely (sorry I know I told you it may be interesting, but I didn't see it in real world)

2) I'd rename Task name to Task now.

I think we should change the content of the Message column entirely, but that's a different PR

yifatmakias:
this PR depends on another PR:
https://github.com/theforeman/foreman-ansible-modules/pull/1232

xprazak2:
The task messages look nice:

report-output

xprazak2:
@ares, any additional suggestions for improvements?

ares:
This looks awesome! Long term, I think we should move this logic to the callback itself, so we avoid storing huge JSONs in our DB. That causes problems on a big setups. However for now it's good we can tweak the report messages while we still have all data. So :+1: from me, good job.

xprazak2:
Thanks!

PR #413, opened 2021-05-11
Fixes #32041 - set API default params

ezr-ondrej:
Add defaults for format and apiv params. Before this default format was
HTML in API and the version part of URL is required for API urls,
what is mismatch from Foreman core.


xprazak2:
Thanks!

PR #414, opened 2021-05-11
Fixes #25580 - allow import with only import perm

ezr-ondrej:
This fixes an issues where we require users to have `view_smart_proxies`
permission to be able to import from them, but we show the button
anyway. This was confusing users.

This just removes the requirement of having the `view_smart_proxies`
persmission, as user can do nothing else then import AnsibleRoles and variables on
that proxy, so it should not be required to have view permissions for
said proxy.


xprazak2:
Thanks.

PR #415, opened 2021-05-14
Fixes #32600 - Use core components in tests

xprazak2:
Uses find-foreman from foreman-js to set up
correct paths to core components in test.


.github/workflows/js_tests.yml
```diff
@@ -15,12 +15,20 @@ jobs:
         node-version: [10, 12]
```
ezr-ondrej, line 15: maybe bumping the node versions might hepl? we definitelly do not need to test with nodejs 10 and we would be ok testing just on 14

xprazak2:
Ha, fixed. Problems were not caused by node version, but I can leave just v14 in the matrix if you wish.

ezr-ondrej (approved):
Nice :+1: I really love this way, mentioning this on discourse would be appriciated, so other plugins can go this way :)
Works in CI, works for me :)

PR #417, opened 2021-05-26
Fixes #24367 - add support for passphrased keys

ares:
SSH keys protected by the passphrase is not possible to use with pure
Ansible. However ansible-runner provides a way to specify the password
through env/passwords file. For that, the ssh key also needs to be
placed under env/ssh_key (symlink is enough). We already have a
passphrase field in REX Job invocation object, but until now we ignored
that in Ansible jobs.

This patch starts respecting the passphrase for Ansible jobs too. First
it parses the passphrase from the action input. During the
ansible-runner directory preparation, it also creates the env directory.
It symlinks the foreman-proxy key in there and if a passphrase was
specified for the Job invocation, it saves it into the env/passwords
file.

Ansible-runner has a bug that we can't easily specify the env/passwords
entry for the ssh-key. The regular expression is only matched with last
100 characters of the line. We can use '^Enter' in the regexp for that
reason. Also the prompt typically contain the full path to the private
key, which is always dynamic. Therefore the regular expression only
contains words artifacts and ssh_key_data: See
https://github.com/ansible/ansible-runner/issues/533 for more details.
The functionality is documented at

* https://ansible-runner.readthedocs.io/en/stable/intro.html#env-ssh-key
* https://ansible-runner.readthedocs.io/en/stable/intro.html#env-passwords

### DIFF
--- a/lib/foreman_ansible_core/runner/ansible_runner.rb
+++ b/lib/foreman_ansible_core/runner/ansible_runner.rb
@@ -1,4 +1,5 @@
 require 'shellwords'
+require 'yaml'
 
 module ForemanAnsibleCore
   module Runner
@@ -14,12 +15,14 @@ module ForemanAnsibleCore
         @verbosity_level = action_input[:verbosity_level]
         @rex_command = action_input[:remote_execution_command]
         @check_mode = action_input[:check_mode]
+        @passphrase = action_input['secrets']['key_passphrase']
       end
 
       def start
         prepare_directory_structure
         write_inventory
         write_playbook
+        write_ssh_key if !@passphrase.nil? && !@passphrase.empty?
         start_ansible_runner
       end
 
@@ -106,6 +109,15 @@ module ForemanAnsibleCore
         File.write(File.join(@root, 'project', 'playbook.yml'), @playbook)
       end
 
+      def write_ssh_key
+        key_path = File.join(@root, 'env', 'ssh_key')
+        File.symlink(ForemanRemoteExecutionCore.settings[:ssh_identity_key_file], key_path)
+
+        passwords_path = File.join(@root, 'env', 'passwords')
+        secrets = YAML.dump({ "for.*/artifacts/.*/ssh_key_data:" => @passphrase })
+        File.write(passwords_path, secrets, perm: 0o600)
+      end
+
       def start_ansible_runner
         env = {}
         env['FOREMAN_CALLBACK_DISABLE'] = '1' if @rex_command
@@ -129,7 +141,7 @@ module ForemanAnsibleCore
       end
 
       def prepare_directory_structure
-        inner = %w[inventory project].map { |part| File.join(@root, part) }
+        inner = %w[inventory project env].map { |part| File.join(@root, part) }
         ([@root] + inner).each do |path|
           FileUtils.mkdir_p path
         end


ares:
the rubocop tells me to use `present?` but this code is running in smart-proxy without rails, so I'm ignoring

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -106,6 +108,21 @@ def write_playbook
...
+        key_path = File.join(@root, 'env', 'ssh_key')
+        File.symlink(ForemanRemoteExecutionCore.settings[:ssh_identity_key_file], key_path)
+
+        if !@passphrase.nil? && !@passphrase.empty?
+          passwords_path = File.join(@root, 'env', 'passwords')
+          secrets = <<~SECRETS
+            ---
+            "for.*/artifacts/.*/ssh_key_data:": "#{@passphrase}"
+          SECRETS
+          File.write(passwords_path, secrets)
```
adamruzicka, line 121: Could we pass the permissions here so there's not even a slightest amount of time during which the file is world readable?

ares: In fact this may not even be necessary, the `/tmp/d2154..../` is accesible only for the owner, I just wanted to make sure auditors won't report this as an issue. But yeah, I can add the permissions too.

ares: done

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -106,6 +108,22 @@ def write_playbook
...
+
+        if !@passphrase.nil? && !@passphrase.empty?
+          passwords_path = File.join(@root, 'env', 'passwords')
+          secrets = <<~SECRETS
+            ---
+            "for.*/artifacts/.*/ssh_key_data:": "#{@passphrase}"
+          SECRETS
+          File.write(passwords_path, '')
+          File.chmod(0o0600, passwords_path)
+          File.write(passwords_path, secrets)
```
adamruzicka, line 123: I meant something along the lines of `File.write(passwords_path, secrets, perm: 0o600)` but this should work as well

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -106,6 +108,22 @@ def write_playbook
         File.write(File.join(@root, 'project', 'playbook.yml'), @playbook)
       end
 
+      def write_ssh_key
+        key_path = File.join(@root, 'env', 'ssh_key')
+        File.symlink(ForemanRemoteExecutionCore.settings[:ssh_identity_key_file], key_path)
```
adamruzicka, line 113: Key file can be specified on a per-host basis.

If I'm reading https://ansible-runner.readthedocs.io/en/stable/intro.html#env-ssh-key right, then ansible-runner doesn't support providing more keys through the env directory and putting a key there overrides the per-host settings. That would also mean this would break the different keys for different hosts use case completely, no matter if a passphrase is used or not.

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -14,12 +14,14 @@ def initialize(input, suspended_action:)
         @verbosity_level = action_input[:verbosity_level]
         @rex_command = action_input[:remote_execution_command]
         @check_mode = action_input[:check_mode]
+        @passphrase = action_input['secrets']['key_passphrase']
```
adamruzicka, line 17: Passphrase can be provided on a per-host basis

lib/foreman_ansible_core/runner/ansible_runner.rb
```diff
@@ -106,6 +108,22 @@ def write_playbook
...
 
+      def write_ssh_key
+        key_path = File.join(@root, 'env', 'ssh_key')
+        File.symlink(ForemanRemoteExecutionCore.settings[:ssh_identity_key_file], key_path)
+
+        if !@passphrase.nil? && !@passphrase.empty?
+          passwords_path = File.join(@root, 'env', 'passwords')
+          secrets = <<~SECRETS
+            ---
+            "for.*/artifacts/.*/ssh_key_data:": "#{@passphrase}"
```
adamruzicka, line 119: Is this supposed to be yaml? Shouldn't we build a hash and `YAML.dump` it to avoid escaping issues?

lib/foreman_ansible_core/runner/ansible_runner.rb
evgeni, line 123: You need to *create* the file with the right permissions, as otherwise an attacker just needs to wait for your create to happen, acquire a file handle (at this moment the permissions might be OK), and wait for you to write data to it.
The chmod won't have any influence on the already open handle.

ares:
As discussed online, we'll support only single passphrase for a single key, the whole logic gets activated only if the passphrase is specified.

With the file permissions you're of course right. It's kind of academic, because it's all in the private tmp directory. But I improved the code to be more self-explanatory.

evgeni:
Sure, it's a bit academic, but if you're already setting perms, we can do it right ;)

ares:
closing in favor of https://github.com/theforeman/smart_proxy_ansible/pull/35, the code has been migrated to other repo meanwhile

ares:
the updated patch is at https://github.com/ares/foreman_ansible/commit/6408a4072ebc7ab845cb16e3cee8c37af66acdc1 if anyone needs to apply it to the older foreman_ansible (prior core extraction)

PR #418, opened 2021-06-03
Fixes #32709 - Drop all code related to ansible-playbook

adamruzicka:


ares:
:+1: I think this can now safely be dropped, however this now needs rebase after the core extraction

adamruzicka:
Sadly it cannot, not yet. If a user disables batch triggering (or it automatically fallbacks to it), then `ansible-playbook` is used on the proxy side. We'll need to make it use runner even without batch triggering.

ezr-ondrej:
Can we proceed with this now that we have batch processing for all?

adamruzicka:
Since last time, all the code moved to smart_proxy_ansible, so I opened theforeman/smart_proxy_ansible#56. Additionally I needed to do some minor changes to make everything work with batch triggering turned off theforeman/foreman-tasks#666. Closing in favor of the smart_proxy_ansible PR

PR #420, opened 2021-06-10
Fixes #32608 - Add variable overrides to host detail

xprazak2:
We might need to think about a best way how to display composite values such as yaml or json in the future. Just showing the raw value for now.

vars


webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,47 @@
...
+  let { element, meta } = variable.currentValue;
+
+  const knownMatchResources = ['os', 'hostgroup', 'domain'];
+
+  if (!knownMatchResources.includes(element)) {
+    return value;
+  }
+
+  if (element === 'os') {
+    element = 'operatingsystem';
```
amirfefer, line 19: does it make sense to store the data in the redux's store?

amirfefer (changes requested):
Thanks @xprazak2 👍 
I left a few inline comments.

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -1,33 +1,38 @@
 import React, { useState } from 'react';
-import { Tabs, Tab, TabTitleText, Label } from '@patternfly/react-core';
-import { InfoCircleIcon } from '@patternfly/react-icons';
+import PropTypes from 'prop-types';
+import { Tabs, Tab, TabTitleText } from '@patternfly/react-core';
 
+import Loading from 'foremanReact/components/Loading';
```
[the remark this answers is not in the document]
xprazak2: I defer to @MariSvirik for a final decision...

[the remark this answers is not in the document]
MariSvirik: @xprazak2 as we know how the populated page will look like, we should use skeleton. https://www.patternfly.org/v4/components/skeleton/design-guidelines

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
xprazak2: Doing that would not bring any benefits.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.js
```diff
@@ -0,0 +1,48 @@
...
+    variables: { hostId: id, id: encodeId('Host', id) },
+  });
+
+  if (loading) {
+    return <Loading />;
+  }
+
+  if (error) {
+    return <div>{error.message}</div>;
+  }
```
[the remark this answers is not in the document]
xprazak2: Yes, and it counts on specific constants to be present, which is not the case here.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const variables = data.host.allAnsibleRoles.nodes.reduce((memo, role) => {
+    const vars = role.ansibleVariablesWithOverrides.nodes.map(variable => ({
+      ...variable,
+      roleName: role.name,
+    }));
+    return memo.concat(vars);
+  }, []);
```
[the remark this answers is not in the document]
xprazak2: Apollo takes care of data, so no action required here. I extracted the function out of the component.

package.json
```diff
@@ -13,7 +13,9 @@
     "react-json-tree": "^0.11.0"
   },
   "devDependencies": {
+    "@apollo/react-testing": "^4.0.0",
     "@babel/core": "^7.7.0",
+    "@testing-library/react": "^11.2.5",
```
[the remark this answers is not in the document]
xprazak2: Yes, removed.

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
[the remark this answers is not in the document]
xprazak2: Thanks, I replaced loader with skeletons.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTable.js
```diff
@@ -0,0 +1,56 @@
...
+    <TableComposable variant="compact">
+      <Thead>
+        <Tr>
+          {columns.map((col, idx) => (
+            <Th key={idx}>{col}</Th>
+          ))}
+        </Tr>
+      </Thead>
+      <Tbody>
+        {variables.map((variable, idx) => (
```
amirfefer, line 37: There are two components at least with the same key due to using `index` as a key.
[react docs](https://reactjs.org/docs/lists-and-keys.html) doesn't recommend using `index` as a key because it does not represent the data correctly and therefore can break dynamic data.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTable.js
```diff
@@ -0,0 +1,56 @@
...
+          <Tr key={idx}>
+            <Td>{variable.key}</Td>
```
amirfefer, line 39: looks like a good candidate for the component key

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
amirfefer: I mean, the data comes from the server, like we have the API middleware and selectors for fetching, storing, cleaning, formatting, and manipulate data. I wrote a suggestion in another comment.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const elementId = hostAttrs[`${element}_id`];
+
+  if (elementId && meta?.canEdit) {
+    return <Link to={`/${element}s/${elementId}/edit`}>{value}</Link>;
```
amirfefer, line 25: if the user can't edit it, we can make it disabled rather than removing it?
@MariSvirik any thoughts?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,46 @@
...
+  const value = variable.currentValue?.value || variable.defaultValue;
+
+  if (variable.parameterType === 'boolean') {
+    return value ? <CheckIcon /> : <TimesIcon />;
+  }
+  const type = variable.parameterType;
+  if (type === 'yaml' || type === 'hash' || type === 'array') {
+    return JSON.stringify(value);
+  }
+  return value;
```
amirfefer, line 45: switch case? to me it more readable and easy to extend / change

```js
switch (paramterType)
{
    case "boolean":
     return  value ? <CheckIcon /> : <TimesIcon />;
    case "yaml":
    case "hash":
    case "array":
      return JSON.stringify(value);
    default:
       return value
}
```

amirfefer (changes requested):
Thanks @xprazak2 !
I left a few inline comments

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
xprazak2: I return value without link 3 lines below.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesHelper.js
```diff
@@ -0,0 +1,8 @@
+export const reorderVariables = variables =>
+  variables.reduce((memo, role) => {
+    const vars = role.ansibleVariablesWithOverrides.nodes.map(variable => ({
```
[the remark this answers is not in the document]
xprazak2: Changed, though this is the data we requested in query so if there is no `role.ansibleVariablesWithOverrides.nodes`, it means we do not wait for request to resolve or we have not handled error response properly.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.test.js
```diff
@@ -0,0 +1,163 @@
...
+  it('should load', async () => {
+    const { container } = render(
+      <MemoryRouter>
+        <TestComponent status="RESOLVED" id={hostId} hostAttrs={hostAttrs} />
+      </MemoryRouter>
+    );
+    expect(
+      container.getElementsByClassName('react-loading-skeleton')
+    ).toHaveLength(5);
+    await waitFor(tick);
```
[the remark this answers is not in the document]
xprazak2: It does not seem to work for me.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.js
```diff
@@ -0,0 +1,41 @@
...
+import AnsibleVariableOverridesTable from './AnsibleVariableOverridesTable';
+
+import { encodeId } from '../../../../globalIdHelper';
+import { reorderVariables } from './AnsibleVariableOverridesHelper';
+import './AnsibleVariableOverrides.scss';
+
+const AnsibleVariableOverrides = ({ id, hostAttrs }) => {
+  const { loading, data, error } = useQuery(variableOverrides, {
+    variables: { hostId: id, id: encodeId('Host', id) },
+  });
```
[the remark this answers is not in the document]
xprazak2: With the hook handling the mapping to status constants, the component looks like the following:

```js
const AnsibleVariableOverrides = ({ hostAttrs }) => {
  const { id } = hostAttrs;

  const { status, data, error } = useQueryWithStatus({ hostId: id, id: encodeId('Host', id) });

  const errorNode = (<div>{error ? error.message : ''}</div>)

  return (
    <SkeletonLoader status={status} errorNode={errorNode}>
      <div className="ansible-tab-margin">
        <AnsibleVariableOverridesTable
          variables={reorderVariables(data ? data.host.allAnsibleRoles.nodes : [])}
          hostAttrs={hostAttrs}
        />
      </div>
    </SkeletonLoader>
  );
```
So I still need to check for errors and data presence in component, but now I use additional hook wrapper and status mapping - I added more things only to get what I already had. I kept it as was for now but I am willing to change if you really think the version with wrapper is better. 

> Also, with this suggested custom hook, the data can cam be stored in foreman's redux store (like useAPI), so data manipulation can be done via selectors and memorized by reselect if needed

Yes, and it would mean having the same data in 2 places - redux store and apollo cache. That would effectively neutralize some of apollo's strong features, such as [auto-updates](https://www.apollographql.com/docs/react/data/mutations/#updating-a-single-existing-entity) after mutation, because we would need to manually update the store to match the data from apollo.

I am strongly against putting any data fetched by apollo into redux store.

package.json
```diff
@@ -13,6 +13,7 @@
     "react-json-tree": "^0.11.0"
   },
   "devDependencies": {
+    "@apollo/react-testing": "^4.0.0",
```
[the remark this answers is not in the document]
xprazak2: Not yet: https://github.com/theforeman/foreman-js/pull/320

xprazak2:
Thank you for your review @amirfefer, updated.

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -18,18 +20,19 @@ const AnsibleHostDetail = props => {
...
           <div className="ansible-host-detail">
-            <Label
-              color="blue"
-              icon={<InfoCircleIcon />}
-              style={{ marginTop: '1.5rem' }}
-            >
-              Ansible Variables coming soon!
-            </Label>
+            <SkeletonLoader status={status} skeletonProps={{ count: 5 }}>
+              <AnsibleVariableOverrides hostAttrs={response} />
```
tbrisker, line 24: This won't show the loading state as there are children

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
+import CheckIcon from '@patternfly/react-icons/dist/js/icons/check-icon';
```
tbrisker, line 3: These imports won't work at build time, you need to import them from @paternfly/react-icons

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTable.js
```diff
@@ -0,0 +1,56 @@
...
+
+import {
+  formatValue,
+  formatSourceAttr,
+} from './AnsibleVariableOverridesTableHelper';
+
+const AnsibleVariableOverridesTable = ({ variables, hostAttrs }) => {
+  const columns = [
+    __('Name'),
+    __('Ansible Role'),
```
tbrisker, line 21: Should the role be the first column and group the variables from the same role together?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,48 @@
...
+import { Link } from 'react-router-dom';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const formatSourceLink = (hostAttrs, { currentValue }) => {
+  const value = `${currentValue.element}: ${currentValue.elementName}`;
+  let { element } = currentValue;
+  const { meta } = currentValue;
+
+  const knownMatchResources = ['os', 'hostgroup', 'domain'];
```
tbrisker, line 12: Why only these three and not other resources?

app/graphql/overriden_ansible_variable_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+    result = @override_resolver.resolve @ansible_variable
+    add_meta result
+  end
+
+  private
+
+  def known_type_class(element)
+    return unless element.is_a? String
+    return Operatingsystem if element == 'os'
+    element.classify.constantize rescue nil
```
tbrisker, line 23:
```suggestion
    element.classify.safe_constantize
```
Would also prevent the need to disable the rubocop

app/graphql/types/ansible_variable_override.rb
```diff
@@ -0,0 +1,10 @@
+module Types
+  class AnsibleVariableOverride < GraphQL::Types::Relay::BaseObject
+    description 'Override value for Ansible Variable'
+
+    field :value, ::Types::RawJson, :null => false
+    field :element, ::Types::RawJson, :null => false
+    field :element_name, ::Types::RawJson, :null => false
```
tbrisker, line 7: These names are a bit confusing, should we rename them to something more descriptive such as `override_source` or `overriding_resource`?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import TimesIcon from '@patternfly/react-icons/dist/js/icons/times-icon';
+import CheckIcon from '@patternfly/react-icons/dist/js/icons/check-icon';
+import { Link } from 'react-router-dom';
+import { translate as __ } from 'foremanReact/common/I18n';
+
+const formatSourceLink = (hostAttrs, { currentValue }) => {
+  const value = `${currentValue.element}: ${currentValue.elementName}`;
```
tbrisker, line 8: Should the element be translated?

tbrisker (changes requested):
Added my still relevant comments from reviewing the other PR, didn't give this PR a full review though.

app/graphql/types/ansible_variable_override.rb
xprazak2: These come directly from classification, are we sure we want to diverge on naming?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
xprazak2: These are the values offered in 'Prioritize attribute order' when editing. There are too many special cases that cannot be reliably handled in a generic way

1. Since matchers can be set up on virtually anything, there is no way to tell if we match on host attribute (such as comment) or on associated resource.
2. Some resources may may have naming mismatches similar to operating system
3. Katello has a very different page paths than Foreman and there is no way to tell where the resource comes from
4. Some resources extend host via facets

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTable.js
xprazak2: No, we do not go for the same design as what we have in host form. The variables from the same role are grouped by default, I imagine we could have sortable header for role in the future.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.js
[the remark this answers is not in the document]
amirfefer: >  I am strongly against putting any data fetched by apollo into redux store.

I read about this a bit further  and I agree with you, we should use apollo as a [redux replacement](https://www.apollographql.com/blog/apollo-client/caching/dispatch-this-using-apollo-client-3-as-a-state-management-solution/)

xprazak2:
@amirfefer, @tbrisker could you take another look please?

lib/foreman_ansible/register.rb
```diff
@@ -98,6 +98,10 @@
 
   register_global_js_file 'global'
 
+  extend_graphql_type :type => ::Types::Host do
+    field :all_ansible_roles, ::Types::AnsibleRole.connection_type, :null => true
```
tbrisker, line 102: This seems to be causing the tests to fail with
```
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "ansible_roles" does not exist
LINE 8:  WHERE a.attrelid = '"ansible_roles"'::regclass
```
I guess it runs before the db is migrated so the table doesn't yet exit?

jest.config.js
```diff
@@ -6,6 +6,9 @@ const foremanReactRelative = 'webpack/assets/javascripts/react_app';
 const foremanFull = foremanLocation();
 const foremanReactFull = foremanRelativePath(foremanReactRelative);
 
+// Makes graphql files work in test
+tfmConfig.transform["\\.(gql|graphql)$"] = "jest-transform-graphql";
```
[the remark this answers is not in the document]
tbrisker: This can be dropped now that it's in `@theforeman/test`

package.json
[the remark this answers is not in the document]
tbrisker: this too

package.json
```diff
@@ -22,6 +23,7 @@
     "@theforeman/vendor-dev": "^8.4.1",
     "babel-eslint": "^10.0.3",
     "eslint": "^6.7.2",
+    "jest-transform-graphql": "^2.1.0",
```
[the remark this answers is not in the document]
tbrisker: This as well, minimum version of `@theforeman/test` should be updated to 8.9.0

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -1,35 +1,38 @@
...
-        <div className="host-details-tab-item">
-          <div className="ansible-host-detail">
-            <Label
-              color="blue"
-              icon={<InfoCircleIcon />}
-              style={{ marginTop: '1.5rem' }}
-            >
-              Ansible Variables coming soon!
-            </Label>
+    <SkeletonLoader status={status} skeletonProps={{ count: 5 }}>
```
tbrisker, line 16: I think the skeleton loader isn't needed here, we can always show the tab itself, and the AnsibleVariableOverrides component already has a skeleton while it's loading

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -1,35 +1,38 @@
...
+      <Tabs activeKey={activeTab} isSecondary>
+        <Tab
+          eventKey="variables"
+          title={<TabTitleText>{__('Variables')}</TabTitleText>}
+        >
+          <div className="host-details-tab-item">
+            <div className="ansible-host-detail">
```
tbrisker, line 23: do we need two separate divs for these classes? could we combine them in one or even put the classnames on one of the other components (`Tab` or `AnsibleVariableOverrides`)?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverrides.js
```diff
@@ -0,0 +1,42 @@
...
+  if (loading) {
+    return <Skeleton count={5} />;
+  }
+
+  if (error) {
+    return <div>{error.message}</div>;
+  }
+
+  return (
+    <div className="ansible-tab-margin">
```
tbrisker, line 29: Here as well, could this be className attribute of the AnsibleVariableOverridesTable?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesHelper.js
```diff
@@ -0,0 +1,8 @@
+export const reorderVariables = variables =>
+  variables.reduce((memo, role) => {
+    const vars = role?.ansibleVariablesWithOverrides?.nodes.map(variable => ({
+      ...variable,
+      roleName: role.name,
+    }));
+    return memo.concat(vars);
+  }, []);
```
tbrisker, line 8: Can we get the data from the server in the format we want it directly and avoid this complex mutation? IIUC all this changes is adding a roleName field on every variable, but I may be misunderstanding?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
tbrisker: > These are the values offered in 'Prioritize attribute order' when editing.

If I'm not mistaken, the default order is defined by a setting, and every role can have it's own override order. 
I agree matchers could be set on anything and it's hard to guess the correct path - perhaps as a first step we could drop the attempt to link to the resource and only display the type and name? Then we also don't need to check the `canEdit` permission

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './AnsibleVariableOverrides';
```
tbrisker, line 1: I think in https://community.theforeman.org/t/rethinking-react-redux-folder-structure/24183 we agreed that the main component should be directly in the index.js file?

app/graphql/types/ansible_variable.rb
```diff
@@ -0,0 +1,18 @@
...
+    field :override, Boolean
+    field :description, String
+    field :hidden_value, Boolean
+    field :parameter_type, String
+    field :omit, Boolean
+    field :required, Boolean
+    field :validator_type, String
+    field :validator_rule, String
+    field :default_value, ::Types::RawJson
+  end
```
tbrisker, line 17: if we add 
```
field :ansible_role_name, String
```
I think we can avoid the complex mutation of the response

app/graphql/overriden_ansible_variable_presenter.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def add_meta(result)
+    return unless result
+    type = known_type_class(result[:element])
+    return result unless type
+    attribute = type.attribute_names.include?('title') ? :title : :name
+    object = type.find_by attribute => result[:element_name]
+    return result unless object
+    result[:meta] = {
+      :can_edit => User.current.can?(object.permission_name(:edit), object)
```
tbrisker, line 34: This can get quite slow for non-admin users if there are a lot of overrides

tbrisker (changes requested):
Thanks @xprazak2, updated inline

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
xprazak2: I am not a css guru, but I do not think we can combine them since each sets a different bg color. I tried to pull on the inner one into the AnsibleVariableOverrides component, but that results in light gray skeleton on slightly darker gray background which does not look good.

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
xprazak2: The page makes request for the host attributes and the skeleton loader here waits for it to resolve so that it has host id available. The loader in AnsibleVariableOverrides waits for the variable values to be fetched.

xprazak2:
@tbrisker, @amirfefer could you take another look please?

tbrisker (changes requested):
Looking much cleaner now! just a few small comments inline

webpack/graphql/queries/variableOverrides.gql
```diff
@@ -0,0 +1,21 @@
+query($id: String!, $hostId: Int!) {
+  host(id: $id) {
+    allAnsibleRoles {
+      nodes {
```
tbrisker, line 4: I might be missing some graphql background here, why do we need the extra `nodes` layer here and below?

tbrisker (commented):
Looks good @xprazak2 ! only a couple of questions inline, and some linting issues that need resolving.

amirfefer (changes requested):
Thanks @xprazak2 ! 
I left a few small comments, plus js tests fail

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <AnsibleVariableOverridesTable
+      variables={reorderVariables(data.host.allAnsibleRoles.nodes)}
+    />
+  );
+};
+
+AnsibleVariableOverrides.propTypes = {
+  hostAttrs: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
xprazak2: True and we [wait for it](https://github.com/theforeman/foreman_ansible/blob/master/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js#L18) to be present before the component is rendered.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+import { encodeId } from '../../../../globalIdHelper';
+import { reorderVariables } from './AnsibleVariableOverridesHelper';
+import './AnsibleVariableOverrides.scss';
+
+const AnsibleVariableOverrides = ({ hostAttrs }) => {
+  const { id } = hostAttrs;
+
+  const { loading, data, error } = useQuery(variableOverrides, {
+    variables: { hostId: id, id: encodeId('Host', id) },
```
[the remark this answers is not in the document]
xprazak2: It might be used instead of `hostId`, but real id is still needed to construct a global id, so the whole component is rendered only after the id is available.

amirfefer (approved):
Thanks @xprazak2 !
LGTM

tbrisker:
[test foreman_ansible]

webpack/graphql/queries/variableOverrides.gql
xprazak2: It is because of Relay style connections that add additional information such as total count of the nodes.
```
allAnsibleRoles {
  totalCount
  pageInfo {
     hasNextPage
  }
  nodes {
   //...
  }
}
```

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,20 @@
+module ForemanAnsible
+  # Service which resolves override values for hosts
+  class OverrideResolver
+    attr_reader :overrides, :ansible_variables
+
+    def initialize(host)
+      if host
```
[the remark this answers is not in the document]
xprazak2: It is probably not an issue in a context of the host detail page, but the method relies on host not being nil.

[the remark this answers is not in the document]
tbrisker: In case this is ever called with a `nil` host (e.g. in case we reuse it somewhere else), should we silently provide an empty hash or would it be better to throw an error?

webpack/graphql/queries/variableOverrides.gql
tbrisker: Do we need this additional information here? if I'm not mistaken, all variables are loaded without pagination and we don't use the count anywhere.

tbrisker: never mind, I played with it a bit and now I understand why this is needed.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -0,0 +1,47 @@
...
+
+export const formatSourceAttr = (hostAttrs, variable) =>
+  variable.currentValue
+    ? formatSourceLink(hostAttrs, variable)
+    : __('Default value');
+
+export const formatValue = variable => {
+  const value = variable.currentValue
+    ? variable.currentValue.value
+    : variable.defaultValue;
```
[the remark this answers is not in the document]
tbrisker: I think that was the implementation in a previous iteration, but that won't work correctly for boolean values

webpack/globalIdHelper.js
```diff
@@ -0,0 +1,11 @@
...
+const versionSeparator = ':';
+const defaultVersion = '01';
+
+export const decodeId = model => {
+  const split = atob(model.id).split(idSeparator);
+  return parseInt(split[split.length - 1], 10);
+};
+
+export const encodeId = (typename, id) =>
+  btoa([defaultVersion, versionSeparator, typename, idSeparator, id].join(''));
```
tbrisker, line 11: Not crucial but I suppose that this should be provided by core rather than every plugin creating its own.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesHelper.js
```diff
@@ -0,0 +1,8 @@
+export const reorderVariables = variables =>
+  variables.reduce((memo, role) => {
+    const vars = role?.ansibleVariablesWithOverrides?.nodes;
+    if (vars) {
+      return memo.concat(vars);
+    }
+    return memo;
+  }, []);
```
tbrisker, line 8: We could use `flatMap` here which doesn't create an intermediate array for each iteration (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#alternative) and is more readable:
```suggestion
roles.flatMap((role)=>(role.ansibleVariablesWithOverrides.nodes))
```
If I'm not mistaken, the conditional chaining operand is not needed since GraphQL guarantees that the response will always include these fields even if they are empty.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesHelper.js
```diff
@@ -0,0 +1,8 @@
+export const reorderVariables = variables =>
```
tbrisker, line 1: The naming here is confusing IMHO, it took me a while to understand what this function actually does. 
We aren't reordering the variables, we are flattening the response, and the argument isn't the variables, it's the roles.
How about something like:
```suggestion
export const extractVariables = roles =>
```

tbrisker (changes requested):

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesHelper.js
xprazak2: Updated, conditional chaining is not needed as mentioned in [previous comment](https://github.com/theforeman/foreman_ansible/pull/420#discussion_r665386867)

webpack/globalIdHelper.js
xprazak2: Yes, this should move to core at some point.

tbrisker (approved):
Thanks @xprazak2 ! Great job pushing this through 🏅

app/services/foreman_ansible/override_resolver.rb
```diff
@@ -0,0 +1,18 @@
+module ForemanAnsible
+  # Service which resolves override values for hosts
+  class OverrideResolver
+    attr_reader :overrides, :ansible_variables
+
+    def initialize(host)
+      raise(Foreman::Exception, 'OverriderResolver needs a host to resolve overrides') unless host
```
tbrisker, line 7: rubocop isn't happy, i think this should be
```suggestion
      raise(Foreman::Exception.new('OverriderResolver needs a host to resolve overrides')) unless host
```

tbrisker (approved):
Thanks @xprazak2 !

PR #422, opened 2021-06-22
Fixes #32203 - Add timeout param to ansible provider options

yifatmakias:
related to this PR:
https://github.com/theforeman/smart_proxy_ansible/pull/41

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -32,7 +32,8 @@ if defined? ForemanRemoteExecution
               template_invocation.template
             ),
             :name => host.name,
-            :check_mode => host.host_param('ansible_roles_check_mode')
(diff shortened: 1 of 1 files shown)


app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -32,7 +32,8 @@ def proxy_command_options(template_invocation, host)
               template_invocation.template
             ),
             :name => host.name,
-            :check_mode => host.host_param('ansible_roles_check_mode')
+            :check_mode => host.host_param('ansible_roles_check_mode'),
+            :job_invocation_timeout => template_invocation.job_invocation[:execution_timeout_interval]
```
ezr-ondrej, line 36: why do we need this, it seems to be sent as `execution_timeout_interval` already, or is it not?

yifatmakias:
This PR is no longer needed.

PR #424, opened 2021-06-30
Fixes #32913 - Show correct roles order for hostgroup

xprazak2:


xprazak2:
Test failure unrelated, I have seen that on other PRs as well. Do we know if there has been any change in core that may be causing this?

ezr-ondrej:
I've got no idea, but this is not the cause so I don't mind merging with this failure, we can investigate in separate PR.
Could we add a test for this tho? Seems like something we would easily forgot why it's there if we refactor in future.

ezr-ondrej (approved):
Thanks @xprazak2 ! :+1:

PR #426, opened 2021-07-01
Fixes #32920 - support conversion of offline systems

ares:
Convert2rhel version 0.22 added a parameter to avoid installing
subscription-manager from online resources. In our case we should use
subscription-manager installed during the registration process, that way
we can perform the upgrade with repositories synced to local
repositories. So we enforce that by --keep-rhsm parameter.

Also the log is always stored at /var/log/convert2rhel/convert2rhel.log
so we don't need the redirection (which doesn't work anyway)


xprazak2:
Thanks!

ezr-ondrej:
I've cherry-picked this manually @ares.
6.3 stable CP: bd6751333cbc901e1c1cbeb565518c2c3698c00a

PR #427, opened 2021-07-06
Fixes #32764 - Add visualization of host inventory data

yifatmakias:
This PR is completely depending on @xprazak2 PR:
https://github.com/theforeman/foreman_ansible/pull/420

and also on this PR:
https://github.com/theforeman/foreman/pull/8495

after they will be merged I will rebase this one.

### DIFF
--- a/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
+++ b/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
@@ -1,35 +1,53 @@
 import React, { useState } from 'react';
+import PropTypes from 'prop-types';
 import { Tabs, Tab, TabTitleText, Label } from '@patternfly/react-core';
 import { InfoCircleIcon } from '@patternfly/react-icons';
-
+import SkeletonLoader from 'foremanReact/components/common/SkeletonLoader';
 import { translate as __ } from 'foremanReact/common/I18n';
 
 import './AnsibleHostDetail.scss';
+import WrappedAnsibleHostInventory from './components/AnsibleHostInventory';
 
-const AnsibleHostDetail = props => {
+const AnsibleHostDetail = ({ response, status }) => {
   // https://projects.theforeman.org/issues/32398
-  const [activeTab] = useState('variables');
+  const [activeTab, setActiveTab] = useState('variables');
+  const handleTabClick = (event, tab) => setActiveTab(tab);
 
   return (
-    <Tabs activeKey={activeTab} isSecondary>
-      <Tab
-        eventKey="variables"
-        title={<TabTitleText>{__('Variables')}</TabTitleText>}
-      >
-        <div className="host-details-tab-item">
-          <div className="ansible-host-detail">
-            <Label
-              color="blue"
-              icon={<InfoCircleIcon />}
-              style={{ marginTop: '1.5rem' }}
-            >
-              Ansible Variables coming soon!
-            </Label>
-          </div>
-        </div>
-      </Tab>
-    </Tabs>
+    <SkeletonLoader status={status} skeletonProps={{ count: 5 }}>
+      {response && (
+        <Tabs onSelect={handleTabClick} activeKey={activeTab} isSecondary>
+          <Tab
+            eventKey="variables"
+            title={<TabTitleText>{__('Variables')}</TabTitleText>}
+          >
+            <div className="host-details-tab-item">
+              <div className="ansible-host-detail">
+                <Label
+                  color="blue"
+                  icon={<InfoCircleIcon />}
+                  style={{ marginTop: '1.5rem' }}
+                >
+                  Ansible Variables coming soon!
+                </Label>
+              </div>
+            </div>
+          </Tab>
+          <Tab
+            eventKey="inventory"
+            title={<TabTitleText>{__('Inventory')}</TabTitleText>}
+          >
+            <WrappedAnsibleHostInventory hostId={response.id} />
+          </Tab>
+        </Tabs>
+      )}
+    </SkeletonLoader>
   );
+};
+
+AnsibleHostDetail.propTypes = {
+  response: PropTypes.object.isRequired,
+  status: PropTypes.string.isRequired,
 };
 
 export default AnsibleHostDetail;
--- a/webpack/components/AnsibleHostDetail/AnsibleHostDetail.test.js
+++ b/webpack/components/AnsibleHostDetail/AnsibleHostDetail.test.js
@@ -1,14 +1,15 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import '@testing-library/jest-dom';
-
 import AnsibleHostDetail from './';
 
 describe('AnsibleHostDetail', () => {
-  it('should show content', () => {
-    render(<AnsibleHostDetail />);
+  it('should show skeleton when loading', () => {
+    const { container } = render(
+      <AnsibleHostDetail status="PENDING" response={{ id: 5 }} />
+    );
     expect(
-      screen.getByText('Ansible Variables coming soon!')
-    ).toBeInTheDocument();
+      container.getElementsByClassName('react-loading-skeleton')
+    ).toHaveLength(5);
   });
 });
--- a/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
+++ b/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import Immutable from 'seamless-immutable';
+import PropTypes from 'prop-types';
+import './AnsibleHostInventory.scss';
+import ReportJsonViewer from '../../../ReportJsonViewer';
+
+const AnsibleHostInventory = ({ inventoryData }) => {
+  const mutableInventory =
+    inventoryData && Immutable.asMutable(inventoryData, { deep: true });
+
+  return (
+    <div className="ansible-host-inventory">
+      <ReportJsonViewer data={mutableInventory} />
+    </div>
+  );
+};
+
+AnsibleHostInventory.propTypes = {
+  inventoryData: PropTypes.object.isRequired,
+};
+
+export default AnsibleHostInventory;
--- a/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.scss
+++ b/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.scss
@@ -1,0 +1,4 @@
+
+.ansible-host-inventory {
+  padding: 1.8rem;
+}
--- a/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.test.js
+++ b/webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.test.js
@@ -1,0 +1,104 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import AnsibleHostInventory from './AnsibleHostInventory';
+
+const inventoryData = {
+  all: {
+    hosts: ['jim-dorsay.tlv.redhat.com'],
+    vars: {},
+  },
+  _meta: {
+    hostvars: {
+      'jim-dorsay.tlv.redhat.com': {
+        foreman: {
+          hostname: 'jim-dorsay',
+          fqdn: 'jim-dorsay.tlv.redhat.com',
+          hostgroup: 'libvirt',
+          foreman_subnets: [
+            {
+              name: 'subnet',
+              network: '192.168.122.0',
+              mask: '255.255.255.0',
+              gateway: '',
+              dns_primary: '',
+              dns_secondary: '',
+              from: '',
+              to: '',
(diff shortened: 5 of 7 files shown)


webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import AnsibleHostInventory from "./AnsibleHostInventory";
+
+
+describe('AnsibleHostInventory', () => {
+    it('should show content', () => {
+        render(<AnsibleHostInventory />);
```
xprazak2, line 9: `render` only renders the component but does not actually do any asserts. A basic check that the expected text is shown should be enough (following example not tested):

```js
 import { screen } from '@testing-library/react'
import '@testing-library/jest-dom'; // might be needed for toBeInTheDocument method

expect(screen.getByText('_meta')).toBeInTheDocument();
```

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../../../node_modules/@patternfly/patternfly/base/patternfly-variables';
+
+.ansible-host-inventory {
+  background-color: $pf-color-white;
```
xprazak2, line 4: Does the background-color rule have any effects on the tab? I haven't noticed any difference when I removed the rule...

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../../../../node_modules/@patternfly/patternfly/base/patternfly-variables';
```
xprazak2, line 1: I think this will cause problems in packaging. Replacing the relative path with '~@patternfly/' should do the trick.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
```diff
@@ -0,0 +1,25 @@
...
+import { foremanUrl } from 'foremanReact/common/helpers'
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+
+const AnsibleHostInventory = ({ hostId }) => {
+    const params = useMemo(() => ({ params: { host_ids: [hostId] }}), [hostId]);
+
+    const url =  foremanUrl('/ansible/api/ansible_inventories/hosts');
+    const {
+        response: inventory,
+    } = useAPI('get', url, params);
```
xprazak2, line 14: You could move the API call into the wrapper component and pass the inventory data as a prop. It will make the testing more simple as you will not need to mock the API call. More on testing below.

xprazak2:
Works nicely, I added a couple of comments. `npm run lint -- --fix` should auto-fix most of the linting issues.

The first tab level is already in the master branch, so I do not think you need the variables PR. You can remove those commits and you will still be able to add the inventory tab to the page.

Let's not rely on routed tabs to get unblocked, we can use component state to handle tab change for now:

```js
// AnsibleHostDetail

const [activeTab, setActiveTab] = useState('variables')

const handleTabClick = (event, tab) = setActiveTab(tab)
//...
<Tabs onSelect={handleTabClick}>
```

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
```diff
@@ -0,0 +1,20 @@
+import React, { useMemo } from 'react';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import PropTypes from 'prop-types';
+import AnsibleHostInventory from './AnsibleHostInventory';
+
+const WrappedAnsibleHostInventory = ({ hostId }) => {
+  const params = useMemo(() => ({ params: { host_ids: [hostId] } }), [hostId]);
```
xprazak2, line 8: Just out of curiosity, why did you decide to `useMemo`?

xprazak2:
Nice work. I think this might be ready as it is.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
yifatmakias: That was @amirfefer input. I think because we want it to change when the hostId changes but Amir if there is a better explanation let us know :)

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const url = foremanUrl('/ansible/api/ansible_inventories/hosts');
+  const { response: inventory } = useAPI('get', url, params);
+
+  return <AnsibleHostInventory inventoryData={inventory} />;
+};
+
+WrappedAnsibleHostInventory.propTypes = {
+  hostId: PropTypes.number.isRequired,
```
amirfefer, line 17: the host id can be undefined in the first rendering, can you remove the  `isRequired` and add a default?

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Immutable from 'seamless-immutable';
+import PropTypes from 'prop-types';
+import './AnsibleHostInventory.scss';
+import ReportJsonViewer from '../../../ReportJsonViewer';
+
+const AnsibleHostInventory = ({ inventoryData }) => {
+  const mutableInventory =
+    inventoryData && Immutable.asMutable(inventoryData, { deep: true });
```
amirfefer, line 9: this triggered on every render, how about moving this into a selector? 
you don't need to create a new reducer for that, just give a `key` to the `useAPI` and reuse `selectAPIResponse`

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -1,17 +1,24 @@
...
 
-const AnsibleHostDetail = props => {
+const AnsibleHostDetail = ({ response, status }) => {
   // https://projects.theforeman.org/issues/32398
-  const [activeTab] = useState('variables');
+  const [activeTab, setActiveTab] = useState('variables');
+  const handleTabClick = (event, tab) => setActiveTab(tab);
+
+  if (status === 'PENDING') {
+    return <Loading />;
```
amirfefer, line 17: the host details page uses `SkeletonLoader`, how about using it also in ansible?

amirfefer (changes requested):
Thanks @yifatmakias !  Looks great,

I wonder if it makes sense to do a "lazy" loading for tabs content, so the content will be fetched when the specific tab is active. I added some inline comments.

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
yifatmakias: Quick question does the SkeletonLoader should wrap the <Tabs> tag or every <Tab> tag separately?

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
yifatmakias: I am not sure what do you mean here, could you explain again?

webpack/components/AnsibleHostDetail/AnsibleHostDetail.test.js
```diff
@@ -1,14 +1,16 @@
...
+
 describe('AnsibleHostDetail', () => {
-  it('should show content', () => {
-    render(<AnsibleHostDetail />);
-    expect(
-      screen.getByText('Ansible Variables coming soon!')
-    ).toBeInTheDocument();
-  });
+  describe('rendering', () =>
+    testComponentSnapshotsWithFixtures(AnsibleHostDetail, fixtures));
```
xprazak2, line 15: I know the existing test was complaining about component not being wrapped in Provider and the snapshot tests are seductively easy to write using those test helpers, but if you take a look at the generated snapshot, it really does not test much and it is not able to detect what is on the page.

I would stay away from using snapshots and modified the existing test so that is passes, let me know if you need a hint.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
xprazak2: @amirfefer , if id is not available, then the component should wait for it to be present. Supplying a fake hostId value (which should be never used to call API!) seems like an awkward thing to do and it tells me we need to reconsider how we compose things together.

xprazak2:
Looks good. The propType for hostId is failing, it is kind of difficult to spot among the other failures on the page:

prop-fail

yifatmakias:
> Looks good. The propType for hostId is failing, it is kind of difficult to spot among the other failures on the page:
> 
> ![prop-fail](https://user-images.githubusercontent.com/2664090/126112180-d9f85992-7c57-4d9f-abfa-78bfacac5887.png

@xprazak2 I wonder what could be a good solution except removing the required and adding a default id which I understand is wired but what other solution is there? @amirfefer any idea?

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
amirfefer: You can wrap the entire component with one `SkeletonLoader` with multiple skeletons
```js 
<SkeletonLoader skeletonProps={{ count: 10 // 10 skeletons }}>
  <Content />
</SkeletonLoader>1
```

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/AnsibleHostInventory.js
amirfefer: I created a small PR that adds this ability directly to `useAPI` hook
https://github.com/theforeman/foreman/pull/8683

amirfefer:
you can remove the required and add add an empty default value:
```js
Component.propTypes {
  hostId: propTypes.number,
}

Component.defaultProps = {
  hostId: undefined,
}
```
The hostId arrives from the API  response, so it will be undefined at first.
Alternatively we can get the hostId directly from the url, but I'd prefer to make that change from the root component  (HostDetails -> tabs)

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
amirfefer: `undefined` is not a fake id, if the host id is undefined no ajax call will be triggered.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
```diff
@@ -0,0 +1,24 @@
+import React, { useMemo } from 'react';
+import { foremanUrl } from 'foremanReact/common/helpers';
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import PropTypes from 'prop-types';
+import AnsibleHostInventory from './AnsibleHostInventory';
+
+const WrappedAnsibleHostInventory = ({ hostId }) => {
+  const params = useMemo(() => ({ params: { host_ids: [hostId] } }), [hostId]);
+
+  const url = foremanUrl('/ansible/api/ansible_inventories/hosts');
```
amirfefer, line 10: This makes the url undefined if no hostId exists, no fake ajax calls will be made 
```suggestion
  const url = hostId && foremanUrl('/ansible/api/ansible_inventories/hosts');
```

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -1,35 +1,51 @@
...
-          <div className="ansible-host-detail">
-            <Label
-              color="blue"
-              icon={<InfoCircleIcon />}
-              style={{ marginTop: '1.5rem' }}
-            >
-              Ansible Variables coming soon!
-            </Label>
+    <SkeletonLoader status={status} skeletonProps={{ count: 5 }}>
+      <Tabs onSelect={handleTabClick} activeKey={activeTab} isSecondary>
```
amirfefer, line 18: If the Skeleton's children won't be undefined, an empty state won't be displayed, this fixes it:
```js
<SkeletonLoader>
  {response && (<Tabs>...
```

amirfefer (changes requested):
Thanks @yifatmakias , I have added some inline comments

xprazak2 (approved):

xprazak2:
Any additional comments @amirfefer ?

amirfefer (approved):
Looking good 👍 
Thanks @yifatmakias , @xprazak2 !

xprazak2:
Thanks everyone!

ares:
@yifatmakias @xprazak2 was this ever demoed? looks like a great candidate :-)

yifatmakias:
> @yifatmakias @xprazak2 was this ever demoed? looks like a great candidate :-)

Not yet, I will make sure it is demoed in the next Demo

PR #408, opened 2021-05-03
fixes #32494 - Removing fact parser

domitea:
Removing fact parser according to moving them to core

This a plugin part of https://github.com/theforeman/foreman/pull/8491 PR where I add a Ansible parser to core.

PR #429, opened 2021-07-09
fixes #32494 - Removing fact parser

ezr-ondrej:
Removing fact parser according to moving them to core

Just a rebase of #408 to see if it fixes tests.


tbrisker:
the original pr has been rebased, closing

PR #430, opened 2021-07-13
Fixes #33001 - Update tests for new settings registry

tbrisker:


ezr-ondrej (approved):

ezr-ondrej:
For the current failures on jenkins, we would need to bump REX dependency to develop. We will need to wait for the next REX release for those to go away.
Thanks @tbrisker for taking care of this! :+1:

PR #432, opened 2021-07-26
Fixes #33139 - Show Ansible roles on Host detail page

xprazak2:


package.json
```diff
@@ -14,12 +14,12 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/builder": "^8.4.1",
-    "@theforeman/eslint-plugin-foreman": "^8.4.1",
-    "@theforeman/find-foreman": "^8.4.1",
-    "@theforeman/stories": "^8.4.1",
-    "@theforeman/test": "^8.4.1",
-    "@theforeman/vendor-dev": "^8.4.1",
+    "@theforeman/builder": "^8.9.0",
```
ezr-ondrej, line 17: this is not in packaging.
Should we use the packaged version (8.7.2)
I believe the others are just develop, but not sure if this wouldn't clash with the builder in packaging.

ezr-ondrej (commented):
I don't love adding the query in the same commit as using it, but it's simple enough, so I'm ok with that. Works as expected, can you only take a look into the builder dependency, if that could be an issue? We could probably manually change the version in packaging, but I'd rather have the versions matching.

ezr-ondrej:
Also rebasing should get the tests passing hopefully :)

ezr-ondrej:
Failures related... it's failure on the GraphQL test. Apart of that, I'm good here, let's get feedback on this.

xprazak2:
Tests are fixed now.

ezr-ondrej (commented):
We do not have inherited roles because of the load by explicit search, is it what we want?
I'm ok with it in the first version, just asking if we have a followup RM at least?

xprazak2:
Yes, we do not want inherited roles in this particular view and there will be a follow-up.

ezr-ondrej (approved):
Thanks @xprazak2 :+1:

PR #433, opened 2021-07-26
Fixes #33142 - Add missing template for variable overrides

xprazak2:
A follow-up fix after Puppet extraction.

### DIFF
--- a/app/views/api/v2/ansible_override_values/index.json.rabl
+++ b/app/views/api/v2/ansible_override_values/index.json.rabl
@@ -1,0 +1,3 @@
+collection @override_values
+
+extends 'api/v2/ansible_override_values/show'
--- a/app/views/api/v2/ansible_variables/show.json.rabl
+++ b/app/views/api/v2/ansible_variables/show.json.rabl
@@ -10,7 +10,7 @@ attributes :id, :variable, :ansible_role, :ansible_role_id, :description, :overr
 node do |ansible_variable|
   {
     :override_values => partial(
-      'api/v2/override_values/index',
+      'api/v2/ansible_override_values/index',
       :object => ansible_variable.lookup_values
     )
   }


tbrisker (changes requested):
There is a small rubocop failure, the failing tests seem to be coming from rex

tbrisker:
Looks like there are still related test failures, e.g:
```
 AccessPermissionsTest#test_0717_route lookup_values/index should have a permission that grants access [/home/runner/work/foreman_ansible/foreman_ansible/test/unit/shared/access_permissions_test_base.rb:29]:

File: app/views/api/v2/ansible_variables/show.json.rabl [hunk 1]
```diff
@@ -10,7 +10,7 @@ attributes :id, :variable, :ansible_role, :ansible_role_id, :description, :overr
 node do |ansible_variable|
   {
     :override_values => partial(
-      'api/v2/override_values/index',
```
tbrisker, line 13: @ezr-ondrej should we have kept this one in core, like we did for base and main?

app/views/api/v2/ansible_override_values/index.json.rabl
```diff
@@ -0,0 +1,3 @@
+collection @override_values
+
+extends 'api/v2/ansible_override_values/show'
```
ezr-ondrej, line 3: Why don't we have `base` instead, that would extend `override_values/main` to stay consistent?
Not a blocker tho

app/views/api/v2/ansible_variables/show.json.rabl
ezr-ondrej: Maybe, this should get merged anyway tho, as using the override_values index was probably missing the `ansible_variable_id`

ezr-ondrej: That being said if ansible is only one failing, I'd rather force providers to choose what to render, as usually they want something extra, but it gets missing, just because "it works with what's in core"

tbrisker:
@ezr-ondrej test failures were related to this change

ezr-ondrej:
> @ezr-ondrej test failures were related to this change

I don't believe so. I'm working on fixing those

PR #435, opened 2021-07-27
Remove ipaddress dependency

ekohl:
This uses the built in resolv module to see if it's an IP address instead of an unmaintained gem.

### DIFF
--- a/app/models/concerns/foreman_ansible/host_managed_extensions.rb
+++ b/app/models/concerns/foreman_ansible/host_managed_extensions.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'ipaddress'
+require 'resolv'
 module ForemanAnsible
   # Relations to make Host::Managed 'have' ansible roles
   module HostManagedExtensions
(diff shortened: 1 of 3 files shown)


app/models/concerns/foreman_ansible/host_managed_extensions.rb
```diff
@@ -58,7 +58,7 @@ module ClassMethods
       def import_host(*args)
         host = super(*args)
         hostname = args[0]
-        if IPAddress.valid?(hostname) &&
+        if Resolv::IPv4::Regex.match?(hostname) || Resolv::IPv6::Regex.match?(hostname)
```
xprazak2, line 61: Is there `&&` missing?

```ruby
(Resolv::IPv4::Regex.match?(hostname) || Resolv::IPv6::Regex.match?(hostname)) && host_nic = Nic::Interface.find_by(:ip => hostname)`, same as in ansible_report_importer?
```

ekohl: Oops. Fixed

adamruzicka (approved):
Test failures look unrelated, looks good to me

xprazak2:
Thanks everyone.

PR #436, opened 2021-07-28
Fixes #33152 - Do not use translated stings as hash keys

xprazak2:


tbrisker (approved):
Code looks good, but I'm not sure if the test failures are related or not... would prefer to have stable green tests before merging

tbrisker (approved):
Thanks @xprazak2 !

PR #437, opened 2021-07-29
Refs #33142 - add missing permissions

ezr-ondrej:
Additionally to #433 we need to add permissions for the parent controller.

### DIFF
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -29,23 +29,27 @@ Foreman::Plugin.register :foreman_ansible do
                :resource_type => 'AnsibleRole'
     permission :view_ansible_variables,
                {
+                 :lookup_values => [:index],
                  :ansible_variables => [:index, :auto_complete_search],
                  :'api/v2/ansible_variables' => [:index, :show]
                },
                :resource_type => 'AnsibleVariable'
     permission :edit_ansible_variables,
-               { :ansible_variables => [:edit, :update],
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
@tbrisker, @xprazak2 could one of you? :)

tbrisker (commented):
Would this give users with ansible permissions the ability to see other types of lookup values such as puppet?

ezr-ondrej:
> Would this give users with ansible permissions the ability to see other types of lookup values such as puppet?

It should not if they have the permissions set correctly :shrug:
It will, if they use the ancestor controller permissions.

tbrisker:
> It should not if they have the permissions set correctly :shrug:
> It will, if they use the ancestor controller permissions.

I'm not sure I follow? what do you mean if they have the permissions set correctly? 
isn't `/lookup_values` the path used by the puppet class parameters page, and ansible uses `/ansible_override_values`?

ezr-ondrej:
> isn't `/lookup_values` the path used by the puppet class parameters page, and ansible uses `/ansible_override_values`?

no, `lookup _values` is not a core route, the controller is in core usable by any plugin, that choses to use override and both ansible and does that. Not sure if it's the right strategy, because then we can't set the permissions for scoped `LookupValue`, but that's the way it works now, so the permissions should be set that way.

I'm not sure it is a big issue tho

tbrisker (approved):
Thanks @ezr-ondrej ! I agree it is odd but looks like it works.

PR #438, opened 2021-07-29
Fixes #33186 - New job template to run OpenSCAP scans via Ansible

phess:
Just like Remote Execution commands can be run with an SSH-based template as well as with an Ansible-based template, I'd like to be able to run OpenSCAP scans via Ansible in addition to the existing SSH-based method.

### DIFF
--- a/app/views/foreman_ansible/job_templates/run_openscap_scans_-_ansible_default.erb
+++ b/app/views/foreman_ansible/job_templates/run_openscap_scans_-_ansible_default.erb
@@ -1,0 +1,20 @@
+<%#
+name: Run OpenSCAP scans - Ansible Default
+job_category: OpenSCAP Ansible Commands
+description_format: Run scan for all OpenSCAP policies on given hosts
+snippet: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+<% raise "Create and assign a policy to this host before proceeding" if @host.policies_enc_raw.empty? -%>
+---
+- hosts: all
+  tasks:
+<% @host.policies_enc_raw.each do |policy| -%>
+    - shell: /usr/bin/foreman_scap_client <%= policy['id'] %>
+      register: out
+    - debug: var=out
+<% end -%>
+


phess:
I see a Redmine issue saying the issue at theforeman.org was opened on the wrong project. I'm not sure what project it should be, then. If this is really a problem, please let me know what project the issue needs to be filed under. Thanks.

app/views/foreman_ansible/job_templates/run_openscap_scans_-_ansible_default.erb
```diff
@@ -0,0 +1,22 @@
...
+%>
+
+<% raise "Create and assign a policy to this host before proceeding" if @host.policies_enc_raw.empty? -%>
+---
+- hosts: all
+  tasks:
+<% @host.policies_enc_raw.map do |policy| -%>
+    - shell:
+        cmd: |
+        <%=       indent(10) { "/usr/bin/foreman_scap_client #{policy['id']}" %>
```
adamruzicka, line 18: Isn't this an overkill? Wouldn't `shell: /usr/bin/foreman_scap_client <%= policy['id'] %>` work just the same?

phess: I agree. I was just mimicking how the "Ansible command" job template calls the shell module. You know, since this template here is really just an Ansible command too. :-)  But I totally agree to simplify it. Will do it right away. Thanks!

phess: Done. Thanks again, @adamruzicka

tbrisker:
ok to test

phess:
I've checked all failed check outputs and I don't really know what to make of them. If something really needs fixing I'll have to ask for an interpreter so I know what/where to fix. :-)

app/views/foreman_ansible/job_templates/run_openscap_scans_-_ansible_default.erb
```diff
@@ -0,0 +1,20 @@
...
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+<% raise "Create and assign a policy to this host before proceeding" if @host.policies_enc_raw.empty? -%>
+---
+- hosts: all
+  tasks:
+<% @host.policies_enc_raw.map do |policy| -%>
```
adamruzicka, line 15: `map` collects results into an array, which is unnecessary here, `each` would be better.

adamruzicka (commented):
One nitpick, otherwise lgtm. Test failures seem unrelated

app/views/foreman_ansible/job_templates/run_openscap_scans_-_ansible_default.erb
phess: Done. Thanks once again.

adamruzicka:
This looks fine to me, @xprazak2 could you check this out? I'm not familiar with openscap at all

xprazak2:
Works nicely, thanks!

PR #442, opened 2021-08-08
Fixes #31101 - copy ansible roles when cloning a hostgroup

adiabramovitch:


tbrisker (changes requested):
Looks good, can you include a test in https://github.com/theforeman/foreman_ansible/blob/master/test/unit/concerns/hostgroup_extensions_test.rb to verify that roles are indeed cloned correctly?

tbrisker (approved):
Thanks @adiabramovitch !

sbernhard:
@adiabramovitch / @tbrisker I cherry-picked this to foreman 2.3 and unfortunately, it does not work. Maybe you can try it again for the latest foreman version if it works.

PR #443, opened 2021-08-10
Refs #24367 - Support setting key passphrase through settings

adamruzicka:


app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -38,6 +38,7 @@ def proxy_command_options(template_invocation, host)
 
         def secrets(host)
           {
+            :key_passphrase => Setting[:remote_execution_ssh_key_passphrase],
```
ezr-ondrej, line 41: I don't know if it matters, but shouldn't this be string?
```suggestion
            'key_passphrase' => Setting[:remote_execution_ssh_key_passphrase],
```

adamruzicka: Most likely it doesn't really matter as it gets dumped as json and is then parsed back. SSHExecutionProvider[1] uses a symbol so I used it here too.

https://github.com/theforeman/foreman_remote_execution/blob/master/app/models/ssh_execution_provider.rb#L34

ezr-ondrej: Make sense :+1:

ezr-ondrej:
Thanks @adamruzicka ! :)

PR #445, opened 2021-08-12
Fixes #32983 - Edit Ansible variables on host detail page

xprazak2:


webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/EditableAction.scss
```diff
@@ -0,0 +1,7 @@
+.hideElement {
```
ezr-ondrej, line 1: don't we have a class for this in core already?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/EditableValueHelper.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import {
+  TextInput,
+  TextArea,
+  FormGroup,
+  FormSelect,
+  FormSelectOption,
+} from '@patternfly/react-core';
+
+const withFormGroup = Component => componentProps => {
```
ezr-ondrej, line 10: This seems we are implementing a Form helper in a plugin, I'd really love if we would manage to do this in core. There are about three in core already and this is yet another implementation :((

xprazak2: I would also love if this was something I did not have to do in a plugin, but core does not provide any PF4 form helpers as far as I looked.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/EditableAction.scss
xprazak2: Looks like I have poor grepping skills, where is it hidden?

ezr-ondrej: it's comming from bootstrap `.hide`.
Tho it might not be smart to rely on it, as core is not abusing it too much, so it might easily go away.

PR #446, opened 2021-08-16
Fixes #33263 - Ensure live output works

adamruzicka:
This briefly worked, but was then broken by an attempt to reduce the
amount of stored data on the proxy side[1].

[1] - https://github.com/theforeman/foreman_ansible/pull/393


adamruzicka:
Rubocop failure is unrelated

ezr-ondrej:
Do we have a PR against `smart_proxy_ansible`?

adamruzicka:
Looks like tests are borked here anyway

ezr-ondrej (approved):
I wouldn't say it was necesary to backport the test :)
But if it's already here... :)

PR #448, opened 2021-08-18
Fixes #32409 - Use new DSL to define settings

adiabramovitch:


xprazak2:
Looks good, but I defer to @ezr-ondrej as he has better knowledge of the DSL.

app/controllers/api/v2/ansible_inventories_controller.rb
```diff
@@ -90,7 +90,7 @@ def action_permission
       private
 
       def schedule_params
-        template_name = Setting::Ansible.find_by(:name => 'ansible_inventory_template').value
+        template_name = Setting.find_by(:name => 'ansible_inventory_template').value
```
ezr-ondrej, line 93: we need to do the following instead, otherwise we would not ge the default value (as it is no longer in the DB)
```suggestion
        template_name = Setting['ansible_inventory_template']
```

ezr-ondrej (commented):
LGTM, but we will need to wait for the validation PR to get in.

db/migrate/20210818083407_fix_setting_category.rb
```diff
@@ -0,0 +1,5 @@
+class FixSettingCategory < ActiveRecord::Migration[6.0]
```
ezr-ondrej, line 1: Can we have more specific name? this is clashing with the REX migration and Rails don't play well if two migrations have the same name.
```suggestion
class AnsibleSettingCategoryToDsl < ActiveRecord::Migration[6.0]
```

lib/foreman_ansible/register.rb
```diff
@@ -3,6 +3,82 @@
...
+                               'additional debugging output when running Ansible playbooks.'),
+              default: '0',
+              full_name: N_('Default verbosity level'),
+              value: nil,
+              collection:
+        { '0' => N_('Disabled'),
+          '1' => N_('Level 1 (-v)'),
+          '2' => N_('Level 2 (-vv)'),
+          '3' => N_('Level 3 (-vvv)'),
+          '4' => N_('Level 4 (-vvvv)') }
```
ezr-ondrej, line 44: This still needs to be a proc (or lambda) even if the result is static Hash, the renderer expects it to be always a `proc`
```suggestion
lambda do
        { '0' => N_('Disabled'),
          '1' => N_('Level 1 (-v)'),
          '2' => N_('Level 2 (-vv)'),
          '3' => N_('Level 3 (-vvv)'),
          '4' => N_('Level 4 (-vvvv)') }
 end
```

ezr-ondrej (changes requested):

lib/foreman_ansible/register.rb
```diff
@@ -3,6 +3,82 @@
...
+              full_name: N_('WinRM cert Validation')
+
+      setting 'ansible_verbosity',
+              type: :integer,
+              description: N_('Foreman will add this level of verbosity for '\
+                               'additional debugging output when running Ansible playbooks.'),
+              default: '0',
+              full_name: N_('Default verbosity level'),
+              value: nil,
+              lambda:
```
ezr-ondrej, line 39:
```suggestion
              collection: lambda do
```

PR #452, opened 2021-08-20
Fixes #33313 - Consider jobs on unreachable hosts failed

adamruzicka:
In [1] we started treating jobs with non-zero exit code, but with no
failures as successful. However, if a host is unreachable, ansible
doesn't treat it as a failure so such jobs hit this condition and were
treated as successful.

[1] - https://github.com/theforeman/smart_proxy_ansible/pull/38


ezr-ondrej (approved):

ezr-ondrej:
I believe this should have been picked into 4.2-stable-core, I'll do that and post tag the last foreman_ansible_core release.

PR #457, opened 2021-08-30
Fixes #26065 - Import playbook as a Job Template

adiabramovitch:
Add the option to Import ansible playbooks smart proxy,  while updating, creating, or deleting  them as job templates

### DIFF
--- a/app/controllers/api/v2/ansible_playbooks_controller.rb
+++ b/app/controllers/api/v2/ansible_playbooks_controller.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # API controller for Ansible Roles
+    class AnsiblePlaybooksController < ::Api::V2::BaseController
+      include ::Api::Version2
+      include ::ForemanAnsible::ProxyAPI
+
+      resource_description do
+        api_version 'v2'
+        api_base_url '/ansible/api'
+      end
+
+      api :PUT, '/ansible_playbooks/sync', N_('Sync Ansible playbooks')
+      param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
+      param :playbooks_names, Array, N_('Ansible  playbooks names to be synced')
+      def sync
+        # importer.import_playbooks(playbooks_names)
+        @task = plan_ansible_sync(proxy.id, playbooks_names)
+      end
+
+      api :GET, '/ansible_playbooks/fetch', N_('Fetch Ansible playbooks available to be synced')
+      param :proxy_id, :identifier, N_('Smart Proxy to fetch from'),
+            :required => true
+      def fetch
+        fetched = fetch_playbooks_names
+        render :json => { :results => { :playbooks_names => fetched } }
+      end
+
+      def action_permission
+        case params[:action]
+        when 'sync'
+          :sync
+        when 'fetch'
+          :fetch
+        else
+          super
+        end
+      end
+
+      def plan_ansible_sync(proxy, playbooks_names)
+        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy, playbooks_names)
+      end
+
+      private
+
+      # rubocop:disable Layout/DotPosition
+      def find_proxy
+        unless params[:proxy_id]
+          msg = _('Smart proxy id is required')
+          return render_error('custom_error', :status => :unprocessable_entity, :locals => { :message => msg })
+        end
+        SmartProxy.authorized(:view_smart_proxies)
+                  .find(params[:proxy_id])
+      end
+      # rubocop:enable Layout/DotPosition
+
+      def fetch_playbooks_names
+        proxy_api = find_proxy_api(proxy)
+        proxy_api.playbooks_names if @proxy
+      end
+
+      def playbooks_names
+        params.fetch(:playbooks_names, [])
+      end
+
+      def proxy
+        @proxy ||= find_proxy
+      end
+
+      def importer
+        @importer ||= ForemanAnsible::PlaybooksImporter.new(proxy)
+      end
+    end
+  end
+end
--- a/app/controllers/api/v2/ansible_playbooks_controller.rb
+++ b/app/controllers/api/v2/ansible_playbooks_controller.rb
@@ -7,6 +7,8 @@ module Api
       include ::Api::Version2
       include ::ForemanAnsible::ProxyAPI
 
+      before_action :find_proxy, only: [:fetch, :sync]
+
       resource_description do
         api_version 'v2'
         api_base_url '/ansible/api'
@@ -16,8 +18,7 @@ module Api
       param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
       param :playbooks_names, Array, N_('Ansible  playbooks names to be synced')
       def sync
-        # importer.import_playbooks(playbooks_names)
-        @task = plan_ansible_sync(proxy.id, playbooks_names)
+        @task = plan_ansible_sync(@proxy.id, playbooks_names)
       end
 
       api :GET, '/ansible_playbooks/fetch', N_('Fetch Ansible playbooks available to be synced')
@@ -30,47 +31,35 @@ module Api
 
       def action_permission
         case params[:action]
-        when 'sync'
-          :sync
-        when 'fetch'
-          :fetch
+        when 'sync', 'fetch'
+          :import
         else
           super
         end
       end
 
-      def plan_ansible_sync(proxy, playbooks_names)
-        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy, playbooks_names)
+      def plan_ansible_sync(proxy_id, playbooks_names)
+        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy_id, playbooks_names)
       end
 
       private
 
-      # rubocop:disable Layout/DotPosition
       def find_proxy
         unless params[:proxy_id]
           msg = _('Smart proxy id is required')
-          return render_error('custom_error', :status => :unprocessable_entity, :locals => { :message => msg })
+          render_error('custom_error', :status => :unprocessable_entity, :locals => { :message => msg })
+          return false
         end
-        SmartProxy.authorized(:view_smart_proxies)
-                  .find(params[:proxy_id])
+        @proxy = SmartProxy.find(params[:proxy_id])
       end
-      # rubocop:enable Layout/DotPosition
 
       def fetch_playbooks_names
-        proxy_api = find_proxy_api(proxy)
+        proxy_api = find_proxy_api(@proxy)
         proxy_api.playbooks_names if @proxy
       end
 
       def playbooks_names
         params.fetch(:playbooks_names, [])
-      end
-
-      def proxy
-        @proxy ||= find_proxy
-      end
-
-      def importer
-        @importer ||= ForemanAnsible::PlaybooksImporter.new(proxy)
       end
     end
   end
--- a/app/jobs/sync_playbooks.rb
+++ b/app/jobs/sync_playbooks.rb
@@ -1,8 +1,8 @@
 module ImportPlaybooksJob
   module Async
     class SyncPlaybooks < ::Actions::EntryAction
-      def plan(proxy, playbooks_names)
-        plan_self(proxy: proxy, playbooks_names: playbooks_names)
+      def plan(proxy_id, playbooks_names)
+        plan_self(proxy_id: proxy_id, playbooks_names: playbooks_names)
       end
 
       def run
@@ -14,7 +14,7 @@ module ImportPlaybooksJob
       end
 
       def proxy
-        SmartProxy.find(input[:proxy])
+        SmartProxy.find(input[:proxy_id])
       end
 
       def playbooks_names
--- a/app/jobs/sync_playbooks.rb
+++ b/app/jobs/sync_playbooks.rb
@@ -1,0 +1,25 @@
+module ImportPlaybooksJob
+  module Async
+    class SyncPlaybooks < ::Actions::EntryAction
+      def plan(proxy, playbooks_names)
+        plan_self(proxy: proxy, playbooks_names: playbooks_names)
+      end
+
+      def run
+        playbooks_importer = ForemanAnsible::PlaybooksImporter.new(proxy)
+        output[:result] = playbooks_importer.import_playbooks(playbooks_names)
+        ForemanAnsible::ImportPlaybooksSuccessNotification.deliver!(task)
+      rescue StandardError => e
+        ForemanAnsible::ImportPlaybooksErrorNotification.new(e, task).deliver!
+      end
+
+      def proxy
+        SmartProxy.find(input[:proxy])
+      end
+
+      def playbooks_names
+        input[:playbooks_names]
+      end
+    end
+  end
+end
--- a/app/services/foreman_ansible/import_playbooks_error_notification.rb
+++ b/app/services/foreman_ansible/import_playbooks_error_notification.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  class ImportPlaybooksErrorNotification < ::UINotifications::Base
+    private
+
+    def initialize(error, task)
+      @job_error = error
+      @subject = task
+    end
+
+    def create
+      ::Notification.create!(
+        :audience => Notification::AUDIENCE_ADMIN,
+        :notification_blueprint => blueprint,
+        :initiator => initiator,
+        :message => message,
+        :subject => subject,
+        :actions => {
+          :links => links
+        }
+      )
+    end
+
+    def blueprint
+      name = 'Sync_playbooks_failed'
+      @blueprint ||= NotificationBlueprint.unscoped.find_by(:name => name)
+    end
+
+    def message
+      _("Failed to import playbooks Due to: #{@job_error}")
+    end
+
+    def links
+      [{ :href => Rails.application.routes.url_helpers.foreman_tasks_task_path(:id => subject.id), :title => N_('Task Details') }]
+    end
+  end
+end
--- a/app/services/foreman_ansible/import_playbooks_error_notification.rb
+++ b/app/services/foreman_ansible/import_playbooks_error_notification.rb
@@ -11,7 +11,7 @@ module ForemanAnsible
 
     def create
       ::Notification.create!(
-        :audience => Notification::AUDIENCE_ADMIN,
+        :audience => Notification::AUDIENCE_USER,
         :notification_blueprint => blueprint,
         :initiator => initiator,
         :message => message,
--- a/app/services/foreman_ansible/import_playbooks_success_notification.rb
+++ b/app/services/foreman_ansible/import_playbooks_success_notification.rb
@@ -6,7 +6,7 @@ module ForemanAnsible
 
     def create
       ::Notification.create!(
-        :audience => Notification::AUDIENCE_ADMIN,
+        :audience => Notification::AUDIENCE_USER,
         :notification_blueprint => blueprint,
         :initiator => initiator,
         :message => message,
--- a/app/services/foreman_ansible/import_playbooks_success_notification.rb
+++ b/app/services/foreman_ansible/import_playbooks_success_notification.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  class ImportPlaybooksSuccessNotification < ::UINotifications::Base
+    private
+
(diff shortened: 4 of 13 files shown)


app/services/foreman_ansible/imprt_playbooks.rb
```diff
@@ -0,0 +1,14 @@
+
+module ForemanAnsible
+  class PlayBookImporter
+
+    def import_playbook
+      # Read a playbook example from github, needs to be fixed to read playbooks from certain directory
+      uri = URI("https://raw.githubusercontent.com/theforeman/forklift/master/playbooks/luna_demo_environment.yml")
+      file = Net::HTTP.get(uri)
+      template = JobTemplate.import_raw(File.read(file))
```
xprazak2, line 9: `import_raw` can certainly create a job template from a file, it is primarily intended for [text representation of a job template](https://github.com/theforeman/foreman_ansible/blob/master/app/views/foreman_ansible/job_templates/capsule_upgrade_-_ansible_default.erb), which also includes metadata. 

Since playbooks do not have metadata, you will need to create it. Metadata is just yaml enclosed in `<#%` and `%>` tags, so [YAML](https://ruby-doc.org/stdlib-2.5.1/libdoc/yaml/rdoc/YAML.html) module might be useful.
Adding name, snippet, model, job_category, provider_type and kind keys would be plenty for now. 
Once you have metadata, it will be a part of  job template content together with playbook file.

ezr-ondrej (commented):
I think we do not want to read all playbooks and then import them, but rather list available files and import read the file one by one (to save memory).

I think, that currently we import all the templates during seeding and I'd probably follow the same model (not because it's the best model necessarily, but because we want to be consistent).

Provisioning templates example: https://github.com/theforeman/foreman/blob/639f102a46bfde5e897d1a5225a34e0477f557d5/db/seeds.d/070-provisioning_templates.rb#L9
or to have example from plugin, JobTemplates would be the closest to our use-case here:
https://github.com/theforeman/foreman_remote_execution/blob/fe4613c6f2e7a0f1d184e51b3b5325115cdedb02/db/seeds.d/70-job_templates.rb

config/routes.rb
```diff
@@ -82,6 +82,13 @@
           end
         end
 
+        resources :ansible_playbooks do
+          collection do
```
xprazak2, line 86: Rails generate routes for all restful actions by default:

playbook-routes

Since we will not be using them, we do not want them to be created. We can use `:only` for that, see block below.

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,34 @@
...
+    def initialize(proxy = nil)
+      @ansible_proxy = proxy
+    end
+
+    def import_playbook
+      import_playbooks_content playbooks if @ansible_proxy.present?
+    end
+
+    def import_playbooks_content(playbooks)
+      playbooks.map | playbook | import_parsed(playbook)
```
xprazak2, line 17: Should this be `playbooks.map { |playbook| import_parsed playbook }` ?

Also, `import_parsed` is not yet implemented.

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      before_action :find_proxy, :only => [:fetch, :sync]
+      before_action :create_importer, :only => [:fetch, :sync]
+
+      api :PUT, '/ansible_playbooks/sync', N_('Sync Ansible playbooks')
+      param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
+      param :playbook_names, Array, N_('Ansible playbook names to be synced')
+      def sync
+        new_playbooks = @importer.import_playbook
```
xprazak2, line 21: Should the name be pluralized since the method will be importing multiple playbooks?

xprazak2:
Fetching names works as expected, sync was not working for me just yet.

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        @imported = new_playbooks
+      end
+
+      api :GET, '/ansible_playbooks/fetch', N_('Fetch Ansible playbooks available to be synced')
+      param :proxy_id, :identifier, N_('Smart Proxy to fetch from'),
+            :required => true
+      def fetch
+        fetched =  @importer.fetch_playbooks_names
+        respond_to do |format|
+          format.json do
```
xprazak2, line 31: I do not think you need those 2 blocks as API controllers respond with json by default...

app/services/foreman_ansible/playbooks_importer.rb
adiabramovitch: I am supposed to use import_parsed from JobTemplate or am I wrong? 
https://github.com/theforeman/foreman_remote_execution/blob/fe4613c6f2e7a0f1d184e51b3b5325115cdedb02/app/models/job_template.rb#L72

xprazak2: Using existing code is good, so use it if fits your needs.

xprazak2:
Nice progress, I added a couple of comments.

xprazak2:
I did a bit more testing, the vars section does not transform correctly for me. When having a playbook:

playbook

the result was:

vars-transform

We probably also want to separate metadata and playbook content with a new line.

Also, I got a the following:
sync-output

the playbooks with empty keys in place of ids were not created - there were errors with duplicate template input name but I did not dig deeper.

app/models/ansible_playbooks.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Simple model to store basic info about the Ansible role
+class AnsiblePlaybook < ApplicationRecord
```
xprazak2, line 4: Why do we need a new model for playbook if we are converting playbooks to job templates?

lib/foreman_ansible/register.rb
```diff
@@ -76,6 +76,14 @@
                :resource_type => 'Hostgroup'
     permission :generate_ansible_inventory,
                { :'api/v2/ansible_inventories' => [:schedule] }
+    permission :import_ansible_playbooks,
+               { :ansible_playbooks => [:sync],
```
xprazak2, line 80: There is only API controller for now, so this line is not needed. Same for the permission below.

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      before_action :find_proxy, :only => [:fetch, :sync]
+      before_action :create_importer, :only => [:fetch, :sync]
+
+      api :PUT, '/ansible_playbooks/sync', N_('Sync Ansible playbooks')
+      param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
+      param :playbooks_names, Array, N_('Ansible  playbooks names to be synced')
+      def sync
+        params = @importer.import_playbooks(playbooks_names)
+        if params[:changed][:create].present? || params[:changed][:update].present?
```
xprazak2, line 22: Is always true because both keys are initialized to `[]`

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,192 @@
...
+      task
+    end
+
+    def import_playbooks(playbooks_names)
+      filter_playbooks = playbooks_names.empty? ? playbooks : find_playbooks(playbooks_names)
+      import_playbooks_content filter_playbooks
+    end
+
+    def import_playbooks_content(playbooks)
+      params = { changed: { 'update': [], 'create': [], 'delete': [] } }
```
xprazak2, line 25: Delete key is never used, and quotes are not needed as there is no space in the key name.

xprazak2:
Creating templates works as expected, updating does not save the changes.

The action renders no content when the task is created,  so I get an empty response, also success notification is not created for me:


notification

I also found that when playbook has vars as array, it gets transformed to hash (the `-` are lost):


playbook-before

playbook-after

I also found that playbooks coming from proxy are always marked for update even if there has been no change on proxy.

app/services/foreman_ansible/playbooks_importer_helper.rb
```diff
@@ -0,0 +1,72 @@
+module ForemanAnsible
+  class PlaybooksImporterHelper
```
xprazak2, line 2: I like that you divide things to keep the classes small. I could not determine the pattern why some methods stay in importer while other move to helper, what is the decision key?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def add_metadata(playbook_name)
+      "<%#
+      name: #{playbook_name}
+      snippet: false
+      job_category: Ansible Playbook - Imported
+      provider_type: Ansible
+      kind: job_template
+      model: JobTemplate
+      %>\n"
```
xprazak2, line 64: Since this is one long string spanning across multiple lines, the whitespace at the beginning of the line is also included and the metadata in template seem a little bit misaligned:

metadata

So we might think of how to make sure we get the formatting we want, perhaps [heredoc](https://www.rubyguides.com/2018/11/ruby-heredoc/) might help.

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+          :fetch
+        else
+          super
+        end
+      end
+
+      private
+
+      def changed?(params)
+        params[:changed][:update].empty? && params[:changed][:create].empty? && params[:changed][:delete].empty? ? false : true
```
xprazak2, line 51: No need to return a boolean directly as it is a result of checking for changes:

```ruby
def changed?(params)
  !(params[:changed][:update].empty? && params[:changed][:create].empty? && params[:changed][:delete].empty?)
end
```

xprazak2:
The success notification now shows for me, the sync still does not render any content.

xprazak2:
The array of vars is no longer transformed into hash, but the var names have additional colon:

vars-before
vars-after

app/services/foreman_ansible/playbooks_importer_helper.rb
adiabramovitch: I thought of helper methods that are not really related to the logic of the importer will be on a different class

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,132 @@
...
+      <<-METADATA
+      <%#
+        name: #{playbook_name}
+        snippet: false
+        job_category: Ansible Playbook - Imported
+        provider_type: Ansible
+        kind: job_template
+        model: JobTemplate
+      %>
+      METADATA
```
xprazak2, line 66: heredoc maintains the whitespace, so keeping the indentation will add 6 spaces before `<%#`

```ruby
    def add_metadata(playbook_name)
<<-METADATA
<%#
  name: #{playbook_name}
  snippet: false
  job_category: Ansible Playbook - Imported
  provider_type: Ansible
  kind: job_template
  model: JobTemplate
%>
METADATA
    end
```

If you think this has poor readability, then we probably need to control the whitespace in s different way:

```ruby
def add_metadata(playbook_name)
      [
        '<%#',
        "  name: #{playbook_name}",
        "  snippet: false",
        "  job_category: Ansible Playbook - Imported",
        "  provider_type: Ansible",
        "  kind: job_template",
        "  model: JobTemplate",
        "%>"
      ].join("\n")
    end
```

app/services/foreman_ansible/import_playbooks_error_notification.rb
```diff
@@ -0,0 +1,31 @@
...
+      )
+    end
+
+    def blueprint
+      name = 'Sync_playbooks_failed'
+      @blueprint ||= NotificationBlueprint.unscoped.find_by(:name => name)
+    end
+
+    def message
+      _("Failed to import playbooks Due to: #{@job_error}")
```
xprazak2, line 28: `Failed to import playbooks due to:` ?

app/jobs/sync_playbooks.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Async
+    class SyncPlaybooks < ::Actions::EntryAction
+      require_relative '../services/foreman_ansible/playbooks_importer'
+
+      def plan(changed, proxy, playbooks_names)
+        plan_self(changed: changed, proxy: proxy, playbooks_names: playbooks_names)
+      end
+
+      def run
+        playbooks_importer = ForemanAnsible::PlaybooksImporter.new(proxy)
```
xprazak2, line 11: There is an instance of importer in already created in the controller, so you can pass it in and re-use it. That way, you do not need a proxy and instantiate a new importer.

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      before_action :find_proxy, :only => [:fetch, :sync]
+      before_action :create_importer, :only => [:fetch, :sync]
+
+      api :PUT, '/ansible_playbooks/sync', N_('Sync Ansible playbooks')
+      param :proxy_id, :identifier, :required => true, :desc => N_('Smart Proxy to sync from')
+      param :playbooks_names, Array, N_('Ansible  playbooks names to be synced')
+      def sync
+        params = @importer.import_playbooks(playbooks_names)
+        if changed?(params)
+          ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, params[:changed], @ansible_proxy, playbooks_names)
```
xprazak2, line 23: Since the rabl view uses `@task`, you need to assign the task object into the instance variable.

xprazak2:
I still see playbooks in the `updated` section even though nothing has changed on proxy. Is it just me?

Also, it might be good to include errors for playbooks that were not imported:

raw_task

xprazak2:
A rough sketch for the notification link:  https://github.com/theforeman/foreman_ansible/commit/f4abbe0500c13142d1c85171000bcabd89490614

adiabramovitch:
[test foreman_ansible]

app/services/foreman_ansible/import_playbooks_success_notification.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  class ImportPlaybooksSuccessNotification < ::UINotifications::Base
+    private
+
+    def create
+      ::Notification.create!(
+        :audience => Notification::AUDIENCE_ADMIN,
```
ezr-ondrej, line 9:
```suggestion
        :audience => Notification::AUDIENCE_USER,
```

ezr-ondrej (commented):
Looks good overal, I'll going to test :) few nits I've noticed.

adiabramovitch:
> Looks good overal, I'm going to test :) few nits I've noticed.

@ezr-ondrej did you get a chance to test it? :)

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      private
+
+      # rubocop:disable Layout/DotPosition
+      def find_proxy
+        unless params[:proxy_id]
+          msg = _('Smart proxy id is required')
+          return render_error('custom_error', :status => :unprocessable_entity, :locals => { :message => msg })
+        end
+        SmartProxy.authorized(:view_smart_proxies)
```
ezr-ondrej, line 54: I'd save this into a variable directly here and I'm not sure we want to require extra permission here. You have permission to fetch playbooks, you should be able to do so from any proxy (be wary, this is only my opinion, I might be wrong here).
```suggestion
        @proxy = SmartProxy.find(params[:proxy_id])
```

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+      end
+
+      def plan_ansible_sync(proxy, playbooks_names)
+        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy, playbooks_names)
+      end
+
+      private
+
+      # rubocop:disable Layout/DotPosition
```
ezr-ondrej, line 48: Why not listen to rubocop instead? :)

ezr-ondrej (commented):
Left comments inline and also two main concerns here:
1. This is only API, do we have a UI follow-up?
2. The playbooks get created without org and loc, can we do something about that ?
  In SeedHelper (https://github.com/theforeman/foreman/blob/e231a35e05dc4b958c3244d674deeb21654b11d8/lib/seed_helper.rb#L127) we just add all, I guess that is good enough strategy

app/controllers/api/v2/ansible_playbooks_controller.rb
adiabramovitch: Yes that makes sense but I  see that we use this format in a lot of places, so isn't it a precaution?

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+        proxy_api.playbooks_names if @proxy
+      end
+
+      def playbooks_names
+        params.fetch(:playbooks_names, [])
+      end
+
+      def proxy
+        @proxy ||= find_proxy
+      end
```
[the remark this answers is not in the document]
adiabramovitch: I think, not sure, that Shimon said that there is an advantage of doing it like that but I honestly don't remember and it seems to work this way as well as most places I saw on the project did it like that :)

lib/foreman_ansible/register.rb
```diff
@@ -158,6 +158,10 @@
                :resource_type => 'Hostgroup'
     permission :generate_ansible_inventory,
                { :'api/v2/ansible_inventories' => [:schedule] }
+    permission :import_ansible_playbooks,
+               { :'api/v2/ansible_playbooks' => [:sync] }
+    permission :fetch_ansible_playbooks,
+               { :'api/v2/ansible_playbooks' => [:fetch] }
```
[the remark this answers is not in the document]
adiabramovitch: yes, you are right, thanks :)

app/controllers/api/v2/ansible_playbooks_controller.rb
ezr-ondrej: we do use it, but I've been removing it in some places already as it forces to have permissions that you do not necessarily need for given operation.

app/controllers/api/v2/ansible_playbooks_controller.rb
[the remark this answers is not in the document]
ezr-ondrej: no, variables needed for the controller action are always fetched in the before_action, it is just automated to extend you don't see it everywhere, but we use `before_action` for most controllers :)

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def action_permission
+        case params[:action]
+        when 'sync'
+          :sync
+        when 'fetch'
+          :fetch
+        else
+          super
+        end
```
ezr-ondrej, line 40: I believe this needs to be updated now:
```suggestion
        case params[:action]
        when 'sync', 'fetch'
          :import
        else
          super
        end
```

ezr-ondrej (approved):
We need to change the permissions for action, but then we seem to be there IMHO, can you change that pls so I can retest? :)

app/controllers/api/v2/ansible_playbooks_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def plan_ansible_sync(proxy, playbooks_names)
+        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy, playbooks_names)
```
ezr-ondrej, line 44:
```suggestion
      def plan_ansible_sync(proxy_id, playbooks_names)
        ForemanTasks.async_task(ImportPlaybooksJob::Async::SyncPlaybooks, proxy_id, playbooks_names)
```

app/controllers/api/v2/ansible_playbooks_controller.rb
adiabramovitch: is should be:   ```
     case params[:action]
        when 'sync', 'fetch'
          :sync
        else
          super
        end
``` ? why :import?

ezr-ondrej: you've called the permission `import_ansible_playbooks` :) thus it needs to be import for it to map the actions to a correct permission.

ezr-ondrej (approved):
Thanks @adiabramovitch ! :+1:

ezr-ondrej:
One issue I'll not block on: https://projects.theforeman.org/issues/34593

PR #458, opened 2021-09-01
Fixes #32984 - Create Ansible variable override on host detail page

xprazak2:
needs https://github.com/theforeman/foreman_ansible/pull/445


xprazak2:
No longer blocked

ezr-ondrej:
Needs rebase now

xprazak2:
1. Update and delete after create should be fixed now.
2. I believe that is expected behavior and we want to show only variables that are already set as `override: true`. But we can change it in the future if we decide to show all variables.

PR #459, opened 2021-09-06
Fixes #33420 - Change roles on host detail page

xprazak2:


ezr-ondrej:
Could we split the GraphQL (new mutation) and the reader adjusts (split to inherited and directly assigned roles) into separate commits, preferably separate PRs?

xprazak2:
I made a split as requested, now depends on https://github.com/theforeman/foreman_ansible/pull/462

PR #460, opened 2021-09-07
Fixes #33424 - Remove gray margins from host details tab

xprazak2:
Needs https://github.com/theforeman/foreman/pull/8764

Before:

before

Now:

now

cc @amirfefer @MariSvirik


ezr-ondrej:
Can we fix the JS tests, or is that unrelated? I don't care on this change too much, I'm ok to get this in :+1:

ezr-ondrej (approved):
Looks bit weird IMHO, the tabs should be more visible, but I'm not a designer, if we want this, lets do it :)
Thanks @xprazak2 ! :+1:

PR #461, opened 2021-09-17
Fixes #30689 - roles taxable with Organization

ezr-ondrej:


app/models/ansible_role.rb
```diff
@@ -38,4 +39,14 @@ class AnsibleRole < ApplicationRecord
   class Jail < Safemode::Jail
     allow :name
   end
+
+  def clone_to_org!(new_org_id = nil)
+    cloned = deep_clone
+    cloned.organization_id = new_org_id if new_org_id
+    cloned.save!
+    ansible_variables.each do |variable|
+      cloned.ansible_variables << variable.deep_clone(except: :ansible_role_id, include: :lookup_values)
```
xprazak2, line 48: Is cloning overrides a good idea? We might end up with cloned roles in OrgB that have overrides for host in OrgA...

app/models/ansible_role.rb
```diff
@@ -6,7 +6,8 @@ class AnsibleRole < ApplicationRecord
   include Authorizable
 
   self.include_root_in_json = false
-  validates :name, :presence => true, :uniqueness => true
+  validates :name, :presence => true, :uniqueness => { scope: :organization_id }
```
xprazak2, line 9: Having taxonomies behave completely differently for Ansible roles than all other models could cause a lot of confusion. Are we sure we want to do this?

Will there be a follow-up for Locations?

app/models/ansible_role.rb
ezr-ondrej: We might, but I'm not sure how else to approach it, droping the overrides sounds worse then having useless ones.

app/models/ansible_role.rb
ezr-ondrej: This was part of a discussion, that taxonomies should behave like this for everything and us trying this concept (Katello concept) in here and if it would be easier to have single org and strict split of resources by orgs

xprazak2: Ok, let's go ahead and see.

app/models/ansible_role.rb
xprazak2: I am more afraid of the cloned ones not being useless and that their value will actually be applied since they are not taxable. I'll need to thoroughly test that.

ezr-ondrej:
Given my priorities currenly, I'm closing this.

PR #462, opened 2021-09-20
Fixes #33521 - Add mutation to assign ansible roles

xprazak2:


ezr-ondrej (commented):
This is confusing... The reordering at the beggining would make sense if we would want to keep the currently assigned, but if we do not, wouldn't it be easier just to simply:
```
host.ansible_role_ids = ansible_role_ids
```
The possitions are recalculated by `acts_as_list` properly, if you do not mix old assignments into it... :thinking: 
Or is there any particular reason we need to do it manually?

xprazak2:
> The possitions are recalculated by `acts_as_list` properly, if you do not mix old assignments into it... thinking

That is exactly the problem - when I am changing the order of the roles that are already assigned, `acts_as_list` just ignores it.

ezr-ondrej:
> That is exactly the problem - when I am changing the order of the roles that are already assigned, `acts_as_list` just ignores it.

Ok, I guess this is safe enough, just bit unreadable.

ezr-ondrej (approved):
Thanks @xprazak2 !

PR #419, opened 2021-06-04
[WIP] Fixes ##32722 - Show recurring job for Ansible runs

xprazak2:
Needs https://github.com/theforeman/foreman_ansible/pull/406 and https://github.com/theforeman/foreman_remote_execution/pull/597

PR #463, opened 2021-09-20
Fixes #33522 - Create Ansible config job

xprazak2:


xprazak2:
Blocked by https://github.com/theforeman/foreman_ansible/pull/419

xprazak2:
Rebased.

webpack/components/AnsibleHostDetail/components/JobsTab/NewRecurringJobHelper.js
```diff
@@ -0,0 +1,100 @@
...
+    jobInvocation: {
+      hostIds: [hostId],
+      feature: 'ansible_run_host',
+      targetingType: 'static_query',
+      scheduling: {
+        startAt: date,
+      },
+      recurrence: {
+        cronLine: toCron(date, repeat),
+        purpose: `ansible-host-${hostId}`,
```
ezr-ondrej, line 56: Could we have a function, that would return the purpose, I see it quite dangerous to have it in two places and rely on them being changed at the same time.

ezr-ondrej (commented):
Works as expected, just one nit if we could extract the `purpose` into a helper function.

webpack/components/AnsibleHostDetail/components/JobsTab/NewRecurringJobHelper.js
xprazak2: I agree, updated.

ezr-ondrej:
Now linter is barking about something....

ezr-ondrej (approved):
Thanks @xprazak2 ! :+1:

PR #467, opened 2021-09-30
Fixes #33602 - Do not set effective user var

adamruzicka:
if it is the same as connection user


ezr-ondrej (approved):
I'd expect ansible to do this for us, but I guess it does not :)

PR #473, opened 2021-10-20
Fixes #33726 - Add view permissions for host detail page

xprazak2:


ezr-ondrej:
Needs rebase now so I do not know if this would be resolved, but there is permission denied twice on the jobs tab, which looks quite odd

xprazak2:
Permission denied should show only once now.

ezr-ondrej (approved):
Works well, thanks @xprazak2 ! :+1:

PR #476, opened 2021-11-02
Fixes #33827 - Check permissions when changing overrides

xprazak2:
Needs https://github.com/theforeman/foreman_ansible/pull/473


xprazak2:
Should be ready.

ezr-ondrej:
It has JS failures and now a conflict as well

xprazak2:
Rebased.

ezr-ondrej:
This also enables delete, but server refuses to delete.
It is still step forward, so I'm going to merge, but it would be nice if we could fix the delete.

PR #477, opened 2021-11-02
Fixes #33182 - Schedule recurring job for hostgroups

xprazak2:
Needs https://github.com/theforeman/foreman_remote_execution/pull/661

### DIFF
--- a/app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
+++ b/app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
@@ -5,11 +5,14 @@
     display_link_if_authorized(_('Run all Ansible roles'), hash_for_play_roles_hostgroup_path(id: hostgroup), :'data-no-turbolink' => true, title: _('Run all Ansible roles on hosts belonging to this host group'))
   end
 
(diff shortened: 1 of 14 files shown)


xprazak2:
Seems like everything is working with the search on purpose, so the targeting search will probably not be needed.

app/views/foreman_ansible/ansible_roles/_hostgroup_ansible_roles_button.erb
```diff
@@ -5,11 +5,14 @@
     display_link_if_authorized(_('Run all Ansible roles'), hash_for_play_roles_hostgroup_path(id: hostgroup), :'data-no-turbolink' => true, title: _('Run all Ansible roles on hosts belonging to this host group'))
   end
 
+  assign_jobs = link_to(_("Configure Ansible Job"), "/ansible/hostgroups/#{hostgroup.id}", { class: 'la' })
```
ezr-ondrej, line 8: Shouldn't we check for permissions here?

xprazak2: We probably should, changed.

ezr-ondrej (approved):
It's a good work around, this will have to be refactored once we have actual Hostgroup details page :)
Thanks @xprazak2 ! :+1:

PR #479, opened 2021-11-02
Fixes #33835 - Show only one empty state for jobs

xprazak2:


ezr-ondrej:
This is not one at a time, but only one, right? So when I schedule a run, I will not see that "there would be a table of previous runs if there would be some"...
How hard would be making the bottom appear ony when we have something scheduled?

xprazak2:
Yes, this is only one.

It would not be a difficult in terms of code complexity to achieve but it would require significant changes to the structure of the tab. Right now, the 'scheduled' and 'previous' are independent sibling components unaware of each other and each fetching its data so it would require moving the data fetching for both components into the common parent.

ezr-ondrej:
I'm ok doing it later on :+1:

PR #486, opened 2021-11-09
Fixes #33883 - Handle permissions for Ansible config jobs

xprazak2:
needs https://github.com/theforeman/foreman-tasks/pull/658

### DIFF
--- a/webpack/components/AnsibleHostDetail/components/JobsTab/RecurringJobsTable.js
+++ b/webpack/components/AnsibleHostDetail/components/JobsTab/RecurringJobsTable.js
@@ -35,6 +35,19 @@ const RecurringJobsTable = ({ jobs, resourceName, resourceId }) => {
     );
   };
 
+  const actionItems = job => {
+    const items = [];
+    if (job.recurringLogic.meta.canEdit) {
+      items.push({
+        title: __('Cancel'),
+        onClick: onJobCancel(job.recurringLogic.id),
+        key: 'cancel',
+      });
+    }
+
+    return { items };
+  };
+
   return (
     <React.Fragment>
       <h3>{__('Scheduled recurring jobs')}</h3>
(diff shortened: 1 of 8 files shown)


adamruzicka:
foreman-tasks-5.2.0 containing the necessary changes was released. Could you bump the dependency here?

xprazak2:
It seems like we do not depend on tasks directly here, but rather on rex. What rex version should I bump to?

adamruzicka:
As far as I know rex doesn't depend on anything that was added in tasks 5.2.0, so just bumping rex dependency won't help you get the bits you need. I'd say relying on transitive dependencies is incorrect in case you explicitly need something

xprazak2:
Good point, I added tasks.

webpack/components/AnsibleHostDetail/components/JobsTab/RecurringJobsTable.js
```diff
@@ -35,6 +35,19 @@ const RecurringJobsTable = ({ jobs, resourceName, resourceId }) => {
     );
   };
 
+  const actionItems = job => {
+    const items = [];
+    if (job.recurringLogic.meta.canEdit) {
+      items.push({
+        title: __('Cancel'),
+        onClick: onJobCancel(job.recurringLogic.id),
```
amirfefer, line 43: doesn't it suppose to run on click? 
```suggestion
        onClick: () => onJobCancel(job.recurringLogic.id),
```

amirfefer (commented):
Thanks @xprazak2! overall looks good 👍

webpack/components/AnsibleHostDetail/components/JobsTab/RecurringJobsTable.js
xprazak2: It runs on click, see l.27.

ezr-ondrej:
The JS tests are failing

ares:
Who has the ball now? :-)

ezr-ondrej (approved):
I tried to fix the tests over the weekend, but I failed.
The failures are not related, but the failing tests would test the changed components, so we don't know if we didn't break something.
Tho I'll solve that once I manage to fix the tests :+1: 
Thanks @xprazak2 !

PR #489, opened 2021-11-23
Fixes #33980 - backport the RHEL OS parsing

ares:


ares:
This is a CP of https://github.com/theforeman/foreman/pull/8943 for the older versions, I didn't bring back the tests since we don't have the infra here for the static fixtures. If tests are green in core, it's unlikely the code wouldn't work here.

adamruzicka (approved):
Looks good, let's get this in

ezr-ondrej:
I guess we are pulling in the new smart_proxy_dynflow, which we should not have (that's why the tests fail).
That got me thinking, won't this happen on debian @adamruzicka?

ezr-ondrej (approved):
Last comment has nothing to do with this PR, it is fine to get in :+1:

adamruzicka:
> won't this happen on debian?

Newer smart_proxy_dynflow should not be available in debian repos IIRC

PR #491, opened 2021-12-06
Fixes #34094 - use report importer registry

lzap:
See https://github.com/theforeman/foreman/pull/8958#issuecomment-986732271

Do not merge until the core PR is in.

Also minimum foreman version will be 3.2 now, I cannot find this in `engine.rb`?

### DIFF
--- a/app/services/foreman_ansible/ansible_report_importer.rb
+++ b/app/services/foreman_ansible/ansible_report_importer.rb
@@ -18,10 +18,6 @@ module ForemanAnsible
         partial_hostname_match(hostname)
       end
 
-      def self.authorized_smart_proxy_features
-        super + ['Ansible']
-      end
-
       def partial_hostname_match(hostname)
         return @host unless @host.new_record?
         hosts = Host.where(Host.arel_table[:name].matches("#{hostname}.%"))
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
> Also minimum foreman version will be 3.2 now, I cannot find this in `engine.rb`?

It is in `lib/foreman_ansible/register.rb`

ezr-ondrej:
Actually this would work without the core patch, wouldn't it? We call super there, so we should get the correct result with `Ansible` with bit of code order luck

lzap:
Hmm tests are failing, any idea?

```
2021-12-10T11:37:26 [W|app|] Failed to register foreman-tasks plugin (PG::UndefinedTable: ERROR:  relation "foreman_tasks_recurring_logics" does not exist
 | LINE 8:  WHERE a.attrelid = '"foreman_tasks_recurring_logics"'::regc...
```

> Actually this would work without the core patch, wouldn't it? We call super there, so we should get the correct result with `Ansible` with bit of code order luck

Unsure, it is a clusterfuck implementation. If you say it will work and tests will pass then sure, let's do it!

ezr-ondrej:
foreman_ansible tests are broken, we are on GH actions here for a while now and so far it works quite nicely

ezr-ondrej (approved):
Thanks @lzap ! :+1:

ezr-ondrej:
CP 7.0: 84b32f0

PR #492, opened 2021-12-07
Refs #33522 - fix ExclamationCircleIcon import

evgeni:


webpack/formHelper.js
```diff
@@ -7,7 +7,7 @@ import {
   DatePicker,
   TimePicker,
 } from '@patternfly/react-core';
-import ExclamationCircleIcon from '@patternfly/react-icons/dist/js/icons/exclamation-circle-icon';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
```
evgeni, line 10: I have no idea why, but CI won't pass w/o the `{}`

ezr-ondrej: it is how the JS import works... only default export can be imported without the `{}` and that can be just one per file. If you export multiple stuff from one file, you need `{}`

ezr-ondrej (approved):
@evgeni FTW :+1:

webpack/formHelper.js
evgeni: Another point on the "reasons Evgeni will never like JS" list

PR #493, opened 2021-12-14
Fixes #34150 -  align secondary tabs in host page

amirfefer:
following https://github.com/theforeman/foreman/pull/8996

before:
Screen Shot 2021-12-14 at 18 40 26

after:
Screen Shot 2021-12-14 at 19 52 45

### DIFF
--- a/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
+++ b/webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
@@ -21,6 +21,7 @@ const AnsibleHostDetail = ({
       {response?.id && (
         <>
           <Tabs
+            className="ansible-host-details-tabs"
             onSelect={(evt, subTab) => hashHistory.push(subTab)}
             activeKey={pathname?.split('/')[2]}
             isSecondary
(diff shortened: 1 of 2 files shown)


Andrewgdewar:
Looking good! I would Ack but I'm an imposter!

webpack/components/AnsibleHostDetail/AnsibleHostDetail.js
```diff
@@ -21,6 +21,7 @@ const AnsibleHostDetail = ({
       {response?.id && (
         <>
           <Tabs
+            className="host-details-tabs"
```
ezr-ondrej, line 24: it is quite generic className, give we are not scoping the css, should we give it bit more specific name (perhaps `ansible-host-details-tabs`?

amirfefer:
@ezr-ondrej, done :)

ezr-ondrej:
> Looking good! I would Ack but I'm an imposter!

Btw @Andrewgdewar imposters can also ack :) it's always nice to see that someone agrees with the solution. It gives the maintainers more confidence while merging :+1:

ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

ezr-ondrej:
CP 7.0: fab805d

PR #497, opened 2022-01-12
Fixes #34253 - Tags input is expected as a string

obol89:
Fixes the issue when provided tags are stored in an array that is saved as a string. It could be achieved differently as well, and we could accept Arrays and Strings by changing InvocationProviderInputValue model. I decided to go this way, as Ansible accepts strings only as tags. So, it should look like that in ansible:
`ansible-playbook example.yml --tags "configuration,packages"`
and like that when we use hammer:
`hammer job-invocation --job-template "Ansible Roles - Example Job" --search-query my-server --tags "configuration,packages"`

https://community.theforeman.org/t/hammer-job-invocation-tags-have-wrong-formatting/26797
https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -88,7 +88,7 @@ if defined? ForemanRemoteExecution
             :children => [
               {
                 :name => :tags,
-                :type => Array,
+                :type => String,
                 :opts => { :required => false, :desc => N_('A comma separated list of tags to use for Ansible run') }
               },
               {


ezr-ondrej (approved):
I agree this is a correct solution, it was a nice thought to enable array wich would feel more natural in hammer, but this matches closer what ansible is using and what we input in UI :+1:
Thanks @obol89 !

ezr-ondrej:
ok to test

ezr-ondrej:
CP 7.0: cfa3a8b

PR #500, opened 2022-01-18
Test Host Detail page as default

ezr-ondrej:
Fixing the tests after https://github.com/theforeman/foreman/pull/9002 got in.


ezr-ondrej:
I'm self-merging as this fixes the tests and will not break anything if something goes wrong.

PR #502, opened 2022-01-18
Fixes #34283 - customize variables EmptyState message

ezr-ondrej:
This corrects the way how we are passing the header for the EmptyState component.
We also can now add a description to explain the page a bit better to Users.


amirfefer (approved):
Thanks @ezr-ondrej !

ezr-ondrej:
CP 7.0: 7393818

PR #503, opened 2022-01-18
Fixes #34286 - add bottom pagination in all new host page tables

amirfefer:
for example:
Screen Shot 2022-01-18 at 23 55 53


ezr-ondrej (approved):
Thanks @amirfefer!  :+1:
I've rebased for you and added the variant prop, just for consistency and this is now :green_circle:

PR #505, opened 2022-01-25
Fixes #34313 - Import roles - disable Submit button

stejskalleos:
Disable submit button when:
* action status is PENDING
* No roles are selected


ezr-ondrej (approved):
Very nice improvement :+1:

PR #506, opened 2022-01-26
Fixes #34318 - Parse playbooks from proxy to job templates

adiabramovitch:
related to https://github.com/theforeman/foreman_ansible/pull/457

### DIFF
--- a/app/services/foreman_ansible/playbooks_importer.rb
+++ b/app/services/foreman_ansible/playbooks_importer.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  # Imports playbooks from smart proxy
+  class PlaybooksImporter
+    include ::ForemanAnsible::ProxyAPI
+    delegate :playbooks, :playbooks_names, :to => :proxy_api
+
+    def initialize(proxy = nil)
+      @ansible_proxy = proxy
+    end
+
+    def import_playbooks(playbooks_names)
+      playbooks = playbooks(playbooks_names)
+      result = { created: {}, updated: {} }
+      playbooks.each do |playbook|
+        parsed_playbook = parse_playbook playbook
+        job_template_id = JobTemplate.where(name: parsed_playbook[:name]).pick(:id)
+        if job_template_id.present?
+          updated = update_job_template(job_template_id, parsed_playbook)
+          result[:updated].merge!(updated) unless updated.nil?
+        else
+          result[:created].merge!(create_job_template(parsed_playbook))
+        end
+      end
+      result
+    end
+
+    def parse_playbook(playbook)
+      content = playbook['playbooks_content']
+      {
+        name: playbook['name'],
+        playbook_content: metadata(playbook['name']) + content,
+        vars: get_vars(content)
+      }
+    end
+
+    def metadata(playbook_name)
+      <<~END_HEREDOC
+        <%#
+          name: #{playbook_name}
+          snippet: false
+          job_category: Ansible Playbook - Imported
+          provider_type: Ansible
+          kind: job_template
+          model: JobTemplate
+        %>
+      END_HEREDOC
+    end
+
+    def get_vars(playbook_content)
+      YAML.safe_load(playbook_content).map { |play| play['vars'] }
+    end
+
+    def create_job_template(playbook)
+      job_template = JobTemplate.create(name: playbook[:name], template: playbook[:playbook_content], job_category: 'Ansible Playbook - Imported', provider_type: 'Ansible')
+      # TODO: Add support for creating template inputs
+      job_template.save
+      { job_template.id => job_template.name }
+    end
+
+    def update_job_template(job_template_id, playbook)
+      # TODO: Add support for updating template inputs
+      inputs = []
+      job_template = JobTemplate.find(job_template_id)
+      should_update = !playbook[:playbook_content].eql?(job_template.template)
+      job_template.template = playbook[:playbook_content] if should_update
+      { job_template.id => job_template.name } unless inputs.empty? && !should_update
+    end
+  end
+end
--- a/test/fixtures/sample_playbooks.json
+++ b/test/fixtures/sample_playbooks.json
@@ -1,0 +1,10 @@
+[
+  {
+    "name": "xprazak2.forklift_collection.foreman_provisioning.yml",
+    "playbooks_content": "- hosts: all\n  become: true\n  vars:\n    libvirt_tftp: true\n  roles:\n    - foreman\n    - libvirt\n    - foreman_provisioning\n"
+  },
+  {
+    "name": "xprazak2.forklift_collection.collect_debug_draft.yml",
+    "playbooks_content": "---\n- hosts: all\n  become: true\n  vars:\n    bats_output_dir: '/root/bats_results'\n    remote_dir: \"/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}\"\n  roles:\n    - sos_report\n  tasks:\n    - name: \"Find bats files\"\n      find:\n        paths: \"{{ bats_output_dir }}\"\n        patterns: \"*.tap\"\n      register: bats_results\n\n    - name: \"Copy bats results\"\n      fetch:\n        src: \"{{ item.path }}\"\n        dest: \"{{ remote_dir }}\"\n      with_items: \"{{ bats_results.files }}\"\n\n    - name: \"Find smoker files\"\n      find:\n        paths: \"{{ smoker_output_dir }}\"\n        patterns: \"*.xml\"\n      register: smoker_results\n\n    - name: \"Copy smoker results\"\n      fetch:\n        src: \"{{ item.path }}\"\n        dest: \"{{ remote_dir }}\"\n      with_items: \"{{ smoker_results.files }}\"\n"
+  }
+]
--- a/test/unit/import_playbooks_test.rb
+++ b/test/unit/import_playbooks_test.rb
@@ -1,0 +1,51 @@
+require 'test_plugin_helper'
+
+class ImportPlaybooksTest < ActiveSupport::TestCase
+  let(:ansible_proxy) { FactoryBot.create(:smart_proxy, :with_ansible) }
+  let(:proxy_api) { ::ProxyAPI::Ansible.new(url: ansible_proxy.url) }
+  subject { ::ForemanAnsible::PlaybooksImporter.new(ansible_proxy) }
+  let(:sample_playbooks) { JSON.parse(File.read(ansible_fixture_file('sample_playbooks.json'))) }
+  let(:playbook1_output) { sample_playbooks.first }
+  let(:playbook2_output) { sample_playbooks.last }
+  let(:first_playbook) { ['xprazak2.forklift_collection.foreman_provisioning.yml'] }
+  let(:second_playbook) { ['xprazak2.forklift_collection.collect_debug_draft.yml'] }
+  let(:playbook1_template) { JSON.parse(File.read(ansible_fixture_file('playbooks_example_output.json')))['template'] }
+
+  setup do
+    subject.stubs(:proxy_api).returns(proxy_api)
+  end
+
+  describe 'check the job templates that were created due to the import' do
+    test 'should just create job templates from playbooks' do
+      subject.expects(:playbooks).with(second_playbook).returns([playbook2_output])
+      actual = subject.import_playbooks(second_playbook)
+      assert_not_nil actual_created = actual[:created]
+      assert_equal 1, actual_created.count
+      job_template = JobTemplate.where(id: actual_created.keys.first).first
+      assert_equal job_template.name, actual_created.values.first
+    end
+  end
+
+  describe 'check the updated job templated due to import playbooks' do
+    test ' should not update an identical job template' do
+      FactoryBot.create(:job_template, id: 111, name: playbook1_output['name'],
+                                       template: playbook1_template)
+      subject.expects(:playbooks).with(first_playbook).returns([playbook1_output])
+      actual = subject.import_playbooks(first_playbook)
+      assert_not_nil actual_updated =  actual[:updated]
+      assert_not_nil actual_created =  actual[:created]
+      assert_equal 0, actual_updated.count
+      assert_equal 0, actual_created.count
+    end
+
+    test ' should just update a job template from playbooks' do
+      job_template = FactoryBot.create(:job_template, id: 112, name: playbook2_output['name'], template: 'check')
+      subject.expects(:playbooks).with(second_playbook).returns([playbook2_output])
+      actual = subject.import_playbooks(second_playbook)
+      assert_not_nil actual_updated = actual[:updated]
+      assert_equal 1, actual_updated.count
+      assert_equal actual_updated.keys.first, job_template.id
+      assert_not_nil job_template.template
+    end
+  end
+end
--- a/test/fixtures/playbooks_example_output.json
+++ b/test/fixtures/playbooks_example_output.json
@@ -1,0 +1,1 @@
+{ "template": "<%#\n  name: xprazak2.forklift_collection.foreman_provisioning.yml\n  snippet: false\n  job_category: Ansible Playbook - Imported\n  provider_type: Ansible\n  kind: job_template\n  model: JobTemplate\n%>\n- hosts: all\n  become: true\n  vars:\n    libvirt_tftp: true\n  roles:\n    - foreman\n    - libvirt\n    - foreman_provisioning\n"}


test/unit/import_playbooks_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    @playbooks_names = ['xprazak2.forklift_collection.collect_debug_draft.yml']
+    @importer = mock('Playbook Importer')
+    @proxy_api = mock('ProxyAPI')
+    @proxy = FactoryBot.create(:smart_proxy, :with_ansible)
+    @params_create = OutputExample.params_create
+    @params_update_no_inputs = OutputExample.params_update_no_inputs
+    @params_update = OutputExample.params_update
+    sample_playbooks = JSON.parse(File.read(ansible_fixture_file('sample_playbooks.json')))
+    @playbook1_output = sample_playbooks['1']
+    @playbook2_output = sample_playbooks['2']
```
ezr-ondrej, line 17: These are used in just few tests each, but initialized for every test (as setup is run for all of them).
It would be better to actually eighter:

1. initialize these just in the tests that are using them or,
2. use `let(:param1) { param_definition }` syntax, that initializes the parameters as local variables once called eg:
```
let(:sample_playbooks) { JSON.parse(File.read(ansible_fixture_file('sample_playbooks.json'))) }
let(:playbook1_output { sample_playbooks['1'] }

test '' do
  expects(playbook1_output).to be_kind_of(Hash)
end

File: test/unit/import_playbooks_test.rb [hunk 2]
```diff
@@ -0,0 +1,74 @@
+require 'test_plugin_helper'
+require_relative '../../app/services/foreman_ansible/playbooks_importer'
+require_relative '../fixtures/output_example_from_playbooks_importer'
+
+class ImportPlaybooksTest < ActiveSupport::TestCase
+  setup do
+    @actual_importer = ::ForemanAnsible::PlaybooksImporter.new(@proxy)
```
ezr-ondrej, line 7: this could be initialized outside of the setup block in a special `subject` block I'd imagine it something like:
```ruby
let(:ansible_proxy) { FactoryBot.create(:smart_proxy, :with_ansible) }
let(:proxy_api) { mock('ProxyAPI') } # this could even be a real thing by { ::ProxyAPI::Ansible.new(url: ansible_proxy.url }
subject { ::ForemanAnsible::PlaybooksImporter.new(ansible_proxy) }

setup do
  subject.stubs(:proxy_api).returns(proxy_api)
end

test '1' do
  proxy_api.stubs(:playbooks).returns([1,2]) # you probably want to return something useful ^_^
  assert_equal subject.playbooks.first, 1 # you probably want to have more elaborate test ^_^
  # .... test stuff..
end
```

test/unit/import_playbooks_test.rb
```diff
@@ -0,0 +1,74 @@
+require 'test_plugin_helper'
+require_relative '../../app/services/foreman_ansible/playbooks_importer'
+require_relative '../fixtures/output_example_from_playbooks_importer'
```
ezr-ondrej, line 3: this should work:
```suggestion
require_dependency 'foreman_ansible/playbooks_importer' # IMO omitting this would also work as it should get auto required :)
require 'fixtures/output_example_from_playbooks_importer'
```

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,131 @@
...
+    def import_playbooks(playbooks_names)
+      import_playbooks_content playbooks(playbooks_names)
+    end
+
+    def import_playbooks_content(playbooks)
+      params = { changed: { update: [], create: [] } }
+      playbooks.each do |playbook|
+        new_playbook = parse_playbook playbook
+        job_template_id = JobTemplate.where(name: new_playbook[:name]).pluck(:id).first
+        old_var_block = PlaybooksImporterHelper.check_vars_block new_playbook[:old_vars_block]
```
ezr-ondrej, line 22: this naming is quite strange. Why do we have `old_var_block` is there a new one as well? why don't we have just `vars`?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,131 @@
...
+        '  job_category: Ansible Playbook - Imported',
+        '  provider_type: Ansible',
+        '  kind: job_template',
+        '  model: JobTemplate',
+        "%>\n"
+      ].join("\n")
+    end
+
+    def convert_to_job_template_input(playbook_content)
+      playbook_content_as_json = YAML.safe_load(playbook_content.join('')).first
```
ezr-ondrej, line 69: This looks like we have an input format for playbook_content to be rows strings with the new lines in an array. Why? Why don't we have the full string instead as that would feel like more standardized way to pass that arround (especially over the wire)?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,131 @@
...
+        '<%#',
+        "  name: #{playbook_name}",
+        '  snippet: false',
+        '  job_category: Ansible Playbook - Imported',
+        '  provider_type: Ansible',
+        '  kind: job_template',
+        '  model: JobTemplate',
+        "%>\n"
+      ].join("\n")
+    end
```
ezr-ondrej, line 66: I do not think we need to add these, these are usefull in the files, there these provide a way to read the metadata, but we can feed the metadata directly.

app/services/foreman_ansible/playbooks_importer.rb
ezr-ondrej, line 69: The use of first is wrong here (because playbook is an list of plays, so one playbook can have multiple plays and you can't just ignore all but first :)

ezr-ondrej (changes requested):
This is quite complex for what it tries to achieve, I've tried to comment on the main pain points of the solution, but I'll say in general that I quite like the idea of processing variables to inputs the way you're trying to do, but it is IMHO unecessarily complex and I believe it could be simplified. I was hoping to get the import done correctly first and then we could get to fancy stuff (like processing variables). That was what confused me in the first PR, how it can be so complex now I see you try to go over and beyond in one PR and that's usually not a good idea :)
Could we first get just the playbooks content simply dumped into the database as a job template? I believe we would find some design flaws and could simplify that before we add the whole variable complexity. By separating, it would be easier to simplify.

app/services/foreman_ansible/playbooks_importer.rb
adiabramovitch: This is the way the playbooks are stored in smart proxy:
```
{"name"=>"xprazak2.forklift_collection.collect_debug_draft.yml",
 "playbooks_content"=>
  ["---\n",
   "- hosts: all\n",
   "  become: true\n",
   "  vars:\n",
   "    bats_output_dir: '/root/bats_results'\n",
   "    remote_dir: \"/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}\"\n",
   "  roles:\n",
   "    - sos_report\n",
   "  tasks:\n",
   "    - name: \"Find bats files\"\n",
   "      find:\n",
   "        paths: \"{{ bats_output_dir }}\"\n",
   "        patterns: \"*.tap\"\n",
   "      register: bats_results\n",
   "\n",
   "    - name: \"Copy bats results\"\n",
   "      fetch:\n",
   "        src: \"{{ item.path }}\"\n",
   "        dest: \"{{ remote_dir }}\"\n",
   "      with_items: \"{{ bats_results.files }}\"\n",
   "\n",
   "    - name: \"Find smoker files\"\n",
   "      find:\n",
   "        paths: \"{{ smoker_output_dir }}\"\n",
   "        patterns: \"*.xml\"\n",
   "      register: smoker_results\n",
   "\n",
   "    - name: \"Copy smoker results\"\n",
   "      fetch:\n",
   "        src: \"{{ item.path }}\"\n",
   "        dest: \"{{ remote_dir }}\"\n",
   "      with_items: \"{{ smoker_results.files }}\"\n"]}
```

In proxy we get playbooks_content from:
```
   playbooks_content: File.readlines(path)
   ```
   Do you think it should be a different way?

File: app/services/foreman_ansible/playbooks_importer.rb (hunk 6 above)
Comment (adiabramovitch) [reply app/services/foreman_ansible/playbooks_importer.rb:66]: what do you mean by feeding the metadata directly? when creating the Jobtemplate itself?

File: app/services/foreman_ansible/playbooks_importer.rb (hunk 5 above)
Comment (ezr-ondrej) [reply app/services/foreman_ansible/playbooks_importer.rb:69]: yeah, I believe it would be better to send it as output of `File.read(path)` instead if possible (just because it is more reliable to send strings then arrays and we are joining it all together on the foreman end anyway).
It is not a major issue tho.

Comment (adiabramovitch) [reply app/services/foreman_ansible/playbooks_importer.rb:69]: so I'll move the join part into proxy in a different PR?

File: app/services/foreman_ansible/playbooks_importer.rb [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+        '<%#',
+        "  name: #{playbook[:name]}",
+        '  snippet: false',
+        '  job_category: Ansible Playbook - Imported',
+        '  provider_type: Ansible',
+        '  kind: job_template',
+        '  model: JobTemplate',
+        "%>\n"
+      ] + playbook[:playbook_content]
+    end
```
ezr-ondrej, line 61: I'm still quite sure these are not needed here :)

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def get_vars(playbook_content)
+      YAML.safe_load(playbook_content.join('')).each { |play| play['vars'] }
+    end
```
ezr-ondrej, line 65: this is not yet needed, right?

app/services/foreman_ansible/playbooks_importer_helper.rb
```diff
@@ -0,0 +1,14 @@
+module ForemanAnsible
+  class PlaybooksImporterHelper
```
ezr-ondrej, line 2: I do not think we need this class just yet :)

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def import_playbooks(playbooks_names)
+      import_playbooks_content playbooks(playbooks_names)
+    end
+
+    def import_playbooks_content(playbooks)
+      params = { changed: { update: [], create: [] } }
+      playbooks.each do |playbook|
+        new_playbook = parse_playbook playbook
+        job_template_id = JobTemplate.where(name: new_playbook[:name]).pluck(:id).first
```
ezr-ondrej, line 21: There is a new method in Rails 6 for this:
```suggestion
        job_template_id = JobTemplate.where(name: new_playbook[:name]).pick(:id)
```

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,79 @@
...
+  class PlaybooksImporter
+    include ::ForemanAnsible::ProxyAPI
+    delegate :playbooks, :playbooks_names, :to => :proxy_api
+
+    def initialize(proxy = nil)
+      @ansible_proxy = proxy
+    end
+
+    def import_playbooks(playbooks_names)
+      import_playbooks_content playbooks(playbooks_names)
```
ezr-ondrej, line 14: This step I'm bit curious about. I guess that we want to phase it into this playbook will be updated, this created, this deleted, but you're already fetching the content of the playbooks for that .
Is it necessary given we operate only on names (and I don't believe we want to do anything else in the future anyway)?

I think it would be better to sort out update/create first with names only and only when we have that sorted (and selected by some input I'd expect to have method that would get names to update, names to create and did what it's told.

ezr-ondrej (commented):
I belive that we are tryting to have phases for import as we do for the variables, but I believe that we are not doing it properly.

Phases IMHO are:

- fetch available playbook names from proxy
- divide those into existing internally and non existing internally
- on input `import: <playbook_names>[, delete: <playbook_names>]` (where delete is optional at this point) load the content for the imported ones and only then import them.

Good excersise of TTD would be to describe these phases firt in tests (have a describe block per phase and test for what the phase should do) and then implement the failing tests :)

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,76 @@
...
+      # TODO: Add support for updating template inputs
+      inputs = []
+      job_template = JobTemplate.find(job_template_id)
+      should_update = !playbook[:playbook_content].eql?(job_template.template)
+      job_template.template = playbook[:playbook_content] if should_update
+      { job_template.id => job_template.name } unless inputs.empty? && !should_update
+    end
+
+    def delete_job_templates(playbooks_names)
+      JobTemplate.where(job_category: 'Ansible Playbook - Imported').where.not(name: playbooks_names).destroy_all.map { |template| { template.id => template.name } } unless playbooks_names.empty?
```
ezr-ondrej, line 73: This sounds dangerous. Do we want to delete all templates that are not explicitly imported?

What if I import some from one proxy and then I import another from different proxy, the second import would delete all the playbooks from the first import, wouldn't it?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+      result[:deleted] = delete_job_templates(playbooks_names)
+      result
+    end
+
+    def parse_playbook(playbook)
+      content = playbook['playbooks_content'].join('')
+      {
+        name: playbook['name'],
+        playbook_content: add_metadata(playbook['name']) + content,
```
ezr-ondrej, line 34: I still don't get why we are adding the metadata, those are useless IMHO.
But it won't hurt eighter I guess :)

app/services/foreman_ansible/playbooks_importer.rb
adiabramovitch: Yes, you are right. 
It was the logic a while ago, but I agree it has some troubles. 
When should we delete job templates?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,76 @@
...
+      [
+        '<%#',
+        "  name: #{playbook_name}",
+        '  snippet: false',
+        '  job_category: Ansible Playbook - Imported',
+        '  provider_type: Ansible',
+        '  kind: job_template',
+        '  model: JobTemplate',
+        "%>\n"
+      ].join('')
```
ezr-ondrej, line 49: Shouldn't we join with `"\n"`? Optionally, we could use [HEREDOC syntax](http://graysoftinc.com/ruby-voodoo/working-with-multiline-strings)

test/unit/import_playbooks_test.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:ansible_proxy) { FactoryBot.create(:smart_proxy, :with_ansible) }
+  let(:proxy_api) { ::ProxyAPI::Ansible.new(url: ansible_proxy.url) }
+  let(:playbooks_names) { ['xprazak2.forklift_collection.collect_debug_draft.yml'] }
+  subject { ::ForemanAnsible::PlaybooksImporter.new(ansible_proxy) }
+
+  setup do
+    subject.stubs(:proxy_api).returns(proxy_api)
+    let(:sample_playbooks) { JSON.parse(File.read(ansible_fixture_file('sample_playbooks.json'))) }
+    let(:playbook1_output) { sample_playbooks['1'] }
+    let(:playbook2_output) { sample_playbooks['2'] }
```
ezr-ondrej, line 13: `let` is lazy_loaded (initialized on the first use) so it is weird to put it in the `setup` block. it should be outside of the block with the other `let` clauses

ezr-ondrej (commented):
It's getting in a quite good shape ^_^

app/services/foreman_ansible/playbooks_importer.rb
ezr-ondrej: I believe there should eighter be a special input for deleted playbooks, or there should be a separate method for deleting them. Optionally, you can just not implement the delete at all (users are able to delete the playbooks from the job_templates directly, so the need to implement it on import is not quite there :))

test/fixtures/sample_playbooks.json
```diff
@@ -0,0 +1,52 @@
+{
```
ezr-ondrej, line 1: Is this format of fixtures given by something? It is weird to have numeric keys, given we are not using that format anywhere else.
I'd prefer to have the format that is given by the foreman_proxy here (which I think is an array of playbooks, right?

test/unit/import_playbooks_test.rb
```diff
@@ -0,0 +1,50 @@
...
+      assert_equal 1, actual_created.count
+      job_template = JobTemplate.where(id: actual_created.keys.first).first
+      assert_equal job_template.name, actual_created.values.first
+    end
+  end
+
+  describe 'check the updated job templated due to import playbooks' do
+    test ' should not update an identical job template' do
+      FactoryBot.create(:job_template, id: 111, name: 'xprazak2.forklift_collection.foreman_provisioning.yml',
+                                       template: "      <%#\n        name: xprazak2.forklift_collection.foreman_provisioning.yml\n        snippet: false\n        job_category: Ansible Playbook - Imported\n        provider_type: Ansible\n        kind: job_template\n        model: JobTemplate\n      %>\n- hosts: all\n  become: true\n  vars:\n    libvirt_tftp: true\n  roles:\n    - foreman\n    - libvirt\n    - foreman_provisioning\n")
```
ezr-ondrej, line 31: can't we get this data (`name` and `template`) from the fixtures?

ezr-ondrej (commented):
I still feel we should update the proxy to send the playbook content as full string as that would be much less error prone :)

test/fixtures/sample_playbooks.json
adiabramovitch: sorry for that, forgot to change it on this branch :D 
I uploaded the new changes, hope it is OK now.

test/unit/import_playbooks_test.rb
adiabramovitch: I added a new file for the template because in the near future we will need to add more exampled output for checking the inputs/variables

adiabramovitch:
[test foreman_ansible]

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,71 @@
...
+      content = playbook['playbooks_content']
+      {
+        name: playbook['name'],
+        playbook_content: add_metadata(playbook['name']) + content,
+        vars: get_vars(content)
+      }
+    end
+
+    def add_metadata(playbook_name)
+      <<END_HEREDOC
```
ezr-ondrej, line 39: The best here would be 
```suggestion
      <<~END_HEREDOC
```
this version allows the closing to be indented, but the result is not indented, which is what we want here IMHO ?

app/services/foreman_ansible/playbooks_importer.rb
```diff
@@ -0,0 +1,71 @@
...
+    def parse_playbook(playbook)
+      content = playbook['playbooks_content']
+      {
+        name: playbook['name'],
+        playbook_content: add_metadata(playbook['name']) + content,
+        vars: get_vars(content)
+      }
+    end
+
+    def add_metadata(playbook_name)
```
ezr-ondrej, line 38: this is now not adding the metadata, so I believe the name should be just `metadata` or `metadata_for`

ezr-ondrej (approved):
I've few nits, but it seems fine now :+1:
Could we have a PR that would change the output of the smart_proxy `playbooks` endpoint (now it returns an array, but we now expect the string, so we need to change it, right?

adiabramovitch:
> I've few nits, but it seems fine now +1 Could we have a PR that would change the output of the smart_proxy `playbooks` endpoint (now it returns an array, but we now expect the string, so we need to change it, right?

Thank you for the review.
opened a PR for the playbooks endpoint fix https://github.com/theforeman/smart_proxy_ansible/pull/62 
and fixed your nitpicks :) .

ezr-ondrej (approved):
Thanks @adiabramovitch ! :+1:

PR #508, opened 2022-01-28
Fixes #34334 - Import roles - Fix pagination

stejskalleos:


webpack/components/AnsibleRolesAndVariables/AnsibleRolesAndVariables.js
```diff
@@ -67,43 +70,23 @@ const ImportRolesAndVariablesTable = ({
...
-      perPageOptions={[
-        { title: '3', value: 3 },
-        { title: '5', value: 5 },
-        { title: '10', value: 10 },
-      ]}
-      titles={{
-        paginationTitle: `${variant} pagination`,
-      }}
+      perPage={pagination.per_page}
+      pagination={pagination}
```
ezr-ondrej, line 88: I don't think there is any prop `pagination` should we keep the `page={pagination.page}` prop instead?

ezr-ondrej: Might not be needed actually, because the pagination component remembers by itself, but would be safer :shrug:

stejskalleos: Yop, not needed, I removed it

ezr-ondrej (approved):
Thanks @stejskalleos ! :+1:

PR #511, opened 2022-02-17
Fixes #34458 - Show Hostgroup jobs on the Host Detail page

stejskalleos:


webpack/components/AnsibleHostDetail/components/JobsTab/JobsTabHelper.js
```diff
@@ -7,21 +7,33 @@ import { showToast } from '../../../../toastHelper';
...
-  `recurring = true && pattern_template_name = "Ansible Roles - Ansible Default" && ${statusSearch} && recurring_logic.purpose = ${ansiblePurpose(
+const jobSearch = (resourceName, resourceId, status, hostGroupId) => {
+  const search = `recurring = true && pattern_template_name = "Ansible Roles - Ansible Default"`;
+  const searchStatus = ` && ${status}`;
+  const searchHost = ` && recurring_logic.purpose = ${ansiblePurpose(
     resourceName,
     resourceId
   )}`;
+  const searchHostGroup = hostGroupId
+    ? ` or recurring_logic.purpose = "ansible-hostgroup-${hostGroupId}"`
```
ezr-ondrej, line 18: Can we use the helper? IMHO it would help us flag places that needs to be changed if we change the purpose in the future.
```suggestion
    ? ` or recurring_logic.purpose = "ansible-ansiblePurpose('hostgroup', hostGroupId)"
```

webpack/components/AnsibleHostDetail/components/JobsTab/JobsTabHelper.js
```diff
@@ -97,3 +109,6 @@ export const readableCron = (cron = '') => {
 
   return 'custom';
 };
+
+export const readablePurpose = (purpose = '') =>
+  purpose.match(/hostgroup/) ? __('(from host group)') : '';
```
ezr-ondrej, line 114: This will be called on the host and hostgroup page, right? Can we do it somewhat explicit?
This works, it's just ambigous to have `from host group` on the hostgroup page.

ezr-ondrej (commented):
This does not deal with nested Hostgroups, so it is not a complete solution, but it is definitelly worth merging ASIS (even without dealing with nested hostgroups) :+1:
I've few nits, but apart of that it LGTM :+1:

stejskalleos:
> This does not deal with nested Hostgroups, so it is not a complete solution, but it is definitelly worth merging ASIS (even without dealing with nested hostgroups) +1

Fixed the nits except the problem with inherited host group. It is not problem in the Ansible but in the Foreman core itself, if you try to search `recurring_logic.purpose = ansible-hostgroup-<id-of-inherited-group>` on `/job_invocation` page you won't get any results neither

ezr-ondrej (approved):
Definitelly a move in the right direction :+1:
Thanks @stejskalleos !

PR #512, opened 2022-02-24
Fixes #34512 - coerce batch size to integer

ezr-ondrej:
The setting is not coerced properly, we need to make sure it is coerced properly.


stejskalleos (approved):
:heavy_check_mark: Works well, LGTM!

PR #517, opened 2022-03-11
Dont update setting defaults

ezr-ondrej:
Updating setting defaults manually is no longer necessary
Would break after theforeman/foreman#9050


adamruzicka (approved):

adamruzicka:
Thank you @ezr-ondrej & @jeremylenz !

PR #520, opened 2022-03-14
Extended type should be string

ezr-ondrej:
Do not autoload the constant prematurely.
It can cause world of trouble if we require the constant too early.
The graphql registration supports string types, we should leverage that.


ezr-ondrej:
Should work, will self-merge so this gets tested properly prior next release.

PR #521, opened 2022-03-21
Fixes #34652 - added git ansible module instead of command

snagoor:


app/views/foreman_ansible/job_templates/ansible_roles_-_install_from_git.erb
```diff
@@ -23,6 +23,9 @@ model: JobTemplate
 ---
 - hosts: all
   tasks:
-    - command: git clone <%= input('git_repository') %> <%= input('location') %>
+    - git:
+        repo: <%= input('git_repository') %>
```
ezr-ondrej, line 27: Would it be safer to add quotes? just as additional safety measure, unrelated to the actual change.
```suggestion
        repo: "<%= input('git_repository') %>"
```

ezr-ondrej (approved):
Make sense, thanks a bunch @snagoor ! :)
I have just one improvement idea, but I definitelly do not insist, this is ok to merge without it.

app/views/foreman_ansible/job_templates/ansible_roles_-_install_from_git.erb
snagoor: @ezr-ondrej agreed. Modified accordingly.

ezr-ondrej (approved):
Thanks @snagoor ! Nice addition :+1:

PR #522, opened 2022-03-30
Fixes #34701 - Make running roles from hostgroup index work again

adamruzicka:
The route to /ansible/hostgroups/$HG_ID/play_roles was shadowed by a
mapping to react controller. This way, the more specific routes are
matched first before the react controller which catches everything.


ezr-ondrej (approved):
:facepalm: Thanks @adamruzicka !

ezr-ondrej:
CP 7.1: #523

PR #525, opened 2022-04-11
Refs #34121 - Add inherited Ansible role flag to host(group) API

ofedoren:
Needed for https://github.com/theforeman/hammer-cli-foreman-ansible/pull/10.

Unfortunatelly, I needed to extract host(group)'s [all_ansible_roles](https://github.com/theforeman/foreman_ansible/blob/master/app/models/concerns/foreman_ansible/host_managed_extensions.rb#L60-L62) method logic into the controller to get inherited ones only to be used in the RABL template, so we don't need to call this for each role. I din't want to call `inherited_ansible_roles` method logic twice since it would call the DB twice. But please tell me if there is a better way since this looks quite ugly :/

### DIFF
--- a/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
@@ -46,7 +46,10 @@ module ForemanAnsible
             find_resource
             return unless @hostgroup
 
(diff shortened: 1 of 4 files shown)


stejskalleos (approved):
:heavy_check_mark: **LGTM**, works well!

stejskalleos:
Thanks @ofedoren !

PR #526, opened 2022-04-12
Fixes #34763 - Drop Cloud Connector job template

ehelms:
Moving to foreman_rh_cloud plugin

See https://github.com/theforeman/foreman_rh_cloud/pull/719


adamruzicka:
If we're dropping the template from f-a, does it make sense to still carry the feature here https://github.com/theforeman/foreman_ansible/blob/master/lib/foreman_ansible/remote_execution.rb#L51 ?

adamruzicka (approved):
`+0 -43`, more PRs like this please

adamruzicka:
Thank you @ehelms !

PR #528, opened 2022-04-14
Refs #34121 - Add owned Ansible role flag to host(group) API

ofedoren:
I'm dumb and I didn't notice that inherited and directly assigned roles can intersect :/ UI doesn't allow that, but via API you can assign a role to a host/group even though it is inherited.

Nevertheless, I guess this is still handy to have.

P.S. Please don't merge just yet untill we agree with @adamruzicka. Just in case there will be a need in other changes :)


app/views/api/v2/hostgroups/ansible_roles.json.rabl
```diff
@@ -7,3 +7,7 @@ extends 'api/v2/ansible_roles/show'
 node :inherited do |role|
   @inherited_ansible_roles.include?(role)
 end
+
+node :owned do |role|
```
ezr-ondrej, line 11: might `directly_assigned` or `own` be better names? :thinking:

ofedoren: I thought about these two and I have no idea which one is more straightforward for the users. `directly_assigned` is more descriptive for me and I'd choose this, but it sounds like something more internal / for devs. Thus I've chosen `own`, but with the same form as with `inherited`.

Naming is hard :face_exhaling:

adamruzicka: I'd go with `directly_assigned`. It might be a slightly more technical name, but I'd say it is warranted here

ofedoren:
Thanks, @ezr-ondrej and @adamruzicka for the suggestion, updated.

adamruzicka (approved):
Works as expected, ACK

PR #529, opened 2022-04-21
Fixes #34796 - Fixes for Rails 6.1 upgrade

stejskalleos:
For https://github.com/theforeman/foreman/pull/9117

Changes:
- loading of host extension helper
- Rails 6 'where.not' use NAND instead of NOR [0]

[0] https://github.com/rails/rails/pull/36029


lib/foreman_ansible/engine.rb
```diff
@@ -29,6 +29,12 @@ class Engine < ::Rails::Engine
 
     initializer 'foreman_ansible.register_plugin', :before => :finisher_hook do
       require 'foreman_ansible/register'
+
+      Foreman::Plugin.register :foreman_ansible do
```
ezr-ondrej, line 33: this is already in `lib/foreman_ansible/register.rb` file, we should not add another registration block, it could do bonker stuff.

ezr-ondrej (commented):
:heart: this, just one comment.

adamruzicka (approved):

adamruzicka:
Thank you @stejskalleos & @ezr-ondrej !

PR #531, opened 2022-05-10
Fixes #34884 - Host detail and roles from host group

stejskalleos:
From [community post](https://community.theforeman.org/t/new-ui-does-not-display-assigned-roles/27897/5)

When host have roles only from the host group,
table with roles on the new host detail page ignores them.

**Changes:**
* Fix issue when host had assigned roles via host group only
* Remove "All roles" modal and display all roles in the table

**Notes:**
* Reloading data in the table after assigning / unassigning roles still doesn't work, but that's a story for another [redmine](https://projects.theforeman.org/issues/34885)


adamruzicka (approved):
Fixes the issue for me. Any idea why we had the modal in first place?

stejskalleos:
> Fixes the issue for me. Any idea why we had the modal in first place?

I don't know TBH, I would ask Ondra but he's not around anymore

ezr-ondrej:
> Fixes the issue for me. Any idea why we had the modal in first place?

It was missunderstandment between UX and Dev ^_^ UX wanted to have a way how to see roles assigned to the host directly and modal was how engineering translated that request :)

Classical agile developement: https://blogs.perficient.com/2011/07/22/how-to-build-a-tire-swing-a-case-for-agile-development/

What is implemented here by @stejskalleos is what UX actually asked for (well how I understand it anyway). So big :+1: from me :)

webpack/components/AnsibleHostDetail/components/RolesTab/RolesTable.js
```diff
@@ -27,7 +26,7 @@ const RolesTable = ({
   hostGlobalId,
   canEditHost,
 }) => {
-  const columns = [__('Name'), __('Variables')];
+  const columns = [__('Name'), __('Variables'), __('Source')];
```
ezr-ondrej, line 29: Maybe source is bit missleading, but I do not have a better word for you, so just pointing out in case someone would have better idea ^_^

stejskalleos: Yeah, I spent more time on this than I want to admin and still couldn't came up with anything better than `Source`. But I'm open to suggestions and ideas!

stejskalleos:
> EDIT: the actual request from UX was to have a switcher, that would hide iherited roles something along the lines of Show inherited roles, so this is only stepping stone for the final solution, but hell of a stepping stone if you ask me ^_^

RM for the switcher: https://projects.theforeman.org/issues/34907

webpack/components/AnsibleHostDetail/components/RolesTab/RolesTable.js
Ron-Lavi: not sure if it's better, how about `'Inherited from host group'`? 

and later on:
```js
<Td>{ role.inherited ? <CheckIcon /> : null }</Td>
```

Ron-Lavi (approved):
Fixed the issue for me, thanks @stejskalleos!

stejskalleos:
PR cause bug with not reloading data in table, moving to draft

ares:
@stejskalleos any update on this one? sad to see this one closed as is was quite close

stejskalleos:
> any update on this one? sad to see this one closed as is was quite close

It was quite close, but I wasn't able to fix the issue described above, not sure where the problem was

PR #533, opened 2022-05-12
Fixes #34911 - prevent reset to initial roles values in EditRolesForm

Ron-Lavi:
after a redux store update, the useEffect in the
EditRolesForm caused to reset back to the initial values.


stejskalleos:
:heavy_check_mark: **LGTM**, works as expected.

Thanks @Ron-Lavi !

PR #544, opened 2022-07-13
Update theforeman dev dependencies to version ^10.0

ekohl:
These are the latest versions. Packaging already uses these versions.

I have no clue if this works, just submitting it to see what CI thinks. This what RPMs run with in production so if it fails here, it will probably also fail in production.


ekohl:
Derp, it was `< 11` in our packaging and I copied that. Updated to 10.

ekohl:
Ruby tests fail due to Rails 6.1.6.1.

package.json
```diff
@@ -15,12 +15,12 @@
...
-    "@theforeman/find-foreman": "^8.4.1",
-    "@theforeman/stories": "^8.4.1",
-    "@theforeman/test": "^8.9.0",
-    "@theforeman/vendor-dev": "^8.4.1",
+    "@theforeman/builder": "^10.0",
+    "@theforeman/eslint-plugin-foreman": "^10.0",
+    "@theforeman/find-foreman": "^10.0",
+    "@theforeman/stories": "^10.0",
+    "@theforeman/test": "^10.0",
+    "@theforeman/vendor-dev": "^10.0",
```
Ron-Lavi, line 23: On other plugins we started to accept `>= 0` for these versions, to reduce the number of changes for plugins:
```suggestion
    "@theforeman/builder": ">= 0",
    "@theforeman/eslint-plugin-foreman": ">= 0",
    "@theforeman/find-foreman": ">= 0",
    "@theforeman/stories": ">= 0",
    "@theforeman/test": ">= 0",
```
btw, `@theforeman/vendor-dev` isn't needed.

same for the peerDeps:
```json
"peerDependencies": {
  "@theforeman/vendor": ">= 0"
},
```

reference: https://github.com/theforeman/foreman_discovery/blob/develop/package.json#L28

ekohl: I don't have time to finish this up, but this is blocking updates. Could you have a look @Ron-Lavi?

Ron-Lavi: Hey Ewoud, I am actually on vacation and can't test it, @mariaaga do you mind taking a look? Thanks

ekohl: Then by all means, enjoy your vacation. No need to reply immediately :)

MariaAga (approved):
I'm against using `">= 0"`, so this lgtm. only breaking change between the versions was related to `react-ace` (and its tests) which foreman_ansible doesnt use.

Ron-Lavi (approved):
Thanks @ekohl, we can add the suggestions on a follow-up

PR #534, opened 2022-05-12
Fixes #34912 - enable to edit roles when all are assigned

Ron-Lavi:

PR #546, opened 2022-07-31
[CP 7.1] - Fixes #34912 - enable to edit roles when all are assigned (#534)

Ron-Lavi:
(cherry picked from commit cdadcd69a323230bb7339aca3e629b8ac8ac7870)


stejskalleos (approved):
:green_apple: **LGTM**

There is still the issue with roles from host group, we should fix that ASAP (https://projects.theforeman.org/issues/34884)

PR #553, opened 2022-08-09
Fixes #27181 - add link to hostgroup count for ansible roles

nofaralfasi:
On the Ansible Roles page, added a link to the hostgroups' counts for the ansible roles. 
This also required adding an additional searchable field of the `ansible role` for the hostgroup.

### DIFF
--- a/app/models/concerns/foreman_ansible/hostgroup_extensions.rb
+++ b/app/models/concerns/foreman_ansible/hostgroup_extensions.rb
@@ -11,6 +11,9 @@ module ForemanAnsible
                -> { order('hostgroup_ansible_roles.position ASC') },
                :through => :hostgroup_ansible_roles,
                :dependent => :destroy
+      scoped_search :relation => :ansible_roles, :on => :name,
+                    :complete_value => true, :rename => :ansible_role,
+                    :only_explicit => true, :operators => ['= ', '!= ', '~ ', '!~ '], :ext_method => :search_by_role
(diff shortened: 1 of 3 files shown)


app/models/concerns/foreman_ansible/hostgroup_extensions.rb
```diff
@@ -38,6 +41,16 @@ def all_ansible_roles
         (inherited_ansible_roles + ansible_roles + host_ansible_roles).uniq
       end
     end
+
+    class_methods do
+      def search_by_role(_key, operator, value)
+        conditions = sanitize_sql_for_conditions(["ansible_roles.name #{operator} ?", value_to_sql(operator, value)])
+        hostgroup_ids = ::Hostgroup.unscoped.with_taxonomy_scope.joins(:ansible_roles).where(conditions).map(&:subtree_ids).flatten
```
stejskalleos, line 48: `unscoped` [0] and `with_taxonomy_scope` can be skipped here, because host groups are scoped in taxonomy by default [1].

With this code we reset scope by `unscoped` and set it back with `with_taxonomy_scope`, that just doesn't make sense IMHO.

[0] https://apidock.com/rails/v6.1.3.1/ActiveRecord/Scoping/Default/ClassMethods/unscoped
[1] https://github.com/theforeman/foreman/blob/develop/app/models/hostgroup.rb#L52

stejskalleos (changes requested):

stejskalleos (approved):
:green_apple: **LGTM**, nice improvement!

Thanks @nofaralfasi

PR #556, opened 2022-08-21
Fixes #36368 - fix ansible roles pagination on host edit page

nofaralfasi:
This PR fixes the pagination issues we have with Ansible roles on the host edit page.

### DIFF
--- a/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js
+++ b/webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js
@@ -26,7 +26,6 @@ class AnsibleRolesSwitcher extends React.Component {
       inheritedRoleIds,
       resourceId,
       resourceName,
-      { page: 1, perPage: 10 },
       excludeAssignedRolesSearch(initialAssignedRoles)
     );
   }
--- a/webpack/components/AnsibleRolesSwitcher/components/AvailableRolesList.js
+++ b/webpack/components/AnsibleRolesSwitcher/components/AvailableRolesList.js
@@ -19,7 +19,9 @@ const AvailableRolesList = ({
       <Pagination
         viewType="list"
         itemCount={itemCount}
-        pagination={pagination}
+        updateParamsByUrl={false}
+        page={pagination.page}
+        perPage={pagination.perPage}
         onChange={onListingChange}
         dropdownButtonId="available-ansible-roles-pagination-row-dropdown"
       />
--- a/webpack/components/AnsibleRolesSwitcher/components/__snapshots__/AvailableRolesList.test.js.snap
+++ b/webpack/components/AnsibleRolesSwitcher/components/__snapshots__/AvailableRolesList.test.js.snap
@@ -16,14 +16,8 @@ exports[`AvailableRolesList should render 1`] = `
       onPerPageSelect={null}
       onSetPage={null}
       page={1}
-      pagination={
-        Object {
-          "page": 1,
-          "perPage": 25,
-        }
-      }
-      perPage={null}
-      updateParamsByUrl={true}
+      perPage={25}
+      updateParamsByUrl={false}
       variant="bottom"
       viewType="list"
     />


nofaralfasi:
@stejskalleos can you take a look at the failing tests? I want to make sure they are not related to my changes.

webpack/components/AnsibleRolesSwitcher/AnsibleRolesSwitcher.js
```diff
@@ -26,7 +26,7 @@ class AnsibleRolesSwitcher extends React.Component {
       inheritedRoleIds,
       resourceId,
       resourceName,
-      { page: 1, perPage: 10 },
+      { page: 1, perPage: 20 },
```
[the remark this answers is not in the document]
nofaralfasi: Maybe @girijaasoni can help me with that?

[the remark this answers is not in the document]
Ron-Lavi: imho we could use the default setting as Leos mentioned, which is already used in the Foreman's Pagination component https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/Pagination/index.js#L32

webpack/components/AnsibleRolesSwitcher/components/AvailableRolesList.js
```diff
@@ -19,7 +19,9 @@ const AvailableRolesList = ({
       <Pagination
         viewType="list"
         itemCount={itemCount}
-        pagination={pagination}
+        updateParamsByUrl={false}
+        page={pagination.page}
+        perPage={pagination.perPage}
```
Ron-Lavi, line 24: if you'll just won't pass it, it will use the default per_page from:

https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/Pagination/index.js#LL34C42-L34C54

webpack/components/AnsibleRolesSwitcher/components/AvailableRolesList.js
```diff
@@ -19,7 +19,9 @@ const AvailableRolesList = ({
       <Pagination
         viewType="list"
         itemCount={itemCount}
-        pagination={pagination}
+        updateParamsByUrl={false}
+        page={pagination.page}
```
Ron-Lavi, line 23: page is also 1 by default: https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/Pagination/index.js#L146

if you don't need to manage it in this component, maybe you can also remove it from the component's props

nofaralfasi: It doesn't work if I don't pass the pagination values, as I did here.

Ron-Lavi: Ok, I will need to debug it to see what's going on

Ron-Lavi: @liorKGOW maybe you have some time to try to debug this with Nofar?

LiorKGOW: Sure
DM me ? @nofaralfasi

webpack/components/AnsibleRolesSwitcher/components/AvailableRolesList.js
nofaralfasi: @Ron-Lavi I'm getting more confused about how this should work... I couldn't find a better solution, where I don't pass these three properties to the Pagination component.

nofaralfasi: P.S. I used the implementation [here](https://github.com/theforeman/foreman/blob/7e4d9b29f193cec9fa711e52eb7fdfc9d778937b/webpack/assets/javascripts/react_app/components/HostDetails/Tabs/Parameters/ParametersTable.js#LL100C14-L100C24) as an example @MariaAga suggested.

stejskalleos (changes requested):
It doesn't work for me. I have 23 roles assigned to the host. When I select 5 rows per page, pagination is disabled and I can't do anything with it.
ansible

nofaralfasi:
> It doesn't work for me. I have 23 roles assigned to the host. When I select 5 rows per page, pagination is disabled and I can't do anything with it. 

From what I see, this is also accruing on the master branch. I think we should open another ticket to fix that, since my changes address the host edit page, and not the new detail page (different react components).

stejskalleos:
> since my changes address the host edit page, not the new detail page (different react components).

Huh, I thought it was for the new UI. Can you create the ticket and fix the issue there too? This needs to be fixed. I will do the re-review ASAP

stejskalleos:
:green_apple: **LGTM**, works as expected.

One thing I noted is that _Assigned Ansible Roles_ doesn't have pagination at all.
Not blocking for this PR, but having same behavior for both tables would be nice.

stejskalleos:
Thanks @nofaralfasi & others!

PR #560, opened 2022-09-13
Fixes #35354 - always display the link to view all assigned roles

Ron-Lavi:
The link was hidden when there was no roles assigned directly to the host, but there are cases when the roles are inherited from the host's hostgroup and users would like to be able to view them.

The link was added in the EmptyState description.

(cherry picked from commit e69fd8611dbe3ecef2a73ddd7fd9ae7559283793)


stejskalleos (approved):
:green_apple: **LGTM** Thanks @Ron-Lavi

PR #554, opened 2022-08-15
Fixes #28374 - disable run ansible roles option from hostgroup dropdown if no roles are assigned

nofaralfasi:
Remove `Run all Ansible roles` option from hostgroup dropdown if no roles are assigned.

PR #561, opened 2022-09-20
Update REX and foreman-tasks dependencies

nofaralfasi:
Update foreman_remote_execution and foreman-tasks dependencies using `~>` instead of `>=`.

### DIFF
--- a/foreman_ansible.gemspec
+++ b/foreman_ansible.gemspec
@@ -20,6 +20,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'acts_as_list', '~> 1.0.3'
   s.add_dependency 'deface', '< 2.0'
-  s.add_dependency 'foreman_remote_execution', '>= 8.0.0'
-  s.add_dependency 'foreman-tasks', '>= 7.0.0'
+  s.add_dependency 'foreman_remote_execution', '~> 8.0'
+  s.add_dependency 'foreman-tasks', '~> 7.0'
 end


adamruzicka:
Are we expecting some breaking changes in upcoming releases of rex or tasks?

nofaralfasi:
> Are we expecting some breaking changes in upcoming releases of rex or tasks?

Not that I know of.

Ron-Lavi:
@nofaralfasi @adamruzicka tests are failing again on:
`ActiveRecord::AssociationTypeMismatch: RemoteExecutionFeature(#72560) expected, got nil which is an instance of NilClass(#40)`

in #554 updating the versions and using `>=` seemed to solve it, any other thoughts on why it happens?

adamruzicka:
I'm failing to reproduce that locally, not sure where it comes from. Only thing I can say is that apparently at the time when the test is run, the feature is not seeded (yet?)

Ron-Lavi (commented):
@nofaralfasi can you try adding the `db:seed` step in the test setup? https://github.com/theforeman/foreman_ansible/blob/master/.github/workflows/ruby_tests.yml#L73

nofaralfasi:
> @nofaralfasi can you try adding the `db:seed` step in the test setup? https://github.com/theforeman/foreman_ansible/blob/master/.github/workflows/ruby_tests.yml#L73

Do we need the `db:test:prepare`? 
From what I checked, `db:prepare` suppose to do all three - db:create, db:migrate, and db:seed. 
So should I add `db:seed` instead the `db:test:prepare`?

Ron-Lavi:
> Do we need the `db:test:prepare`? 
> From what I checked, `db:prepare` suppose to do all three - db:create, db:migrate, and db:seed. 
> So should I add `db:seed` instead the `db:test:prepare`?   

Let's try and see if it fixes the tests then :)

foreman_ansible.gemspec
```diff
@@ -20,6 +20,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'acts_as_list', '~> 1.0.3'
   s.add_dependency 'deface', '< 2.0'
-  s.add_dependency 'foreman_remote_execution', '>= 8.0.0'
-  s.add_dependency 'foreman-tasks', '>= 7.0.0'
+  s.add_dependency 'foreman_remote_execution', '~> 8.0.0'
```
ekohl, line 23: This pins to 8.0.x. Did you intend 8.x?
```suggestion
  s.add_dependency 'foreman_remote_execution', '~> 8.0'
```

Same below for foreman-tasks.

ekohl (commented):
> From what I checked, db:prepare suppose to do all three - db:create, db:migrate, and db:seed.

In the past I've seen problems where db:create or db:migrate didn't properly flush caches and it had stale table information. That's why I always run them as separate commands.

foreman_ansible.gemspec
nofaralfasi: Yes, thank you @ekohl!

stejskalleos:
The issue with `ActiveRecord::AssociationTypeMismatch: RemoteExecutionFeature` is also reproducible locally on `master` branch, you have to destroy db, recreate it, migrate it and then you will see the error too. 
But only once, for second run of `bundle exec rake test:foreman_ansible -v --trace` you won't see it again.

Ron-Lavi:
@nofaralfasi can you rebase please?

Ron-Lavi (approved):
It's finally green thanks to #571 !
thanks @nofaralfasi :)

PR #562, opened 2022-09-23
Fixes #35557 - host details variables page cant load json

MariaAga:
* to reproduce:
create an ansible variable with type json and default value {"test":1} and override true
open host details page, ansible, variables tab
result:
error page with this error:
```
uncaught Error: Objects are not valid as a React child (found: object with keys {test}). If you meant to render a collection of children, use an array instead.
in EditableValue (created by AnsibleVariableOverridesTable)
```

* after fixing that issue, when editing the value there was a console error saying:
```
TextArea: TextArea requires either an id or aria-label to be specified 
```
fixed that too

* and edit values were not parsed correctly when editing and that is also fixed -
json values turned into `[Object object]` array values were spread so `[1,2,3]` was `1,2,3`


MariaAga:
test failures dont look related

Ron-Lavi (approved):
Thanks @MariaAga !

PR #564, opened 2022-09-29
Fixes #35579 - Pass parameters as keyword arguments

ekohl:
This raises a deprecation warning in Ruby 2.7 and breaks in Ruby 3.


ekohl:
Rebased on latest master, but still relevant.

PR #566, opened 2022-10-13
Fixes #35628 - Keep logs from failed Github actions

stejskalleos:


adamruzicka:
Having logs could help here. This is what we do in rex https://github.com/theforeman/foreman_remote_execution/pull/610

stejskalleos:
> Having logs could help here. This is what we do in rex https://github.com/theforeman/foreman_remote_execution/pull/610

Nice, that'll be useful!

adamruzicka:
Would be nice if the tests actually failed now so we could see whether it really works

adamruzicka (approved):
I couldn't see it in action, but we have the exact same thing in REX and there it works well. We can always go back and fix it in case there are any issues with it

adamruzicka:
Thank you @stejskalleos !

PR #567, opened 2022-10-19
Fixes #35646 - Remove deprecated must_equal in tests

stejskalleos:


adamruzicka:
Do I see it correctly that the whole test suite actually runs three times?

stejskalleos:
> Do I see it correctly that the whole test suite actually runs three times?

Yes you are right:
```
Execute test:foreman_remote_execution # I don't know where is this coming from
Execute test:foreman_ansible # Correct
Execute test # That's the access_permissions_test.rb
```

Ron-Lavi (approved):
Thanks @stejskalleos!
as per @adamruzicka comment, let's investigate more about it in #561 whether we need to run the REX tests before this plugin.

PR #568, opened 2022-10-20
Refs #35143 - Use a productized version macro

adamruzicka:


stejskalleos (changes requested):
I just realized that the template name is `Capsule Upgrade Playbook`, which really doesn't make sense for the upstream, we should rename it to `Smart Proxy Upgrade Playbook`

Otherwise it works well, tested together with https://github.com/theforeman/foreman/pull/9474 and proxy was successfully upgraded.

adamruzicka:
Done

stejskalleos:
Tests are failing due to issue fixed in https://github.com/theforeman/foreman_ansible/commit/5ce2d3cde21ded2baba0e0841cbd56ed8563f2be, can you rebase please?

PR #571, opened 2022-10-25
Fixes #35669 - Fix Ansible Provider Test

nofaralfasi:
Adding a name field to the RemoteExecutionFeature, to save it correctly in the DB. This is required for the `when it is using the ansible_run_host feature` test to succeed.

### DIFF
--- a/test/unit/ansible_provider_test.rb
+++ b/test/unit/ansible_provider_test.rb
@@ -19,11 +19,8 @@ class AnsibleProviderTest < ActiveSupport::TestCase
     end
 
     context 'when it is using the ansible_run_host feature' do
-      let(:rex_feature) do
-        RemoteExecutionFeature.where(:label => 'ansible_run_host').first_or_create
-      end
-
       it 'has remote_execution_command false' do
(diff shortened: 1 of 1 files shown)


test/unit/ansible_provider_test.rb
```diff
@@ -20,7 +20,7 @@ class AnsibleProviderTest < ActiveSupport::TestCase
```
stejskalleos, line 20: Can you try this:

```ruby
    context 'when it is using the ansible_run_host feature' do
      it 'has remote_execution_command false' do
        rex_feature = RemoteExecutionFeature.where(label: 'ansible_run_host', name: 'Run Ansible roles').first_or_create
        template_invocation.template.remote_execution_features << rex_feature
        assert_not command_options[:remote_execution_command]
      end
    end
```

I am able to reproduce the issue locally when I run `RemoteExecutionFeature.destroy_all` before the test,
fix for me was adding `name` feature and remove `let`, it's not useful anyway in this case.

stejskalleos (approved):
And it's :green_apple: :green_apple: :green_apple: 

Let's get this in, thanks @nofaralfasi !

PR #572, opened 2022-11-07
Fixes #35585 - add an option to enable the callback plugin for job templates

nofaralfasi:
Added the property ansible_callback_enabled to the templates table. Created a Pagelet on the template's form to add the callback option to the UI.

Requires:
- https://github.com/theforeman/foreman/pull/9503

A few questions regarding these changes:
1. Do we want to allow to change the callback-plugin value if the template is locked?
2. Should we add a check for the `Job Category` type (if it is for Ansible), to decide whether to display the new ansible-callback tab? 
3. Is this change requires updating the plugin's version? 

A screenshot with the new tab in the UI:
image

### DIFF
--- a/app/controllers/foreman_ansible/api/v2/job_templates_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/api/v2/job_templates_controller_extensions.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ForemanAnsible
+  module Api
+    module V2
+      # Extends the job_templates api controller to support creating/updating with ansible callback plugin
+      module JobTemplatesControllerExtensions
+        extend Apipie::DSL::Concern
+
+        update_api(:create, :update) do
+          param :job_template, Hash do
+            param :ansible_callback_enabled, :bool, :desc => N_('Enable the callback plugin for this template')
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -123,7 +123,7 @@ if defined? ForemanRemoteExecution
 
         def ansible_command?(template)
           template.remote_execution_features.
-            where(:label => 'ansible_run_host').empty?
+            where(:label => 'ansible_run_host').empty? && !template.ansible_callback_enabled
         end
       end
     end
--- a/app/views/job_templates/_job_template_callback_tab_content.html.erb
+++ b/app/views/job_templates/_job_template_callback_tab_content.html.erb
@@ -1,0 +1,3 @@
+<div class="tab-pane" id="ansible_callback_enabled">
+  <%= checkbox_f f, :ansible_callback_enabled, :label => _('Enable Ansible Callback'), :disabled => @template.locked? %>
+</div>
--- a/app/views/job_templates/_job_template_callback_tab_headers.html.erb
+++ b/app/views/job_templates/_job_template_callback_tab_headers.html.erb
@@ -1,0 +1,13 @@
+<li><a id="ansible_tab_header" href="#ansible_callback_enabled" data-toggle="tab"><%= _("Ansible") %></a></li>
+<script type="text/javascript">
+    $(document).ready(function () {
+        var provider_type = $('#job_template_provider_type');
+        provider_type.change(setAnsibleTabVisibilityByProvider);
+        provider_type.change();
+
+        function setAnsibleTabVisibilityByProvider() {
+            var tab_header = $("#ansible_tab_header");
+            this.value === 'Ansible' ? tab_header.show() : tab_header.hide();
+        }
+    });
+</script>
--- a/db/migrate/20221003153000_add_ansible_callback_enabled_to_templates.rb
+++ b/db/migrate/20221003153000_add_ansible_callback_enabled_to_templates.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddAnsibleCallbackEnabledToTemplates < ActiveRecord::Migration[6.0]
+  def change
+    add_column :templates, :ansible_callback_enabled, :boolean, default: false
+    RemoteExecutionFeature.where(label: 'ansible_run_host').each do |rex_feature|
+      Template.find(rex_feature.job_template_id).update(ansible_callback_enabled: true)
+    end
+  end
+end
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -237,6 +237,8 @@ Foreman::Plugin.register :foreman_ansible do
     register_report_origin 'Ansible', 'ConfigReport'
   end
 
+  extend_rabl_template 'api/v2/job_templates/show', 'api/v2/job_templates/job_templates'
+
   describe_host do
     multiple_actions_provider :ansible_hosts_multiple_actions
   end
@@ -256,5 +258,17 @@ Foreman::Plugin.register :foreman_ansible do
                         :name => _('Update Smart Proxy'),
                         :partial => 'foreman/smart_proxies/update_smart_proxy',
                         :onlyif => ->(proxy, view) { view.can_update_proxy?(proxy) }
+  end
+  extend_page('templates/_form') do |context|
+    context.add_pagelet :tab_headers,
+                        :name => _('Ansible'),
+                        :partial => 'job_templates/job_template_callback_tab_headers',
+                        :onlyif => ->(subject, _view) { subject.is_a? JobTemplate }
+  end
+  extend_page('templates/_form') do |context|
+    context.add_pagelet :tab_content,
+                        :name => _('Ansible'),
+                        :partial => 'job_templates/job_template_callback_tab_content',
+                        :onlyif => ->(subject, _view) { subject.is_a? JobTemplate }
   end
 end
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Foreman::Plugin.register :foreman_ansible do
-  requires_foreman '>= 3.5'
+  requires_foreman '>= 3.6'
 
   settings do
     category :ansible, N_('Ansible') do
--- a/app/controllers/concerns/foreman/controller/parameters/job_template_extensions.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/job_template_extensions.rb
@@ -1,0 +1,17 @@
+module Foreman
+  module Controller
+    module Parameters
+      module JobTemplateExtensions
+        extend ActiveSupport::Concern
+
+        class_methods do
+          def job_template_params_filter
+            super.tap do |filter|
+              filter.permit :ansible_callback_enabled
+            end
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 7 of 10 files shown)


db/migrate/20221003153000_add_ansible_callback_enabled_to_templates.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddAnsibleCallbackEnabledToTemplates < ActiveRecord::Migration[6.0]
+  def change
+    add_column :templates, :ansible_callback_enabled, :boolean, default: false
+    Template.where(name: 'Ansible Roles - Ansible Default').update_all(ansible_callback_enabled: true)
```
adamruzicka, line 6: In here we should also find the templates which are assigned to `ansible_run_host` feature and set the flag for them as well.

We could even go one step further and set the flag only for the templates with the feature to not introduce any surprising behavior.

app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -122,8 +122,7 @@ def proxy_batch_size
         private
 
         def ansible_command?(template)
-          template.remote_execution_features.
-            where(:label => 'ansible_run_host').empty?
+          !template.ansible_callback_enabled
```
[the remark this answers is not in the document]
nofaralfasi: But in the UI we still give the option to check/uncheck the `enable ansible callback` option. 
Wouldn't that be confusing? 

If you say it's OK then I will keep both of them, just making sure.

db/migrate/20221003153000_add_ansible_callback_enabled_to_templates.rb
nofaralfasi, line 6: @adamruzicka did you mean doing this:
` Template.find(RemoteExecutionFeature.where(label: 'ansible_run_host').first.job_template_id).update_all(ansible_callback_enabled: true)`?

adamruzicka: Almost. The way the database schema is laid out, it is possible that one feature will have more than one template assigned to it.

app/models/foreman_ansible/ansible_provider.rb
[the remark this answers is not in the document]
adamruzicka: >  Wouldn't that be confusing?

That's true, but still probably worth it over breaking existing workflows.

db/migrate/20221003153000_add_ansible_callback_enabled_to_templates.rb
adamruzicka: > You mean that we should give the option to enable the callback only for templates assigned to the ansible_run_host feature?

Yes, but only in this migration

> And also implement it in the UI?

No, users should be free to do whatever they please. It is just that the migration should honor whatever changes the users might have already made

nofaralfasi:
Hi @adamruzicka, I updated some of the things we talked about, but not all.
I have some changes related to the question I asked here https://github.com/theforeman/foreman/pull/9503#discussion_r1022588995, and also I couldn't find a way yet to add the new `ansible_callback_enabled` field to the `job_templates` API.

adamruzicka (changes requested):

lib/foreman_ansible/register.rb
```diff
@@ -257,4 +259,16 @@ def ansible_variables_with_overrides
                         :partial => 'foreman/smart_proxies/update_smart_proxy',
                         :onlyif => ->(proxy, view) { view.can_update_proxy?(proxy) }
   end
+  extend_page('templates/_form') do |context|
+    context.add_pagelet :tab_headers,
+                        :name => _('Ansible Callback'),
+                        :partial => 'job_templates/job_template_callback_tab_headers',
+                        :onlyif => ->(subject, view) { view.correct_template? }
```
nofaralfasi, line 266: @adamruzicka is there a way to implement this method call in one line? (without having to create the `JobTemplatesHelper`?)

nofaralfasi: For some reason I am getting that `subject` is nil

adamruzicka: Ah, interesting. It works for `:tab_content`. Most likely because you are not providing the subject for the headers in here https://github.com/theforeman/foreman/pull/9503/files#diff-652d7608bcc9387e191b8c1a53a8cf45bb75cbf21e776863d7b14ca206d9774fR15

nofaralfasi: I found out why. I didn't pass it in foreman core, when calling the `render_pagelets_for(:tab_headers)`.

adamruzicka:
Personally I'd just ignore it. Rubocop thinks it is good enough.

If you wanted to do it though, this would be one of the possible approaches:
1) Identify a large chunk of `register.rb` - good candidates would be the `settings` block, `security_block` or the `extend_page` calls
2) Move it to a separate file (and module as a `self.something` method)
3) Require the file in `register.rb`
4) Call the module from step 2 inside the `Foreman::Plugin.register` block and pass self to it

To me going through all this isn't really worth it. The changes would be mostly unrelated to the actual changes you're trying to do here and as far as I know codeclimate isn't really used all that much in the wider foreman ecosystem.

nofaralfasi:
> To me going through all this isn't really worth it. The changes would be mostly unrelated to the actual changes you're trying to do here and as far as I know codeclimate isn't really used all that much in the wider foreman ecosystem.

Then I will leave it that way. Thanks for the explanation!

app/views/job_templates/_job_template_callback_tab_headers.html.erb
```diff
@@ -0,0 +1 @@
+<li><a href="#ansible_callback_enabled" data-toggle="tab"><%= _("Ansible Callback") %></a></li>
```
adamruzicka, line 1: The name predestines the entire tab to never having anything else in it apart from the single checkbox. How about we go with more generic "Ansible"?

app/views/job_templates/_job_template_callback_tab_content.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="tab-pane" id="ansible_callback_enabled">
+  <%= checkbox_f f, :ansible_callback_enabled, :label=>_('Enable Ansible Callback'), :disabled => @template.locked? %>
```
adamruzicka, line 2: Spaces around the rocket after `:label`

adamruzicka, line 2: It is somewhat unfortunate that right now the tab is displayed even if the provider type is not ansible. I was able to put together a makeshift solution with jquery, although I don't really dare to suggest you to go down that path.

@MariaAga If we have an old erb page that we're extending from a plugin using pagelets and we want the content to dynamically change based on field from the original, non-extended page, do we have any other option apart from registering an onchange handler onto the fields we're interested in?

MariaAga: pagelets are really not my area, so we might have to use that suggested jquery

app/views/job_templates/_job_template_callback_tab_headers.html.erb
```diff
@@ -0,0 +1,13 @@
+<li><a id="ansible_tab_header" href="#ansible_callback_enabled" data-toggle="tab"><%= _("Ansible") %></a></li>
+<script type="text/javascript">
+    $(document).ready(function () {
+        var provider_type = $('#job_template_provider_type');
+        provider_type.change(ansibleTabVisibilityByProvider);
+        provider_type.change();
+
+        function ansibleTabVisibilityByProvider() {
+            var tab_header = $("#ansible_tab_header");
+            $("#select2-chosen-1").text() === 'Ansible' ?  tab_header.show() : tab_header.hide();
```
nofaralfasi, line 10: @adamruzicka this is the jquery code I added. 
Do you think we should add this line of code to the false statement (rewrite the if condition):
` $("#job_template_ansible_callback_enabled").prop('checked', false);`

This is for cases the user sets the `ansible_callback_enabled` to true and then changes the provider to `Script`. I'm not sure if it's necessary.

adamruzicka: That's probably not necessary. They flag might be set, but if the provider_type is not ansible, it will not be used in any way.

app/views/job_templates/_job_template_callback_tab_headers.html.erb
```diff
@@ -0,0 +1,13 @@
+<li><a id="ansible_tab_header" href="#ansible_callback_enabled" data-toggle="tab"><%= _("Ansible") %></a></li>
+<script type="text/javascript">
+    $(document).ready(function () {
+        var provider_type = $('#job_template_provider_type');
+        provider_type.change(ansibleTabVisibilityByProvider);
+        provider_type.change();
```
adamruzicka, line 6: what does this do?

app/views/job_templates/_job_template_callback_tab_headers.html.erb
adamruzicka, line 10: This function can take an argument - the element that changed. From it we can pull the value without resorting to touching select2's internals.

app/views/job_templates/_job_template_callback_tab_headers.html.erb
nofaralfasi: The first call (`provider_type.change(ansibleTabVisibilityByProvider)`) attaches the function to execute when a change event occurs.
The second call (`provider_type.change()`) triggers a change event (to call the `ansibleTabVisibilityByProvider` function when the page loads).

adamruzicka:
I don't think the change in foreman will be picked to 3.5 so maybe bumping the required foreman version would be in order?

Also I'm not sure what the test failures are about

nofaralfasi:
> Also I'm not sure what the test failures are about

I updated foreman dependency, and now the tests are passing.

app/views/job_templates/_job_template_callback_tab_content.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="tab-pane" id="ansible_callback_enabled">
+  <%= checkbox_f f, :ansible_callback_enabled, :label => _('Enable Ansible Callback'), :disabled => @template.locked? %>
```
nofaralfasi, line 2: @adamruzicka do we need to add translations for that? (I've never done this so I'm not sure what it includes).

app/controllers/foreman_ansible/api/v2/job_templates_controller_extensions.rb
```diff
@@ -0,0 +1,18 @@
...
+module ForemanAnsible
+  module Api
+    module V2
+      # Extends the job_templates api controller to support creating/updating with ansible callback plugin
+      module JobTemplatesControllerExtensions
+        extend Apipie::DSL::Concern
+
+        update_api(:create, :update) do
+          param :job_template, Hash do
+            param :ansible_callback_enabled, :bool, :desc => N_('Enable the callback plugin for this template')
```
nofaralfasi, line 12: @adamruzicka same question here about the translations

app/views/job_templates/_job_template_callback_tab_content.html.erb
adamruzicka: Add translations as in "mark the strings for translation"? Yes

Usually marked strings are extracted by the maintainer outside of the changes which introduced the new strings so if you meant extracting strings or even translating them then that would be no

adamruzicka (approved):
I don't really have anything substantial to add here.

The migration could be made more efficient, but it doesn't really matter since in almost all cases we'll be updating a single record.

Codeclimate is not happy about this change, but it seems the consensus is to get rid of codeclimate completely so I'll go ahead and ignore it.

tl;dr: let's get this in

adamruzicka:
Thank you @nofaralfasi !

PR #519, opened 2022-03-14
Use reusable GitHub Actions for testing

ekohl:
This reduces the duplication between projects. For example, this ensures testing on Ruby 2.7 which previously didn't happen.

There are various preparation PRs, see the commits for details.

Right now this is incomplete because it loses the caching step that happens on a push to master. It also still uses the repository in my own namespace. This is just to see if it works.

PR #574, opened 2022-11-08
Don't run ruby tests on push to master

stejskalleos:


nofaralfasi:
Hi @Ron-Lavi, do you know if we need to run tests when things are merged to the master branch?

Ron-Lavi (approved):

ekohl:
> do you know if we need to run tests when things are merged to the master branch?

This creates caches which can be reused. It greatly speeds up PR testing between multiple users.

The background is that PR testing creates a cache that can **only** be used by the user who submitted the PR. If you also run it on test it creates a shared cache that's used for everyone.

I'd suggest to revert this change.

nofaralfasi:
>  If you also run it on test it creates a shared cache that's used for everyone.

What do you mean by that? Maybe you meant to say that if you also run the tests on **master** it creates a shared cache...
And thanks for the explanation @ekohl!

ekohl:
Yes, that is exactly what I intended.

For what it's worth, back in March I tried to unify all actions in reusable ones so most plugin authors don't need to think about it anymore. https://github.com/theforeman/actions is the repo and https://github.com/theforeman/actions/issues/1 has a list of TODO items.

https://github.com/theforeman/foreman_ansible/pull/519 needs a rebase after this PR, but that is how I think we should manage testing using GitHub Actions.

PR #575, opened 2022-11-09
Updating bump version command for DEB branches

nofaralfasi:
I think we were missing the `ruby-` prefix before `foreman-ansible` in the bump version command for the `deb` branches in the releasing docs.
I updated the command to the command that I was using which worked for me.


nofaralfasi:
[test test_ruby]

PR #577, opened 2022-11-29
Fixes #35799 - fix ansible report returns 500 error code

nofaralfasi:
Handle the case when job execution failed, and display the error message in the UI.

Requires:
- https://github.com/theforeman/foreman/pull/9561

[show full error message on config_reports page](https://user-images.githubusercontent.com/38184193/207911268-acfc516d-f58e-437b-afe6-12e7d0bd0b60.webm)

### DIFF
--- a/app/helpers/foreman_ansible/ansible_reports_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_reports_helper.rb
@@ -27,6 +27,8 @@ module ForemanAnsible
 
     def ansible_module_message(log)
       msg_json = parsed_message_json(log)
+      return _("Execution error: #{msg_json['msg']}") if msg_json['failed'].present?
+
       module_action = msg_json['module']
       case module_action
       when 'package'
@@ -65,6 +67,17 @@ module ForemanAnsible
       module_name(log).present?
     rescue StandardError
       false
+    end
+
+    def show_full_error_message_value(message_value)
+      tag.div class: 'replace-hidden-value' do
+        link_to_function(icon_text('plus', '', class: 'small'), 'replace_value_control(this, "div")',
+                         title: _('Show full value'),
+                         class: 'replace-hidden-value pull-right') +
+          (tag.span class: 'full-value' do
+            message_value
+          end)
+      end
     end
 
     private
--- a/app/views/foreman_ansible/config_reports/_ansible.html.erb
+++ b/app/views/foreman_ansible/config_reports/_ansible.html.erb
@@ -29,8 +29,12 @@
                 <% end %>
               </ul>
             <% else %>
-              <%= log_message %>
-          <% end %>
+              <% allowed_length = 150 %>
+              <div class='pull-left'>
+                <%= truncate(log_message, length: allowed_length) %>
+              </div>
+              <%= show_full_error_message_value(log_message) if log_message.length > allowed_length %>
+            <% end %>
           </td>
         </tr>
       <% end %>
--- a/test/unit/helpers/ansible_reports_helper_test.rb
+++ b/test/unit/helpers/ansible_reports_helper_test.rb
@@ -19,4 +19,18 @@ ANSIBLELOG
       ansible_module_message(log).to_s
     )
   end
+
+  test 'module message extraction with error' do
+    log_value = <<-ANSIBLELOG.strip_heredoc
+    {"msg": "AnsibleUndefinedVariable", "changed": false, "_ansible_no_log": false, "failed": true, "module": "template", "exception": "raise AnsibleUndefinedVariable"}
+ANSIBLELOG
+    message = FactoryBot.build(:message, value: log_value)
+    log = FactoryBot.build(:log, message: message)
+    log.message = message
+
+    assert_match(
+      'Execution error: AnsibleUndefinedVariable',
+      ansible_module_message(log).to_s
+    )
+  end
 end
--- a/foreman_ansible.gemspec
+++ b/foreman_ansible.gemspec
@@ -20,8 +20,6 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'acts_as_list', '~> 1.0.3'
   s.add_dependency 'deface', '< 2.0'
-  # foreman_remote_execution 9.0 requires Foreman 3.6+, while
-  # foreman_ansible still works with 3.5+, so we allow 8.x and 9.x
-  s.add_dependency 'foreman_remote_execution', '>= 8.0', '< 10'
+  s.add_dependency 'foreman_remote_execution', '~> 9.0'
   s.add_dependency 'foreman-tasks', '~> 7.0'
 end
--- a/lib/foreman_ansible/version.rb
+++ b/lib/foreman_ansible/version.rb
@@ -4,5 +4,5 @@
 # This way other parts of Foreman can just call ForemanAnsible::VERSION
 # and detect what version the plugin is running.
 module ForemanAnsible
-  VERSION = '10.2.0'
+  VERSION = '11.0.0'
 end


app/helpers/foreman_ansible/ansible_reports_helper.rb
```diff
@@ -32,6 +32,7 @@ def ansible_module_message(log)
       when 'package'
         msg_json['results'].empty? ? msg_json['msg'] : msg_json['results']
       when 'template'
+        return _("#{msg_json['msg']}") if msg_json['failed']
```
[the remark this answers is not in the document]
nofaralfasi: I suggest keeping the error handling only for the `tempalte` case for now.
The reason is that we don't have all the types of messages of all other errors, and we can't always display them properly.
So if we apply this code for all cases, we might get one of these:
image

nofaralfasi:
Requires:
- https://github.com/theforeman/foreman/pull/9561

nofaralfasi:
Requires:
- https://github.com/theforeman/foreman/pull/9561

[show full error message on config_reports page](https://user-images.githubusercontent.com/38184193/207911268-acfc516d-f58e-437b-afe6-12e7d0bd0b60.webm)

nofaralfasi:
There are still some failing tests, that need to be fixed (not related to the changes in my PR).

evgeni:
Interesting, I think https://github.com/theforeman/foreman_ansible/pull/580 handles a *similar* but maybe not exactly the same issue?

nofaralfasi:
> Interesting, I think #580 handles a _similar_ but maybe not exactly the same issue?

Yes, it handles a very similar issue. 
The changes from your PR solve the 500 error I was trying to fix, but they don't display the error message. 
So if you are getting an error for one of the cases with the `invocation` field, you would get an empty message in the `config_report` UI:
image

app/views/foreman_ansible/config_reports/_ansible.html.erb
```diff
@@ -29,8 +29,12 @@
                 <% end %>
               </ul>
             <% else %>
-              <%= log_message %>
-          <% end %>
+              <% allowed_length = 150 %>
+              <div class='pull-left'>
+                <%= truncate(log_message, :length => allowed_length) %>
```
stejskalleos, line 34: Nit: `length: allowed_length`

test/unit/helpers/ansible_reports_helper_test.rb
```diff
@@ -19,4 +19,18 @@ class AnsibleReportsHelperTest < ActiveSupport::TestCase
       ansible_module_message(log).to_s
     )
   end
+
+  test 'module message extraction with error' do
+    log_value = <<-ANSIBLELOG.strip_heredoc
+    {"msg": "AnsibleUndefinedVariable", "changed": false, "_ansible_no_log": false, "failed": true, "module": "template", "exception": "raise AnsibleUndefinedVariable"}
+ANSIBLELOG
+    message = FactoryBot.build(:message)
```
stejskalleos, line 27:
```ruby
message = FactoryBot.build(:message, value: log_value)
```

test/unit/helpers/ansible_reports_helper_test.rb
```diff
@@ -19,4 +19,18 @@ class AnsibleReportsHelperTest < ActiveSupport::TestCase
...
+    message.value = log_value
+    log = FactoryBot.build(:log)
```
stejskalleos, line 29:
```ruby
log = FactoryBot.build(:log, message: message)
```

test/unit/helpers/ansible_reports_helper_test.rb
```diff
@@ -19,4 +19,18 @@ class AnsibleReportsHelperTest < ActiveSupport::TestCase
...
+    log.message = message
+    assert_match(
```
stejskalleos, line 31: Put empty line between variable definitions and the assert, it's more readable

stejskalleos (changes requested):
I see that title is `fix ansible report returns 500 error code`, but I don't see anywhere the change for it, only for notifications. Do I miss something?

And could you please share how to reproduce the error as in the video? All I get is an error without any further details, message field is always `No additional data`

app/views/foreman_ansible/config_reports/_ansible.html.erb
```diff
@@ -29,8 +29,12 @@
...
+              </div>
+              <%= show_full_error_message_value(log_message) if log_message.length > allowed_length %>
```
Ron-Lavi, line 36: I think what you were looking for is the [trunc_with_tooltip](https://github.com/theforeman/foreman/blob/3347fa49d500964f0209122d8d36c920d1feafcc/app/helpers/layout_helper.rb#L175) helper.

I am not saying it's the best UX, but it's been used all across foreman and plugins. In addition, if we want to cherry-pick this to older versions, I would avoid adding the dependency PR in foreman 3.6, which will make the cherry-pick more difficult.

Ron-Lavi (commented):
Thanks @nofaralfasi !

app/views/foreman_ansible/config_reports/_ansible.html.erb
nofaralfasi: The problem with `trunc_with_tooltip` helper, is that it doesn't display long texts properly. 
And as you can see in the attached [video](https://github.com/theforeman/foreman_ansible/pull/577#issue-1468313812), sometimes we are getting long error messages.  
So I can use that, but not all messages will display properly.

nofaralfasi:
> I see that title is `fix ansible report returns 500 error code`, but I don't see anywhere the change for it, only for notifications. Do I miss something?
> 

The notifications was the cause for the 500 error code. We are trying to access `msg_json['invocation']['module_args']`, from the `ansible_module_message` [method](https://github.com/theforeman/foreman_ansible/blob/f7af100cec68eefde8f4b7bc9305a712ba113c7f/app/helpers/foreman_ansible/ansible_reports_helper.rb#L35), however when we get an error - these keys don't exist. 

> And could you please share how to reproduce the error as in the video? All I get is an error without any further details, message field is always `No additional data`

For me, the easiest way to reproduce this error is to:
1. Set the `Default verbosity level` (Administer -> Settings -> Ansible tab) to `Level 4 (-vvvv)`. 
2. Run ansible roles on a non-running host (shutoff or that you can't access it).
3. The report for this execution should display the error.

app/views/foreman_ansible/config_reports/_ansible.html.erb
Ron-Lavi: yeah, your solution looks way better than the tooltip,
tested with Foreman and it works great!

Ron-Lavi:
Thanks @nofaralfasi, tested with the Foreman PR and it works great :)
waiting for the test changes that @stejskalleos requested before merging.

test/unit/helpers/ansible_reports_helper_test.rb
nofaralfasi: This won't work, because of this [line](https://github.com/theforeman/foreman/blob/f3881ee0fca2b8b3d78467a5013c7e4b0347828c/test/factories/reports_related.rb#L41) in Foreman core.
We can change it to: `log.message = FactoryBot.create(:message) unless log.message`, for it to work. The question is if we want to do it.

app/helpers/foreman_ansible/ansible_reports_helper.rb
```diff
@@ -67,6 +69,17 @@ def ansible_report?(log)
       false
     end
 
+    def show_full_error_message_value(message_value)
+      tag.div class: 'replace-hidden-value' do
+        link_to_function(icon_text('plus', '', class: 'small'), 'replace_value_control(this, "div")',
```
Ron-Lavi, line 74: since this change requires foreman 3.6,
can you update the requires_foreman line in this plugin?
and can you also update the REX dependency in https://github.com/theforeman/foreman_ansible/blob/master/foreman_ansible.gemspec#L23?

Ron-Lavi (commented):
tests are failing on: `NoMethodError: undefined method `name' for "Audited::Audit":String`
but I doesn't seem to be related to your PR

evgeni (commented):
This breaks compatibility with 3.5 and older, while I think the change to `ansible_module_message` is a valid bugfix on it's own?

So how about splitting this into two: one PR to fix the 500 error, which can be backported, and one for the UI improvement that remains 3.6+?

Ron-Lavi:
@evgeni Merry Christmas ⛄ aren't you on vacation? :)

@nofaralfasi had an idea to make the UI change compatible with Foreman 3.5 with a small hack... I suggested cherry pick this commit later to the Foreman 3.5 compatible version and change just the UI code a bit, what do you think?

evgeni:
> @evgeni Merry Christmas snowman aren't you on vacation? :)

Sure, sure ;-)

Ron-Lavi:
@nofaralfasi can you also update the requires_foreman version and REX dependency?

Ron-Lavi:
> Should I also bump foreman_ansible version to 11.0.0? Or do we want to do it in a different PR? 

Yes you can, it should be the final commit (e.g if there are tests fixes they should be in a commit before the bump commit)

nofaralfasi:
@Ron-Lavi 
I'm updating the last commit in this PR due to the changes from this [PR](https://github.com/theforeman/foreman_ansible/pull/572) that got merged. 
The foreman version was already bumped to 3.6 so I'm only adding the ansible and REX versions update.

Ron-Lavi (commented):
Thanks @nofaralfasi !
can you also update REX dependency and remove the comment there? https://github.com/theforeman/foreman_ansible/blob/master/foreman_ansible.gemspec#L23-L26

nofaralfasi:
> Thanks @nofaralfasi ! can you also update REX dependency and remove the comment there? https://github.com/theforeman/foreman_ansible/blob/master/foreman_ansible.gemspec#L23-L26

I updated Rex dependency on the second commit in this PR https://github.com/theforeman/foreman_ansible/pull/577/commits/7fab14d913eb9c104f7fc078683d53f084a910ba

Ron-Lavi (approved):
Great work @nofaralfasi ! LGTM

PR #578, opened 2022-12-02
Pass parameters as keyword arguments in test

ekohl:
I noticed this in https://github.com/theforeman/foreman_ansible/pull/576 while looking at the failed test. This only solves the warning, not the actual test failure.


nofaralfasi:
This PR contains identical changes to those found in https://github.com/theforeman/foreman_ansible/pull/564. Closing it as a duplicate.

PR #580, opened 2022-12-20
Fixes #35883 - don't fail if data is missing due to no_log

evgeni:
When you run an Ansible module with `no_log: true`, the report does not contain any `invocation` data and thus also no `module_args`.

### DIFF
--- a/app/helpers/foreman_ansible/ansible_reports_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_reports_helper.rb
@@ -30,11 +30,11 @@ module ForemanAnsible
       return _("Execution error: #{msg_json['msg']}") if msg_json['failed'].present?
 
       module_action = msg_json['module']
+      module_args = msg_json.fetch('invocation', {}).fetch('module_args', {})
       case module_action
       when 'package'
         msg_json['results'].empty? ? msg_json['msg'] : msg_json['results']
       when 'template'
-        module_args = msg_json['invocation']['module_args']
         _("Rendered template #{module_args['_original_basename']} to #{msg_json['dest']}")
       when 'service'
         _("Service #{msg_json['name']} #{msg_json['state']} (enabled: #{msg_json['enabled']})")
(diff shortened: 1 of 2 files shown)


fidalcastro:
I am no ruby expert, is there any way that I can help to get this MR merged ? or What is it required to get this MR merged ?

nofaralfasi:
> I am no ruby expert, is there any way that I can help to get this MR merged ? or What is it required to get this MR merged ?

Hi @fidalcastro, so basically since all the tests pass, this PR is now waiting for a review before it can be merged by one of the contributors.
You can find more information here: https://github.com/theforeman/foreman/blob/develop/CONTRIBUTING.md

nofaralfasi:
I reproduced this issue by following the steps from the Redmine [issue](https://projects.theforeman.org/issues/35883) and got this result (I'm referring to the last row: the `lol` task with the `Copy to` message):
image

Are we good with displaying the message like that?
Other than that, LGTM.

evgeni:
> Are we good with displaying the message like that?

I thought about this, and it's certainly better than a crash :see_no_evil: 
We could go further and try to display "unknown" or something for values that aren't present, or hide the message altogether if we detect "no_log: True"? In that case I'd probably do a follow-up cosmetic change to do that, as I think the more defensive way of accessing the data as done here in the PR is beneficial either way.

nofaralfasi:
> I thought about this, and it's certainly better than a crash see_no_evil We could go further and try to display "unknown" or something for values that aren't present, or hide the message altogether if we detect "no_log: True"? In that case I'd probably do a follow-up cosmetic change to do that, as I think the more defensive way of accessing the data as done here in the PR is beneficial either way.

Sounds good :)

fidalcastro:
Thanks @evgeni @nofaralfasi for working on this issue.

May I know what foreman &foreman-ansible version includes this fix ?

 I checked in https://github.com/theforeman/foreman_ansible/releases but I couldn't find the `35884` in the changelogs & I also see the `v11.0.0` is meant for foreman release 3.6 but I don't see any foreman release `v3.6`  in https://github.com/theforeman/foreman.

I tried running foreman `v3.5.1` with foreman ansible `v10.1.0` (BTW, I don't see any 10.1.0 version in foreman_ansible release page), still I see the same error.

nofaralfasi:
Hey @fidalcastro, you are right, this fix wasn't released yet. It will be included in the next foreman_ansible release (for foreman 3.6). You can find more details about the release of foreman 3.6 [here](https://community.theforeman.org/t/foreman-3-6-schedule-and-planning/31871). 

I added a new [release](https://github.com/theforeman/foreman_ansible/releases/tag/v10.1.0) for `v10.1.0`. Hope that helps!

PR #585, opened 2023-01-09
Refs #35883 - log useful message if ansible output was censored

evgeni:


app/helpers/foreman_ansible/ansible_reports_helper.rb
```diff
@@ -28,6 +28,7 @@ def check_mode_log?(log)
     def ansible_module_message(log)
       msg_json = parsed_message_json(log)
       return _("Execution error: #{msg_json['msg']}") if msg_json['failed'].present?
+      return msg_json['censored'] if msg_json['censored'].present?
```
evgeni, line 31: I wonder if we should return the Ansible message, or write something ourself?

evgeni:
somehow GHA didn't trigger :(

app/helpers/foreman_ansible/ansible_reports_helper.rb
nofaralfasi: I think the Ansible message is good. Do you want to ask someone for a third opinion?

evgeni: nah, if you're good with that, so am I :)

nofaralfasi:
Thanks @evgeni! (third time's a charm)

PR #587, opened 2023-01-26
Fixes #35994 - check if user has administrator permissions by its usergroup

nofaralfasi:


nofaralfasi:
I'm wondering if would it be enough to fix the tests, or do we need to create a new query for usergroups as discussed with @Ron-Lavi

Ron-Lavi:
> I'm wondering if would it be enough to fix the tests, or do we need to create a new query for usergroups as discussed with @Ron-Lavi

I think in this case you are correct, reading the redmine again, looks like we need to fetch the usergroup of a user :)

Ron-Lavi (approved):
Thanks @nofaralfasi ! tested and I see that the user with usergroup of admin roles can access the Ansible sub tabs without getting permissions error

PR #541, opened 2022-07-10
Fixes #35190 - added ansible actions to hostgroups actions array

nofaralfasi:
Added ansible action to the actions' dropdown on the Hostgroups page without overriding the existing ones.

PR #592, opened 2023-01-31
Fixes #36021 - add an option to configure Ansible job for hostgroup

nofaralfasi:
The option to configure Ansible job for hostgroup existed in the past and was removed here https://github.com/theforeman/foreman_ansible/pull/541.
We are adding this option back as it was originally.


Ron-Lavi (approved):
Works well @nofaralfasi! thank you for fixing it

PR #597, opened 2023-02-16
Fixes #36015 - API endpoints to add/remove single roles

ofedoren:


app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -64,6 +64,30 @@ def assign_ansible_roles
...
+
+          def add_ansible_roles
+            find_resource
+            process_response @hostgroup.ansible_roles << @ansible_roles
+          end
+
+          api :DELETE, '/hostgroups/:id/remove_ansible_roles',
+              N_('Remove directly assigned Ansible roles from a hostgroup')
+          param :id, :identifier, :required => true
+          param :ansible_role_ids, Array,
```
nofaralfasi, line 83: How do you pass an array of ansible_role_ids? I wasn't able to do that (but it works with just one ansible_role_id)

ofedoren: You can simply use JSON object with array: { "ansible_role_ids": [1,2,3] }

test/functional/api/v2/hosts_controller_test.rb
```diff
@@ -87,6 +88,34 @@ class HostsControllerTest < ActionController::TestCase
...
+             :session => set_session_user
+        assert_response :success
+        assert_equal assigns('host').ansible_roles, [@ansible_role3, @ansible_role2, @ansible_role1]
+      end
+
+      test 'should remove only specified roles from a host' do
+        host = FactoryBot.create(:host,
+                                 :managed => false,
+                                 :ansible_role_ids => [@ansible_role3.id, @ansible_role2.id, @ansible_role1.id])
+        post :remove_ansible_roles,
```
nofaralfasi, line 110: The only thing I couldn't figure out yet is how the tests are passing even though it is a post request (same for the host). Should it be a delete request?

ekohl (changes requested):
I'm not commenting inline, since `/add_X` and `remove_x` are used in both the hostgroup and the hosts controllers. It's not a very REST thing include the action in the endpoint: the HTTP verb describes what you want.

In REST you have collections and resources. A collection is `/hosts` and a resource is `/hosts/:id`. Then the HTTP verb describes the action.

More practical, that means you have a collection `/hosts/:id/ansible/roles`. You can `GET` them to get a listing, `POST` to it to create a new link, which you also direct you to the individual resources.

The individual resources would likely be on /hosts/:host_id/ansible/roles/:role_id`, which you can `GET` to view and `DELETE` to remove. If you want to update it, `PUT` is typically used.

https://restfulapi.net/resource-naming/ does mention that sometimes you do have a more procedural concept, but in this case it maps very well to collections and resources.

Now I will note you here add and delete collections, but https://restfulapi.net/http-methods/#delete really focuses on individual resources. That's what REST is all about. The PR summary describes you say add/remove single roles, so per resource is probably sufficient for that. If that's really what you want to support bulk actions, I'm not sure REST is correct. If you want it in the UI, the graphql API may be more appropriate than REST.

test/functional/api/v2/hosts_controller_test.rb
ofedoren: That's a good question... :/

nofaralfasi: I think@adamruzicka describes it pretty well [here](https://matrix.to/#/!dxfWqeLNsImROheGXo:matrix.org/$16770832718951BGFWd:libera.chat?via=libera.chat&via=matrix.org&via=ctrl-c.liu.se)

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -64,6 +65,28 @@ def assign_ansible_roles
...
+
+          api :POST, '/hostgroups/:id/ansible_roles',
+              N_('Directly add an Ansible role to a hostgroup')
+          param :id, :identifier, :required => true
+          param :ansible_role_id, :identifier,
+                N_('Ansible role to add to a hostgroup'),
+                :required => true
+
+          def add_ansible_role
+            process_response @hostgroup.ansible_roles << @ansible_role
```
ekohl, line 77: What if the role is already in there? Is `ansible_roles` a set or a list?

test/functional/api/v2/hostgroups_controller_test.rb
```diff
@@ -63,6 +63,32 @@ class HostgroupsControllerTest < ActionController::TestCase
...
+      test 'should append a role to a hostgroup with a correct ordering' do
+        hostgroup = FactoryBot.create(:hostgroup,
+                                      :ansible_role_ids => [@ansible_role1.id])
+        post :ansible_roles,
+             :params => {
+               :id => hostgroup.id,
+               :ansible_role_id => [@ansible_role2.id]
+             },
+             :session => set_session_user
+        assert_response :success
```
ekohl, line 76: I'd expect HTTP 201 Created if it was added, not HTTP 200 Success

test/functional/api/v2/hostgroups_controller_test.rb
```diff
@@ -63,6 +63,32 @@ class HostgroupsControllerTest < ActionController::TestCase
...
+      test 'should remove only specified roles from a hostgroup' do
+        hostgroup = FactoryBot.create(:hostgroup,
+                                      :ansible_role_ids => [@ansible_role2.id, @ansible_role1.id])
+        delete :ansible_roles,
+               :params => {
+                 :id => hostgroup.id,
+                 :ansible_role_id => [@ansible_role2.id]
+               },
+               :session => set_session_user
+        assert_response :success
```
ekohl, line 89: HTTP 204 (No Content) is common after a DELETE. If it's 200 Success, I wonder what's in the content.

test/functional/api/v2/hostgroups_controller_test.rb
```diff
@@ -63,6 +63,32 @@ class HostgroupsControllerTest < ActionController::TestCase
...
+             :params => {
+               :id => hostgroup.id,
+               :ansible_role_id => [@ansible_role2.id]
+             },
+             :session => set_session_user
+        assert_response :success
+        assert_equal assigns('hostgroup').ansible_roles, [@ansible_role1, @ansible_role2]
+      end
+
+      test 'should remove only specified roles from a hostgroup' do
```
ekohl, line 80: Another use case: what if the role doesn't exist? Does it return 404 or 200/204 since it's not part of the role collection anymore? I'd expect the latter.

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
ofedoren: If the role is already assigned, it will return
```JSON
{
    "error": {
        "message": "Validation failed: Ansible role has already been taken"
    }
}
```
with `500` error code.

test/functional/api/v2/hostgroups_controller_test.rb
ofedoren: The server responds with correct `201`.

P.S. Apparently `post` and others are expecting `:action` instead of route from `routes` definition. `ofedoren.experience += 1`. Also, `assert_response` here considers both `200` and `201` as successful via mapping, that's why the test doesn't fail, but I've changed to use the specific number instead.
P.P.S. I hate test frameworks :/

test/functional/api/v2/hostgroups_controller_test.rb
ofedoren: The server will respond with `200`.

test/functional/api/v2/hostgroups_controller_test.rb
ofedoren: We have the same behaviour with responses almost everywhere else since we use the same logic/methods. In this case the server will respond with `200` and the modified entity (hostgroup in this case).

ofedoren:
[test foreman_ansible]

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
ofedoren: I agree that `500` code is weird for such response, but Rails will treat this as a validation error anyway. I left `422` instead of suppressing the error. If you still find this wrong, I can change it to suppress the `ActiveRecord::RecordInvalid` and return `200`.

nofaralfasi (approved):
LGTM. Everything works as expected. 
@ekohl can I merge? :)

PR #600, opened 2023-02-22
Fixes #36136 - make sure validatorRule is a string

Ron-Lavi:
variable.validatorRule.split was raising a TypeError: string.split is not a function


nofaralfasi:
Thanks! @Ron-Lavi

PR #602, opened 2023-02-23
Fixes #36136 - make sure validatorRule is a string

nofaralfasi:
variable.validatorRule.split was raising a TypeError: string.split is not a function

(cherry picked from commit bba4f31e6955b0833973cda768461a9521aca6a6)


Ron-Lavi (approved):
test failures are not related,
thanks @nofaralfasi !

PR #609, opened 2023-03-07
Fixes #36171 - Config report with empty log message

stejskalleos:


ekohl (approved):
Not a foreman_ansible maintainer, but looks safe and correct to me. The only thing I'd consider is displaying some placeholder text if it's really empty, but I'd be happy to prevent a crash first.

PR #618, opened 2023-05-04
add packit config

evgeni:


evgeni:
```
/var/tmp/rpm-tmp.cNaeQr: line 33: scl: command not found
```

right, this ain't gonna work with those SCL'ed templates, lemme fix that.

evgeni:
there you go, green \o/

PR #619, opened 2023-05-04
Fixes #36358 - Translation improvements

stejskalleos:
* Extract missing strings for the translations
* Update config for the transifex-cli
* Update *.po files


webpack/components/AnsibleRolesAndVariables/index.js
```diff
@@ -11,11 +12,11 @@ import { selectApiImportStatus } from './AnsibleRolesAndVariablesSelectors';
...
-    { title: 'Name' },
-    { title: 'Operation' },
-    { title: 'Variables' },
-    { title: 'Hosts Count' },
-    { title: 'Hostgroups count' },
+    { title: __('Name') },
+    { title: __('Operation') },
+    { title: __('Variables') },
+    { title: __('Hosts Count') },
+    { title: __('Hostgroups count') },
```
ekohl, line 19: Should it be `Host count` and `Hostgroup count`?

ekohl (changes requested):

locale/action_names.rb
```diff
@@ -1,5 +1,6 @@
 # Autogenerated!
 _("Action with sub plans")
-_("Import facts")
 _("Import Puppet classes")
+_("Import facts")
+_("Preupgrade job")
```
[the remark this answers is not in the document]
stejskalleos: Huh, interesting, no idea how it get there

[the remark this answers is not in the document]
ekohl: If you have it in your Foreman bundler.d then it's loaded.

.gitignore
```diff
@@ -10,6 +10,7 @@ build/
 dist/
 *.egg-info/
 extras/*.pyc
+locale/*.mo
```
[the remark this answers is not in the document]
stejskalleos: What's the correct approach here - commit `mo` files to the repo or not?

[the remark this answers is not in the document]
ekohl: It's an excellent question I've been looking into, but haven't found a good solution. I have a feeling we should use some rake task to compile them and invoke that in packaging.

I think https://github.com/theforeman/foreman/pull/9692 could provide a good base but given we have about a week of development time left in 3.7 I'm not sure we can get it done in time.

foreman_ansible.gemspec
```diff
@@ -3,6 +3,12 @@
 require File.expand_path('../lib/foreman_ansible/version', __FILE__)
 require 'date'
 
+Dir['locale/**/*.po'].each do |po|
+  mo = po.sub(/foreman_ansible\.po$/, 'LC_MESSAGES/foreman_ansible.mo')
+  warn "WARNING: File #{mo} does not exist, generate with 'make all-mo'!" unless File.exist?(mo)
+  warn "WARNING: File #{mo} outdated, regenerate with 'make all-mo'" if File.mtime(po) > File.mtime(mo)
+end
```
ekohl, line 10: This now produces a warning, but that doesn't solve it. I think you somehow need to guarantee that the .mo files end up in packaging.

I've been thinking about this, but I don't know about the release process. You can use `gem build` and because it lists `locale` it will include files in that directory. But because it's a warning there is no guarantee they are present or in sync.

If you always release via rake tasks you can maybe hook into that.

The benefit of including the .mo files in the released .gem is that also in non-packaged environments (bundle install, like a container) you get them.

Any idea what would be preferred?

stejskalleos: > Any idea what would be preferred?

I've been thinking about it, would be nice to unify it across all plugins, for example in the foreman_leapp we [ignore](https://github.com/theforeman/foreman_leapp/blob/master/.gitignore#L8) the `.mo` files.

I agree that it makes sense to commit the `mo` files, and also have this warning in the `gemspec` file.

ekohl: I'd still like to avoid adding the .mo files to git, but at this moment I'm not sure if there is a better way. One way could be to make the warning a fatal error. It may introduce more problems though.

stejskalleos: I would rather avoid raising an error, that could cause more problems as you say.

So the approach is:
* Have a warning about outdated/missing `.mo` files 
* Don't commit `.mo` files

Just want to confirm so I do the same in other plugins

.tx/config
```diff
@@ -1,9 +1,10 @@
...
-
-[foreman.foreman_ansible]
-file_filter = locale/<lang>/foreman_ansible.edit.po
-source_file = locale/foreman_ansible.pot
-source_lang = en
-type = PO
+host = https://app.transifex.com
 
+[o:foreman:p:foreman:r:foreman_ansible]
+file_filter   = locale/<lang>/foreman_ansible.po
```
adamruzicka, line 5: As @ofedoren pointed out somewhere else and from what I've seen, this should stay `locale/<lang>foreman_ansible.edit.po`

adamruzicka:
Could you please also do the po_to_json dance here while you're at it?

adamruzicka:
Shouldn't this be kept around? iirc this pr marked some additional strings for extraction

stejskalleos:
> Shouldn't this be kept around? iirc this pr marked some additional strings for extraction

I will post it again with the proper changes and updated strings as described in the foreman plugin docs

PR #624, opened 2023-05-13
allow foreman-tasks 8.x

evgeni:


evgeni:
@nofaralfasi could you please make a release once this and #623 are in? should unblock plugin/luna nightlies.

nofaralfasi:
Thanks @evgeni!

PR #632, opened 2023-05-25
Fixes #36358 - Translation improvements

stejskalleos:


locale/action_names.rb
```diff
@@ -1,5 +1,5 @@
 # Autogenerated!
-_("Action with sub plans")
-_("Import Puppet classes")
-_("Import facts")
 _("Remote action:")
+_("»Action with sub plans«")
+_("»Import Puppet classes«")
+_("»Import facts«")
```
adamruzicka, line 5: Latest tasks should be adding a newline at the end

stejskalleos: I have latest `develop`

adamruzicka: tasks don't have develop?

ekohl, line 5: Why do these get `»` and `«`? I think it's the "mark for translation" setting that adds this, but is it leading to double translations?

ekohl (changes requested):

locale/action_names.rb
stejskalleos: Interesting, it wasn't there before I updated `foreman-tasks` plugin. I will fix it

stejskalleos:
@ekohl @adamruzicka Updated, now it should be ok. I hope.

adamruzicka (commented):
The changes look sane, but I'm not sure what's up with the tests

stejskalleos:
Okay, so changes in the translation somehow cause Ruby tests to fail. I don't touch `*.rb` files at all, so I'm kinda baffled about what and why is happening ... In other PRs everything is fine

nofaralfasi:
> Okay, so changes in the translation somehow cause Ruby tests to fail. I don't touch `*.rb` files at all, so I'm kinda baffled about what and why is happening ... In other PRs everything is fine

Rerunning the tests fixed that :)

stejskalleos:
> Rerunning the tests fixed that :)

Huh, I did rerun the tests, didn't help. I observed the same behavior on my local setup (latest `master`), where when I was running tests on a fresh database dynflow failed with an error. The second run was okay though.

stejskalleos:
@adamruzicka It's :green_apple:

adamruzicka:
Thank you @stejskalleos !

PR #633, opened 2023-05-29
Fixes #36273 - Use proper permission for editing Ansible variable

pmoravec:
Requires:
* https://github.com/theforeman/foreman/pull/9803

### DIFF
--- a/lib/foreman_ansible/register.rb
+++ b/lib/foreman_ansible/register.rb
@@ -167,7 +167,8 @@ Foreman::Plugin.register :foreman_ansible do
         :create_job_invocations, :view_job_templates,      # to allow the play_roles
         :create_template_invocations, :view_smart_proxies, # ...
         :view_ansible_roles, :destroy_ansible_roles,
-        :import_ansible_roles, :view_ansible_variables,
+        :import_ansible_roles, :view_ansible_variables, :view_lookup_values,
+        :create_lookup_values, :edit_lookup_values, :destroy_lookup_values,
         :create_ansible_variables, :import_ansible_variables,
         :edit_ansible_variables, :destroy_ansible_variables, :import_ansible_playbooks]
 
--- a/app/controllers/api/v2/ansible_override_values_controller.rb
+++ b/app/controllers/api/v2/ansible_override_values_controller.rb
@@ -24,7 +24,7 @@ module Api
       param_group :ansible_override_value, :as => :create
 
       def create
-        @ansible_variable = AnsibleVariable.authorized(:edit_external_variables).
+        @ansible_variable = AnsibleVariable.authorized(:edit_ansible_variables).
                             find_by(:id => params[:ansible_variable_id].to_i)
         @override_value = @ansible_variable.lookup_values.create!(lookup_value_params['override_value'])
         @ansible_variable.update_attribute(:override, true)
--- a/app/models/ansible_variable.rb
+++ b/app/models/ansible_variable.rb
@@ -24,7 +24,7 @@ class AnsibleVariable < LookupKey
   end
 
   def editable_by_user?
-    AnsibleVariable.authorized(:edit_external_parameters).
+    AnsibleVariable.authorized(:edit_ansible_variables).
       where(:id => id).exists?
   end
 end
--- a/app/views/ansible_variables/index.html.erb
+++ b/app/views/ansible_variables/index.html.erb
@@ -22,7 +22,7 @@
             icon_text((variable.override ? "flag": ""), variable.key.to_s, :kind => 'fa', :title => _('Overriden')),
             hash_for_edit_ansible_variable_path(:id => variable).
             merge(:auth_object => variable,
-                  :permission => 'edit_external_parameters',
+                  :permission => 'edit_ansible_variables',
                   :authorizer => authorizer)
         ) %></td>
         <td class="ellipsis"><%= link_to_if_authorized(


nofaralfasi:
Hey @pmoravec, I agree with the change you made in this PR. However, the fix only applies to the API. We also need to address the problem in the web UI. 
Furthermore, I noticed that there are no tests for this functionality. Could you please add them?

adamruzicka:
> We also need to address the problem in the web UI.

That's true, there are two other places where arguably wrong permission (`edit_external_parameters`) is used.

>  Could you please add them?

I'd be willing to let this slide. We can reason whether the permission used is the right one or not, but the whole authorization framework is thoroughly tested in Foreman itself.

pmoravec:
I see. Sadly I dont have bandwidth to complete the fix as requested - please let somebody else adds a fix for WebUI "variant" and add some tests.

lib/foreman_ansible/register.rb
```diff
@@ -137,6 +137,12 @@
                  :'api/v2/ansible_variables' => [:create]
                },
                :resource_type => 'AnsibleVariable'
+    permission :view_lookup_values,
```
stejskalleos, line 140: Permission `:view_ansible_roles` already got `{ :lookup_values => [:index] }`

lib/foreman_ansible/register.rb
```diff
@@ -137,6 +137,12 @@
...
+               { :lookup_values => [:index] },
+               :resource_type => 'LookupValue'
+    permission :create_lookup_values,
```
stejskalleos, line 143: Permission `:create_ansible_variables` already have `:lookup_values => [:create]`

stejskalleos (changes requested):

lib/foreman_ansible/register.rb
nofaralfasi: Since we don't even have permissions for the `lookup_values` resource, non-admin users can't access it (see https://github.com/theforeman/foreman/blob/develop/app/services/authorizer.rb#L34 and https://github.com/theforeman/foreman/blob/develop/app/graphql/mutations/base_mutation.rb#L31). 

You can refer to the output log below, which demonstrates the absence of filters related to the permission for the LookupValue class:
`no filters found for given permission`
`checking permission  for class LookupValue`
`organization_ids: []`
`location_ids: []`

Let me know if you need more details.

ofedoren:
Agree on that it should be using `:*_ansible_variables` here, since as noted @stejskalleos it somehow defined to use `lookup_values` permissions as well: https://github.com/theforeman/foreman_ansible/blob/5b44890713842fd3c5f2646590ecb18265f52ebf/lib/foreman_ansible/register.rb#L119-L124.

The only thing that bothers me, I didn't find any defined permissions on LookupValue model in Foreman itself :/

nofaralfasi:
> The only thing that bothers me, I didn't find any defined permissions on LookupValue model in Foreman itself :/

I see the LookupValue is defined in foreman core. Would it solve the issue if I add the permissions for the LookupValue resource there?

ofedoren:
> Would it solve the issue if I add the permissions for the LookupValue resource there?

It should :shrug:

nofaralfasi:
How to test:
1. Create a non-admin user, with `'Ansible Roles Manager'` and `View hosts` permissions.
2. UI - Login as the new user, and try to create, edit and delete Ansible variables from the host details page.
3. API - Test the following requests using the new user credentials: `POST /ansible/api/ansible_override_values`, `DELETE /ansible/api/ansible_override_values/:id`.

Note:
* Even if the user doesn't have the `Edit Host` permission, he will still be able to modify Ansible variables from the host details page. This functionality existed prior to this PR's changes, and if we want to change that behaviour, it would require disabling the corresponding option within the `Specify Matchers` section on the ansible_variable edit page.

nofaralfasi:
The tests are not passing due to the dependency on the merge of https://github.com/theforeman/foreman/pull/9803.

ofedoren (approved):
Thanks, @nofaralfasi and @pmoravec, can be merged once https://github.com/theforeman/foreman/pull/9803 is in.

lib/foreman_ansible/register.rb
```diff
@@ -160,14 +160,19 @@
                { :'api/v2/ansible_inventories' => [:schedule] }
     permission :import_ansible_playbooks,
                { :'api/v2/ansible_playbooks' => [:sync, :fetch] }
+    permission :view_lookup_values, {}, :resource_type => 'LookupValue'
+    permission :create_lookup_values, {}, :resource_type => 'LookupValue'
+    permission :edit_lookup_values, {}, :resource_type => 'LookupValue'
+    permission :destroy_lookup_values, {}, :resource_type => 'LookupValue'
```
ofedoren, line 166: Why is this needed? I mean, I've removed those lines and it still works...

nofaralfasi: Did you first remove these permissions from the DB? 
From the rails console: `Permission.where("name LIKE '%_lookup_values%'").destroy_all` 
And then re-run the rails server to test it.

ofedoren: I've done this and it looks... not quite good.

I've destroyed permissions, but we still need to create them, to use here. Thus, I needed to run `db:seed` to re-create them (we have defined them in core), but due to Ansible plugin needing these perms to create ARM role, it failed. Thus, after I removed Ansible plugin from loading, I've re-created the permissions, then enabled the plugin and run the server. I was able to change variable values even without this block.

I mean, we already create these perms in the core, why would we try to re-create them here? Seems weird. I've checked what we do with e.g. `create_job_invocations`, since this permission is used for ARM role as well: it seems REX plugin creates it and Ansible plugin then simply uses it to define a role. As you can see, Ansible plugin doesn't call `permission :create_job_invocations`, which is right.

Since I'm not so well familiar with the internals of permission logic, @adamruzicka, maybe you can help us out?

nofaralfasi: I forgot I need to make a change in the `db/seeds.d/020-permissions_list.rb` file for the `db:seed` command to include that file. That's why it didn't work as expected for me.
Therefore, as long as we run the `db:seed` command, there's no requirement to additionally include these permissions in the `register.rb` file, right?

ofedoren: :shrug: that's what I hoped for...

ofedoren: But the test failures now show what I was facing as well :/ Not sure why it complains about perms that are being created in DB by Foreman itself, isn't here some issues with load ordering?

PR #637, opened 2023-06-01
do not pass initial `nil` value to Settings DSL

evgeni:
this has been deprecated for a while and will be removed in Foreman 3.8


ezr-ondrej (approved):
Thanks @evgeni, this is safe enough, so I hope I won't cause any trouble if I merge this.

So sorry if I do cause trouble and please let me know in that case @nofaralfasi, @adamruzicka, I promise to help remedy :)

PR #642, opened 2023-06-08
Fixes #35585 - add an option to enable the callback plugin for job templates

nofaralfasi:
Added the property ansible_callback_enabled to the templates table. Created a Pagelet on the job template's form to add the callback option to the UI.

(cherry picked from commit 34dd51c86938813a0ee64f510fde5f40c8f6015c)


nofaralfasi:
Note:
While these CP changes require upgrading to foreman-3.6 as in the original PR https://github.com/theforeman/foreman_ansible/pull/572, we decided not to update it here.

PR #645, opened 2023-06-17
Fix typo in convert2rhel job description

ekohl:
I also noticed that the line break tags will show up literally in the UI. I think they should be removed. The current UI doesn't appear to convert line breaks to newlines either.


nofaralfasi:
Thanks @ekohl!

PR #646, opened 2023-06-20
don't depend on theforeman/stories

evgeni:
this was dropped from core in https://github.com/theforeman/foreman/commit/5e7343ecdad7632a94efa26fcba84889e66b06b4


nofaralfasi:
Thanks @evgeni!

nofaralfasi:
@evgeni do we need to require foreman version >= 3.8 for this?

evgeni:
I don't think so, no.

IMHO (@MariaAga can correct me here) having a `@theforeman/stories` in plugins was wrong anyway, unless they published own stories, which almost nobody did anyway (`foreman_ansible` did not for sure)

PR #647, opened 2023-06-26
Fixes #36469 - Fix ansible roles and variables pagination

girijaasoni:
@MariaAga @nofaralfasi 
Fixed the pagination for the roles tab, for the variables tab however, we need total variables count. There is not get API to fetch the total overriden variables and their count. Do we create one?

### DIFF
--- a/webpack/components/AnsibleHostDetail/components/RolesTab/index.js
+++ b/webpack/components/AnsibleHostDetail/components/RolesTab/index.js
@@ -23,7 +23,7 @@ const RolesTab = ({ hostId, history, canEditHost }) => {
   const [assignModal, setAssignModal] = useState(false);
   const renameData = data => ({
     ansibleRoles: data.host.ownAnsibleRoles.nodes,
-    totalCount: data.host.ownAnsibleRoles.totalCount,
+    totalCount: allAnsibleRoles.length,
   });
 
   const useFetchFn = () =>


nofaralfasi:
> @MariaAga @nofaralfasi Fixed the pagination for the roles tab, for the variables tab however, we need total variables count. There is not get API to fetch the total overriden variables and their count. Do we create one?

As we discussed in DM, I suggest adding such API, as we do for the roles tab.

MariaAga:
If we add the api, we might want to drop the graphql? to avoid double calls

nofaralfasi:
> If we add the api, we might want to drop the graphql? to avoid double calls

You are right, I take it back.
By looking at the graphql [documentation](https://graphql.org/learn/pagination/) I think we should keep the graphql query and update its pagination parameters.

MariaAga:
Graphql doesnt support page pagination as we use it in Foreman AFAIK, and if we want to use it that way we need to add the aggregate functionality.

I talked to Girija offline, and I think for the ansible roles/vars we can:
1 fetch all items and do the pagination calculations in ui (showing only x items each time by slicing the array), only ok if we know customers arent using too many items (max 100~)
2 do it with a new api endpoint instead of using gql ( so remove the gql call, and rely only on api)
3 try to add aggregate functionality to the roles and variable types in ruby ( which might be the most complicated as we dont have anything like that in our code)

nofaralfasi:
> Just a mention: recently there was a fix related to showing total count for nodes in Foreman: [theforeman/foreman@cb88310](https://github.com/theforeman/foreman/commit/cb88310aafc92e704238fc70ded56ee3b173fad9). Maybe this will clearify things a bit?

Thank you Oleh, I wasn't familiar with that, so for me that helped. 
I see now it is more complicated than I thought. Perhaps @stejskalleos could offer his perspective on this matter?

nofaralfasi:
Closing for now. Feel free to reopen.

PR #648, opened 2023-07-05
Fixes #36568 - Hostgroup API controller also returns roles assigned to hosts in a host group

amasolov:
I believe the API call shouldn't return the list of roles assigned to hosts in a hostgroup directly. Just a list of assigned and inherited roles is enough.

### DIFF
--- a/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
@@ -50,7 +50,7 @@ module ForemanAnsible
             @inherited_ansible_roles = @hostgroup.inherited_ansible_roles
             @directly_assigned_roles = @hostgroup.ansible_roles
             @ansible_roles = (
-              @directly_assigned_roles + @inherited_ansible_roles + @hostgroup.host_ansible_roles
+              @directly_assigned_roles + @inherited_ansible_roles
             ).uniq
           end
 


amasolov:
Not sure why the tests are failing though

nofaralfasi:
Thanks @amasolov. The failing tests weren't related.

nofaralfasi (approved):

app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -50,7 +50,7 @@ def ansible_roles
             @inherited_ansible_roles = @hostgroup.inherited_ansible_roles
             @directly_assigned_roles = @hostgroup.ansible_roles
             @ansible_roles = (
-              @directly_assigned_roles + @inherited_ansible_roles + @hostgroup.host_ansible_roles
```
ofedoren, line 53: That was a slight refactor of https://github.com/theforeman/foreman_ansible/blob/5b44890713842fd3c5f2646590ecb18265f52ebf/app/models/concerns/foreman_ansible/hostgroup_extensions.rb#L37-L42 usage for response. I don't think we need to change that.

amasolov: @ofedoren I believe that should be implemented as a separate API call then. I do see a strong reason behind getting a list of roles applied to a hostgroup but including roles assigned directly to hosts to that list shouldn't be default, maybe optional.

nofaralfasi: I agree with Alexey here. 
IMHO we shouldn't return Ansible roles that are not directly assigned to the hostgroup by default.
For example, if I have a host which is a member of a hostgroup, and that host has ansible roles assigned to it, when I run `hammer hostgroup ansible-roles list --id hostgroup_id` (or via API) I will get these roles in the results, even though they are not directly assigned to that hostgroup (or to its sub-hostgroups).
I think that can be confusing.

aplace-lab (approved):

ofedoren (approved):
If so, sure, I won't block on this :)

nofaralfasi:
Thanks @amasolov!

PR #649, opened 2023-07-12
Fixes #36136 - handle the case where Ansible variable is an array

nofaralfasi:


nofaralfasi:
In cases where we have more than 1 matcher, the UI would look like this:
image

Details of the Ansible variable:
image

Would we like to change it to make it clearer?

MariaAga:
@nofaralfasi How did it look before the change? Should we also change how host parameters behave? (the code in host parameters is similar to code of ansible variables, we should probably do some merging between them when we have the time)

MariaAga:
@nofaralfasi Didnt notice that variables have an option to merge overrides, and the parameters dont. So it should be fine as only a fix in Ansible :)

MariSvirik:
We can either use slash as you proposed for editing mode OR we can even add an extra field for each merged value under the first one when in editing mode. That might be visually easier to distinguish if the users want to edit the thing from this page. 
I think I had in my designs a tooltip saying that merging overrides are allowed. We should definitely display the information somewhere e.g. a tooltip or in the source attribute column.

stejskalleos:
@nofaralfasi 
I'm not sure what are we fixing here. Displaying variable on host detail or editing matchers?

nofaralfasi:
> @nofaralfasi I'm not sure what are we fixing here. Displaying variable on host detail or editing matches?

This PR is for fixing the "TypeError: string.split is not a function" error when accessing ansible variables on a host.

I took the opportunity to ask @MariSvirik's opinion regarding the Ansible variables UI. In my opinion, there is room for improvement in terms of how we present the information, as it may not be sufficiently clear.
To continue this discussion, I will create a community post since it is unrelated to the current issue at hand.

stejskalleos:
> This PR is for fixing the "TypeError: string.split is not a function" error when accessing ansible variables on a host.

How do I reproduce the issue then? I have Foreman + Ansible + Theme plugin and works fine for me. Tried two ansible variables, one `string`, second `array`, and the page is loading fine. Any concrete steps on how to reproduce it?

> To continue this discussion, I will create a community post since it is unrelated to the current issue at hand.

+1, let's not mix it here.

nofaralfasi:
> How do I reproduce the issue then? I have Foreman + Ansible + Theme plugin and works fine for me. Tried two ansible variables, one `string`, second `array`, and the page is loading fine. Any concrete steps on how to reproduce it?

1. Create a new Ansible variable with an array type. Default value is ["test"]
2. Enable both 'Merge Overrides' and 'Merge Default'
3. Add the variable to a role
4. Attach the role to a host
5. Check the ansible variables on the host details page

stejskalleos (approved):
:green_apple: **LGTM**
it works and fix the issue

stejskalleos:
Thanks @nofaralfasi

PR #650, opened 2023-07-19
Fixes #36569 - List only supported Job Templates on upstream

nofaralfasi:


lib/foreman_ansible/engine.rb
```diff
@@ -66,6 +66,7 @@ class Engine < ::Rails::Engine
       ::Api::V2::HostgroupsController.include ForemanAnsible::Api::V2::HostgroupsControllerExtensions
       ::Api::V2::HostgroupsController.include ForemanAnsible::Api::V2::HostgroupsParamGroupExtensions
       ::ConfigReportImporter.include ForemanAnsible::AnsibleReportImporter
+      ::Api::V2::JobTemplatesController.prepend ForemanAnsible::Api::V2::JobTemplatesControllerExtensions
       ::Api::V2::JobTemplatesController.include ForemanAnsible::Api::V2::JobTemplatesControllerExtensions
```
stejskalleos, line 70: The line with `include` should be removed.

app/controllers/concerns/foreman/controller/parameters/job_template_extensions.rb
```diff
@@ -4,6 +4,17 @@ module Parameters
       module JobTemplateExtensions
         extend ActiveSupport::Concern
 
+        UNSUPPORTED_JOB_TEMPLATES = [
```
stejskalleos, line 7: This file is only for extending parameters, I don't think this should be here.

app/controllers/foreman_ansible/api/v2/job_templates_controller_extensions.rb
```diff
@@ -6,12 +6,24 @@ module V2
...
           param :job_template, Hash do
             param :ansible_callback_enabled, :bool, :desc => N_('Enable the callback plugin for this template')
           end
         end
+
+        def index
+          return super if Foreman::Plugin.find('foreman_theme_satellite').present?
+
+          @job_templates = super
+          @job_templates = @job_templates.where.not(name: JobTemplatesControllerExtensions::UNSUPPORTED_JOB_TEMPLATES)
```
stejskalleos, line 25: What if instead of this we update the [db seed](https://github.com/theforeman/foreman_ansible/blob/364ed0fc1b272ca6be7cf639bcf26f65a3f054ec/db/seeds.d/75_job_templates.rb) for the Ansible job templates?
```ruby
    JobTemplate.without_auditing do
      Dir[File.join("#{ForemanAnsible::Engine.root}/app/views/foreman_ansible/"\
                    'job_templates/**/*.erb')].each do |template|
        sync = !Rails.env.test? && Setting[:remote_execution_sync_templates]
        template = JobTemplate.import_raw!(File.read(template),
                                           :default => true,
                                           :lock => true,
                                           :update => sync)
        template.organizations = organizations if template.present?
        template.locations = locations if template.present?
      end
```

lib/foreman_ansible/engine.rb
stejskalleos: And it won't be needed if we update the db seed file instead. See the comment ^^

stejskalleos (changes requested):

lib/foreman_ansible/engine.rb
nofaralfasi: > The line with include should be removed.

If I remove it then the `update_api` method in this module won't be included.

app/controllers/foreman_ansible/api/v2/job_templates_controller_extensions.rb
nofaralfasi: If my understanding is correct, this code block is responsible for importing new templates and storing them in the database. However, as the `Capsule Upgrade Playbook` already exists in our database, we must create a migration to remove it, in addition to making changes in the code. The same process applies to handling the `Smart Proxy Upgrade Playbook` in the reverse scenario for the downstream.

On one hand, this approach ensures that unsupported job templates are not saved in the database, thereby avoiding the need for further code changes in the API. However, on the other hand, it requires the necessary migrations to delete data.

@ShimShtein, would you like to share your thoughts on this matter?

stejskalleos: > However, as the Capsule Upgrade Playbook already exists in our database, we must create a migration to remove it, in addition to making changes in the code.

Do we need migration? Was it released to production already? In downstream it wasn't, not sure about the upstream.

nofaralfasi: AFAIK, it was released both downstream and upstream. 
[10.4.2](https://github.com/theforeman/foreman_ansible/pull/630) (downstream, foreman 3.5), [11.1.1](https://github.com/theforeman/foreman_ansible/pull/590) (foreman 3.6, foreman 3.7).

ShimShtein: Removing the template in seeds/migrations seems more clean, but we will need a code that explicitly removes the unwanted template there.
Something like:

``` ruby
unsupported_templates = ['capsule_upgrade.erb']
unsupported_templates = ['smart_proxy_upgrade.erb'] if plugin.exists? 'foreman_theme_satellite'
Dir["templates/*.erb"].except(unsuppored_templates).each do |template|
  create_template(template)
end

JobTemplate.where(name: unsupported_template).delete_all
```

ShimShtein:
Thank you! This  looks much cleaner indeed.

db/seeds.d/75_job_templates.rb
```diff
@@ -23,6 +30,13 @@
         template.organizations = organizations if template.present?
         template.locations = locations if template.present?
       end
+
+      unsupported_existed_templates = JobTemplate.where(name: unsupported_templates.keys)
+
+      unless unsupported_existed_templates.empty?
```
stejskalleos, line 36: IMHO `if unsupported_existed_templates.any?` is more readable

db/seeds.d/75_job_templates.rb
```diff
@@ -23,6 +30,13 @@
         template.organizations = organizations if template.present?
         template.locations = locations if template.present?
       end
+
+      unsupported_existed_templates = JobTemplate.where(name: unsupported_templates.keys)
```
stejskalleos, line 34: how about `unsupported_templates_in_db`?

stejskalleos (changes requested):

stejskalleos (approved):
:green_apple: **LGTM**

Works well for both scenarios (upstream/downstream) and the code looks good too.
@ShimShtein if you don't have any other comments I'm :+1:  to merge

PR #651, opened 2023-07-25
i18n - pulling from tx

stejskalleos:


locale/action_names.rb
```diff
@@ -1,5 +1,7 @@
 # Autogenerated!
 _("Action with sub plans")
+_("Check for long running tasks")
+_("Deliver notifications about long running tasks")
```
ekohl, line 4: Same question here as in leapp: where do these tasks come from? It's best to update the translations with minimal plugins present.

stejskalleos: When I did the changes only plugin enabled was foreman_ansible, nothing else

ekohl: https://github.com/theforeman/foreman-tasks/commit/03ce25bcd6f64bee07ffef665db422a89ae09c5c added them to foreman-tasks so we're going to see this in every repo that uses foreman-tasks. No way around it.

ekohl (changes requested):
It looks like Japanese went wrong in Transifex. I'm going to check my script where I synced it. Now I worry it may be wrong in multiple places.

ekohl:
I updated the translations with a correct encoding (https://github.com/ekohl/transync/commit/e0fea33a5fcd7b16a988a225d2982f1952c6a3dc). Please pull again.

stejskalleos:
@ekohl updated, hope it's fine now.

ekohl (approved):
Yep, thanks!

PR #654, opened 2023-08-07
Fixes #36136 - handle the case where Ansible variable is an array

nofaralfasi:
(cherry picked from commit cea1ecaa40e0ffc88fff28bb24a6bc5568713f02)


nofaralfasi:
@MariaAga @stejskalleos 
The test_js (12) is failing because we are using [Optional chaining (?.)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), which was added in Node.js 14 and is not directly available in Node 12.

Is there a way to fix that, or should we ignore this?

PR #668, opened 2023-11-27
Mark compatible with foreman-tasks 9.x

ekohl:


adamruzicka (approved):
+1, tasks 9.0.0 shouldn't contain any changes that would be breaking for f-ansible, it just requires a newer version of Foreman

PR #670, opened 2023-11-29
Refs #36953 - Fetch configured ansible tags on job rerun

girijaasoni:
Currently in progress, 
fixed:  fetching the tags,
needs update: setting the default tags value.

This PR is blocked by
`Fixes: [91318d8f24695d6f5d7c848e70ca7bcaaec6e71b](https://github.com/theforeman/foreman_remote_execution/pull/852)

### DIFF
--- a/app/models/foreman_ansible/ansible_provider.rb
+++ b/app/models/foreman_ansible/ansible_provider.rb
@@ -61,19 +61,23 @@ if defined? ForemanRemoteExecution
           true
         end
 
-        def provider_inputs
+        def provider_inputs(provider_input_values = [])
+          tags = provider_input_values&.find { |input| input[:name] == 'tags' }&.fetch(:value) || ''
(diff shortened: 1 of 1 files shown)


app/models/foreman_ansible/ansible_provider.rb
```diff
@@ -61,19 +61,23 @@ def supports_effective_user?
           true
         end
 
-        def provider_inputs
+        def provider_inputs(provider_input_values = {})
+          tags = provider_input_values&.find { |input| input[:name] == 'tags' }&.fetch(:value) || ''
```
adamruzicka, line 65: I'm just reading this without much context so I might be missing something, but at a glance, If `provider_input_values` is a hash as the default value suggests, then `input` inside `find` is an array of two elements (they key and the value) so accessing `item[:name]` on it will most likely fail.

girijaasoni: Thanks Adam!

PR #672, opened 2023-11-30
use shared test workflow from theforeman/actions

evgeni:


ekohl:
There's overlap with https://github.com/theforeman/foreman_ansible/pull/672.

Some differences: mine continues running `test/unit/foreman/access_permissions_test.rb`. Yours renames the file as well. I'm not sure I like `main` and would probably prefer `ci` or `test`. Once we add flows like `release` I think that makes more sense.

evgeni:
yes and yes

PR #676, opened 2023-12-08
Fixes #36971 - GUI to allow cloning of Ansible roles from VCS

Thorben-D:
This feature adds a new modal to /ansible/ansible_roles, which allows the user to clone a git-repo with roles to a specified SmartProxy.

[RedMine-Ticket](https://projects.theforeman.org/issues/36971)

GUI:

Screenshot_20231208_134233

**Depends on** https://github.com/theforeman/smart_proxy_ansible/pull/85

### DIFF
--- a/app/controllers/api/v2/vcs_clone_controller.rb
+++ b/app/controllers/api/v2/vcs_clone_controller.rb
@@ -1,0 +1,150 @@
+module Api
+  module V2
+    class VcsCloneController < ::Api::V2::BaseController
+      include ::ForemanAnsible::ProxyAPI
+      include ::Api::Version2
+
+      resource_description do
+        api_version 'v2'
+        api_base_url '/ansible/api'
+      end
+
+      def_param_group :repo_information do
+        param :repo_info, Hash, :desc => N_('Hash containing info about the Ansible role to be installed') do
+          param :vcs_url, String, :desc => N_('URL of the repository'), :required => true
+          param :role_name, String, :desc => N_('Name of the Ansible role'), :required => true
+          param :ref, String, :desc => N_('Branch / Tag / Commit reference'), :required => true
+        end
+      end
+
+      rescue_from ActionController::ParameterMissing do |e|
+        render json: { 'error' => e.message }, status: :bad_request
+      end
+
+      skip_before_action :verify_authenticity_token
+
+      before_action :set_proxy_api
+
+      api :GET, '/smart_proxies/:smart_proxy_id/ansible/vcs_clone/repository_metadata',
+          N_('Retrieve metadata about the repository associated with a Smart Proxy.')
+      param :smart_proxy_id, Array, N_('Name of the Smart Proxy'), :required => true
+      param :vcs_url, String, N_('URL of the repository'), :required => true
+      error 400, :desc => N_('Invalid or missing parameters')
+      def repository_metadata
+        vcs_url = params.require(:vcs_url)
+        render json: @proxy_api.repo_information(vcs_url)
+      end
+
+      api :GET, '/smart_proxies/:smart_proxy_id/ansible/vcs_clone/roles',
+          N_('Returns an array of Ansible roles installed on the provided Smart Proxy')
+      formats ['json']
+      param :smart_proxy_id, Array, N_('Name of the SmartProxy'), :required => true
+      error 400, :desc => N_('Invalid or missing parameters')
+      def installed_roles
+        render json: @proxy_api.list_installed
+      end
+
+      api :POST, '/smart_proxies/:smart_proxy_id/ansible/vcs_clone/roles',
+          N_('Launches a task to install the provided role')
+      formats ['json']
+      param_group :repo_information
+      param :smart_proxy_id, Array, N_('Smart Proxy where the role should be installed')
+      error 400, :desc => N_('Invalid or missing parameters')
+      def install_role
+        payload = verify_install_role_parameters(params)
+        start_vcs_task(payload, :install)
+      end
+
+      api :PUT, '/smart_proxies/:smart_proxy_id/ansible/vcs_clone/roles',
+          N_('Launches a task to update the provided role')
+      formats ['json']
+      param_group :repo_information
+      param :smart_proxy_id, Array, N_('Smart Proxy where the role should be installed')
+      error 400, :desc => N_('Invalid or missing parameters')
+      def update_role
+        payload = verify_update_role_parameters(params)
+        payload['name'] = params.require(:role_name)
+        start_vcs_task(payload, :update)
+      end
+
+      api :DELETE, '/smart_proxies/:smart_proxy_id/ansible/vcs_clone/roles/:role_name',
+          N_('Launches a task to delete the provided role')
+      formats ['json']
+      param :role_name, String, :desc => N_('Name of the role to be deleted')
+      param :smart_proxy_id, Array, N_('Smart Proxy to delete the role from')
+      error 400, :desc => N_('Invalid or missing parameters')
+      def delete_role
+        payload = params.require(:role_name)
+        start_vcs_task(payload, :delete)
+      end
+
+      private
+
+      def set_proxy_api
+        unless params[:id]
+          msg = _('Smart proxy id is required')
+          return render_error('custom_error', :status => :unprocessable_entity, :locals => { :message => msg })
+        end
+        ansible_proxy = SmartProxy.find_by(id: params[:id])
+        if ansible_proxy.nil?
+          msg = _('Smart proxy does not exist')
+          return render_error('custom_error', :status => :bad_request, :locals => { :message => msg })
+        else unless ansible_proxy.has_capability?('Ansible', 'vcs_clone')
+               msg = _('Smart Proxy is missing foreman_ansible installation or Git cloning capability')
+               return render_error('custom_error', :status => :bad_request, :locals => { :message => msg })
+             end
+        end
+        @proxy = ansible_proxy
+        @proxy_api = find_proxy_api(ansible_proxy)
+      end
+
+      def permit_parameters(params)
+        params.require(:vcs_clone).
+          permit(
+            repo_info: [
+              :vcs_url,
+              :role_name,
+              :ref
+            ]
+          ).to_h
+      end
+
+      def verify_install_role_parameters(params)
+        payload = permit_parameters params
+        %w[vcs_url role_name ref].each do |param|
+          raise ActionController::ParameterMissing.new(param) unless payload['repo_info'].key?(param)
+        end
+        payload
+      end
+
+      def verify_update_role_parameters(params)
+        payload = permit_parameters params
+        %w[vcs_url ref].each do |param|
+          raise ActionController::ParameterMissing.new(param) unless payload['repo_info'].key?(param)
+        end
+        payload
+      end
+
+      def start_vcs_task(op_info, operation)
+        case operation
+        when :update
+          job = UpdateAnsibleRole.perform_later(op_info, @proxy)
+        when :install
+          job = CloneAnsibleRole.perform_later(op_info, @proxy)
+        when :delete
+          job = DeleteAnsibleRole.perform_later(op_info, @proxy)
+        else
+          raise Foreman::Exception.new(N_('Unsupported operation'))
+        end
+
+        task = ForemanTasks::Task.find_by(external_id: job.provider_job_id)
+
+        render json: {
+          task: task
+        }, status: :ok
+      rescue Foreman::Exception
+        head :internal_server_error
+      end
+    end
+  end
+end
--- a/app/helpers/foreman_ansible/ansible_roles_helper.rb
+++ b/app/helpers/foreman_ansible/ansible_roles_helper.rb
@@ -31,6 +31,12 @@ module ForemanAnsible
                            ansible_proxy_links(hash))
     end
 
+    def vcs_import
+      select_action_button("",
+                           { :primary => true, :class => 'roles-import' },
+                           link_to(_("Download from Git"), "#vcs_download"))
+    end
+
     def import_time(role)
       _('%s ago') % time_ago_in_words(role.updated_at)
     end
--- a/app/jobs/clone_ansible_role.rb
+++ b/app/jobs/clone_ansible_role.rb
@@ -1,0 +1,12 @@
+class CloneAnsibleRole < ::ApplicationJob
+  queue_as :default
+
+  def humanized_name
+    _('Download Ansible Role from Git')
+  end
+
+  def perform(repo_info, proxy)
+    vcs_cloner = ForemanAnsible::VcsCloner.new(proxy)
+    vcs_cloner.install_role repo_info
+  end
+end
--- a/app/lib/proxy_api/ansible.rb
+++ b/app/lib/proxy_api/ansible.rb
@@ -4,7 +4,7 @@ module ProxyAPI
   # ProxyAPI for Ansible
   class Ansible < ::ProxyAPI::Resource
     def initialize(args)
-      @url = args[:url] + '/ansible/'
+      @url = "#{args[:url]}/ansible/"
       super args
     end
 
@@ -52,6 +52,63 @@ module ProxyAPI
       parse(get("playbooks/#{playbooks_names}"))
     rescue *PROXY_ERRORS => e
       raise ProxyException.new(url, e, N_('Unable to get playbooks from Ansible'))
+    end
+
+    def repo_information(vcs_url)
+      parse(get("vcs_clone/repo_information?vcs_url=#{vcs_url}"))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 400
+        raise Foreman::Exception.new N_('Error requesting repository metadata. Check Smart Proxy log.')
+      else
+        raise
+      end
+    end
+
+    def list_installed
+      parse(get('vcs_clone/roles'))
+    rescue *PROXY_ERRORS
+      raise Foreman::Exception.new N_('Error requesting installed roles. Check log.')
+    end
+
+    def install_role(repo_info)
+      parse(post(repo_info, 'vcs_clone/roles'))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 409
+        raise Foreman::Exception.new N_('A repo with the name %s already exists.') % repo_info['repo_info']&.[]('name')
+      when 400
+        raise Foreman::Exception.new N_('Git Error. Check log.')
+      else
+        raise
+      end
+    end
+
+    def update_role(repo_info)
+      name = repo_info.delete('name')
+      parse(put(repo_info, "vcs_clone/roles/#{name}"))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 400
+        raise Foreman::Exception.new N_('Error updating %s. Check Smartproxy log.') % name
+      else
+        raise
+      end
+    end
+
+    def delete_role(role_name)
+      parse(delete("vcs_clone/roles/#{role_name}"))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 400
+        raise Foreman::Exception.new N_('Error deleting %s. Check Smartproxy log.') % role_name
+      else
+        raise
+      end
     end
   end
 end
--- a/app/views/ansible_roles/index.html.erb
+++ b/app/views/ansible_roles/index.html.erb
@@ -1,7 +1,13 @@
+
+<%= webpacked_plugins_js_for :foreman_ansible %>
+
+<%= csrf_meta_tag %>
+
 <% title _("Ansible Roles") %>
 
-<% title_actions ansible_proxy_import(hash_for_import_ansible_roles_path),
-  documentation_button('#4.1ImportingRoles', :root_url => ansible_doc_url) %>
+<% title_actions ansible_proxy_import(hash_for_import_ansible_roles_path), vcs_import,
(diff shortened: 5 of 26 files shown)


nofaralfasi:
I'll test it again after the tests are successful because it is not working for me now.

Thorben-D:
I fixed the linting issues and retested everything.
Tests are running fine on my machine.

nofaralfasi:
I'm getting the following error from the GUI:
```
Performing CloneAnsibleRole (Job ID: 311cb572-ffe0-43c1-be85-3253315eb4d3) from Dynflow(default) enqueued at 2023-12-14T10:15:52Z with arguments: {"vcs_url"=>"https://github.com/theforeman/foreman.git", "name"=>"theforeman.foreman", "ref"=>"", "update"=>false}, #<GlobalID:0x00007fb5c0bda290 @uri=#<URI::GID gid://foreman/SmartProxy/1>>
Error performing CloneAnsibleRole (Job ID: 311cb572-ffe0-43c1-be85-3253315eb4d3) from Dynflow(default) in 65.03ms: Foreman::Exception (ERF42-7661 [Foreman::Exception]: »Git Error. Check log.«):
```

Thorben-D:
I added the missing permissions. The pipeline should now complete.

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      api :GET, '/vcs_clone/get_installed_roles', N_('Returns an array of roles installed on the provided proxy')
+      formats ['json']
+      param :smart_proxy, Array, N_('Name of the SmartProxy'), :required => true
+      error 400, :desc => N_('Parameter unfulfilled')
+      error 500, :desc => N_('Internal server error')
+      def installed_roles
+        smart_proxy = params.require(:smart_proxy)
+        ansible_proxy = SmartProxy.find_by(name: smart_proxy)
```
ekohl, line 34: You should at least require it to have the Ansible feature:
```suggestion
        ansible_proxy = SmartProxy.with_features('Ansible').find_by(name: smart_proxy)
```

But you can also verify a capability:
```suggestion
        ansible_proxy = SmartProxy.find_by(name: smart_proxy)
        raise Exception unless ansible_proxy.has_capability?('Ansible', 'vcs')
```

app/jobs/clone_ansible_role.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CloneAnsibleRole < ::ApplicationJob
+  queue_as :default
+
+  def humanized_name
+    'Clone Ansible Role from VCS'
```
ekohl, line 7: Aren't these typically translatable?
```suggestion
    _('Clone Ansible Role from VCS')
```

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      rescue_from ActionController::ParameterMissing do |e|
+        render json: { 'error' => e.message }, status: :bad_request
+      end
+
+      rescue_from Git::GitExecuteError do |_e|
+        head :internal_server_error
+      end
+
+      api :GET, '/vcs_clone/get_repo_info', N_('Returns information about the repo')
```
ekohl, line 16: REST encodes the method in the HTTP method, so `get_repo_info` is redundant. See further suggestions in the routes part.

config/routes.rb
```diff
@@ -100,6 +100,14 @@
           end
         end
 
+        resources :vcs_clone, :only => [] do
+          collection do
+            get 'get_repo_information', to: 'vcs_clone#repo_information'
+            post 'install_role', to: 'vcs_clone#install_role'
+            get 'get_installed_roles', to: 'vcs_clone#installed_roles'
```
ekohl, line 107: REST encodes methods in the HTTP method, not in the path. So I'd think this is a more natural API:
```suggestion
            get 'repo_information', to: 'vcs_clone#repo_information'
            post 'roles', to: 'vcs_clone#install_role'
            get 'roles', to: 'vcs_clone#installed_roles'
```

You `POST` to `roles` to create a new one and `GET` to `/roles` to get the installed roles.

Then uninstall would be `DELETE /roles/:id` and updating `PUT /roles/:id`.

foreman_ansible.gemspec
```diff
@@ -22,4 +22,5 @@ Gem::Specification.new do |s|
   s.add_dependency 'deface', '< 2.0'
   s.add_dependency 'foreman_remote_execution', '>= 9.0', '< 13'
   s.add_dependency 'foreman-tasks', '>= 7.0', '< 10'
+  s.add_dependency 'git', '~> 1.0'
```
ekohl, line 25: I'd prefer to avoid the whole git dependency. See more in the overall review.

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      api :POST, '/vcs_clone/install', N_('Launches a task to install the provided role')
+      formats ['json']
+      param :repo_info, Hash, :desc => N_('Dictionary containing info about the role to be installed') do
+        param :vcs_url, String, :desc => N_('Url of the repo'), :required => true
+        param :name, String, :desc => N_('Name of the repo'), :required => true
+        param :ref, String, :desc => N_('Branch / Tag / Commit reference'), :required => true
+        param :update, String, :desc => N_('Whether an existing role with the provided name should be updated'), :default => false
+      end
+      param :smart_proxy, Array, N_('Array of SmartProxies the role should get installed to')
```
ekohl, line 50: Why is this an array, but only the first ID gets used?

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+                      :ref,
+                      :update
+                    ],
+                    smart_proxy: []
+                  )
+
+        payload['repo_info']['update'] = false unless payload['repo_info'].key? 'update'
+
+        # 'smart_proxy' is an array to later allow a role to be installed on multiple SPs
+        ansible_proxy = SmartProxy.find_by(name: payload['smart_proxy'][0])
```
ekohl, line 68: Similar to the above: please verify the feature and capability.

app/controllers/api/v2/vcs_clone_controller.rb
Thorben-D: Thank you very much for the detailed explanation!
It does indeed make sense to only have one of the two plugins be dependent on git.
Adding a new capability to denote cloning functionality is also a nice idea.

app/controllers/api/v2/vcs_clone_controller.rb
Thorben-D: This has been indirectly resolved by passing the proxy-id as a url parameter.

foreman_ansible.gemspec
Thorben-D: All git functionality has been moved to the smartproxy plugin. I removed the dependency.

config/routes.rb
Thorben-D: Indeed, this makes more sense semantically.
I moved the functionality behind a single endpoint with GET, POST, PUT and DELETE methods.

Thorben-D:
I forgot to remove the 'git' dependency and foolishly force pushed again after removing it.
That kind of broke the 'Compare' function...
The actual commit that addresses the feedback from @ekohl is ae450111d4c96a6cacfa7c0e6a727b4263f28df0

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+        task = ForemanTasks::Task.find_by(external_id: job.provider_job_id)
+
+        render json: {
+          task: task
+        }, status: :ok
+      rescue Foreman::Exception
+        head :internal_server_error
+      end
+
+      private
```
ekohl, line 136: There's already a `private` statement on line 80.
```suggestion
```

app/lib/proxy_api/ansible.rb
```diff
@@ -53,5 +53,61 @@ def playbooks(playbooks_names = [])
...
+
+    def repo_information(vcs_url)
+      parse(get("vcs_clone/repo_information?vcs_url=#{vcs_url}"))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 400
+        raise Foreman::Exception.new N_('Error requesting repo-info. Check Smartproxy log.')
+      else
+        raise e
```
ekohl, line 65: I think this causes a different trace back and you can simply use it without arguments to raise it again:
```suggestion
        raise
```

app/lib/proxy_api/ansible.rb
```diff
@@ -53,5 +53,61 @@ def playbooks(playbooks_names = [])
     rescue *PROXY_ERRORS => e
       raise ProxyException.new(url, e, N_('Unable to get playbooks from Ansible'))
     end
+
+    def repo_information(vcs_url)
+      parse(get("vcs_clone/repo_information?vcs_url=#{vcs_url}"))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
```
ekohl, line 60: I'd only catch the specific exception:
```suggestion
    rescue RestClient::RequestFailed => e
```

app/controllers/api/v2/vcs_clone_controller.rb
```diff
@@ -0,0 +1,150 @@
...
+      error 400, :desc => N_('Parameter unfulfilled')
+      def install_role
+        payload = verify_task_parameters(params)
+        start_vcs_task(payload, :install)
+      end
+
+      api :PUT, '/smart_proxies/:proxy_name/ansible/vcs_clone/roles',
+          N_('Launches a task to update the provided role')
+      formats ['json']
+      param :repo_info, Hash, :desc => N_('Dictionary containing info about the role to be installed') do
```
ekohl, line 57: Semantically I'd expect a `PUT` on a collection to ensure that the provided list of items is the exact set of items present. Or put in other words, `PUT` on `.../roles` with `my_role` would ensure that only `my_role` is present and all other roles are removed.

If the name is immutable (and I think it's safe to assume so) you could make it part of the URL and make it `PUT .../roles/:name` to update things (like VCS URL and ref). Just like you did with `DELETE`.

ekohl (changes requested):
Thanks for addressing my concerns. I've only looked at the Ruby side of it, not the front end but I think this generally looks good.

I didn't see it in the PR, but for myself: this relies on https://github.com/theforeman/smart_proxy_ansible/pull/85.

app/lib/proxy_api/ansible.rb
Thorben-D: I had something different in my mind, but just now realized that your suggestion works fine and is cleaner...
I'll add it after the new changes have been reviewed.

app/controllers/api/v2/vcs_clone_controller.rb
Thorben-D: The whole block was useless. Not sure, how I missed that...

app/controllers/api/v2/vcs_clone_controller.rb
Thorben-D: I added `role_name` as a parameter to the URL, which does indeed make more sense now!

Thorben-D:
The latest commit splits up the large React-components into smaller, more manageable ones. The Ruby-part stays untouched.

gardar:
Excited to see some git integration being added for foreman_ansible, but I have few thoughts.
  - How about syncing collections too? Ansible is moving into the direction of collections and away from standalone roles.
  - Wouldn't it have been better to do the integration with ansible-galaxy? ansible-galaxy can clone both roles and collections from git but also fetch them from a galaxy server.

With that being said, do you have any plans to add further git integration with foreman_ansible? Would love to be able to sync the ansible variables to and from a git repo.

OttaviaB:
That is a good idea about collections!
In foreman, collections are still handled in a fundamentally different way than roles though, right?
Roles are imported and assigned to hosts/host groups, while collections must be simply installed once and (at least in my experience) are usually not updated as often as roles.
I guess that the main pain point is with roles for now, but it definetely makes sense to think about collections as well in the future.
 
Using ansible-galaxy instead of git is a great suggestion. Maybe we can do that later?

gardar:
> In foreman, collections are still handled in a fundamentally different way than roles though, right?

Yes and no, a collection can be a collection of roles or even a single role. The collection can also include ansible modules, playbooks, etc. All of this is optional so that the collections don't have to contain roles but they also don't have to contain modules, etc.

> Roles are imported and assigned to hosts/host groups, while collections must be simply installed once and (at least in my experience) are usually not updated as often as roles.

By installing a collection that contains roles on a foreman server you get ansible roles that you can import in foreman and can then assign those to hosts/host groups.

It's absolute nonsense that collections are usually not updated as often as roles, especially when collections can include roles.

When roles are packaged in a collection you can't install just the roles unless you manually copy the roles out of the collection which is something that should be avoided at all cost.
I think it's more trouble than it's worth to do this without using ansible-galaxy as ansible-galaxy already handles the roles and collections correctly.

OttaviaB:
Thanks for your feedback. While this PR aims to streamline a specific workflow we see being commonly used, we will think about whether we want to address your suggestions in a later PR.

sbernhard:
@nofaralfasi Would it be possible for you to have a look at this PR or assign it to someone else?

nofaralfasi:
> @nofaralfasi Would it be possible for you to have a look at this PR or assign it to someone else?

Currently, I am unable to test the PR due to some errors I am encountering (see my comment here: https://github.com/theforeman/smart_proxy_ansible/pull/85/files#r1510263657).

nofaralfasi (commented):
I'm encountering some issues when I navigate to the `ansible/ansible_roles` page in the UI. The new form automatically appears when I navigate to this page (without pressing anything), and the `Import from VCS...` button doesn't always work. 
In addition, when I enter a URL and select a smart-proxy, clicking on the `Examin` button suddenly closes the popup form.

nofaralfasi:
Unfortunately, this is still not working for me as accepted. 
I've messaged @Thorben-D the error I'm encountering. Please advise on how we should proceed from here.

app/lib/proxy_api/ansible.rb
```diff
@@ -53,5 +53,62 @@ def playbooks(playbooks_names = [])
...
+      raise Foreman::Exception.new N_('Error requesting installed roles. Check log.')
+    end
+
+    def install_role(repo_info)
+      parse(post(repo_info, 'vcs_clone/roles'))
+    rescue *PROXY_ERRORS, RestClient::Exception => e
+      raise e unless e.is_a? RestClient::RequestFailed
+      case e.http_code
+      when 409
+        raise Foreman::Exception.new N_('A repo with the name %rName already exists.') % repo_info['name']
```
nofaralfasi, line 81: Shouldn't it be:
```suggestion
        raise Foreman::Exception.new N_('A repo with the name %rName already exists.') % repo_info['repo_info']['name']
```
And I would suggest adding the `&` operator: `repo_info['repo_info']&.[]('name')`.

Thorben-D: Indeed. Also, good call on the safe navigation operator. Did not know it could be used that way.

nofaralfasi:
Here are some suggested improvements to the GUI:
1. Enhance the GUI by including an asterisk next to the mandatory fields.
2. When clicking on the `Examine` button with a non-existing Git link (in the correct format), the following issues occur:
   - No error message is displayed.
   - The application becomes unresponsive and requires a restart to function properly again.
3. Clicking on the `Confirm` button does not close the modal upon successful completion.
4. Enhance the form by adding an icon featuring a question mark, providing additional details about each field when hovered over.

nofaralfasi:
Thanks, @Thorben-D, for the improvements!

Other things I'm noticing:
- The tabs appear disabled even when they are enabled.
- The "SmartProxy required" message should be placed under the smart proxy selection.
- The blue border around the "Collision handling" switch is unnecessary. Additionally, I believe the text of the switch should remain "Update existing" and only the switch should be toggled on/off.

Renaming suggestions:
- Rename "Smart Proxy name" to "Choose Smart Proxy"
- Rename "Import from VCS" to "Clone from Git"
- Rename "Get Ansible roles from VCS" to "Clone Ansible roles from Git"
- "When toggled, Foreman will override already installed roles with the given name." to "When toggled, Foreman will overwrite existing roles with the same name."

app/helpers/foreman_ansible/ansible_roles_helper.rb
```diff
@@ -31,6 +31,12 @@ def ansible_proxy_import(hash)
                            ansible_proxy_links(hash))
     end
 
+    def vcs_import
+      select_action_button("",
```
nofaralfasi, line 35: We should not display this button if the user does not have the required permissions.

nofaralfasi (changes requested):
Some actions, such as listing and deleting, are not directly related to the ability to import Ansible roles from Git. Perhaps I'm overthinking it, but I suggest we should decide whether to move these actions to the `AnsibleRolesController` or keep the structure as is.

MariaAga (changes requested):

webpack/components/VcsCloneModalContent/components/BranchTagSelectionMenu.js
```diff
@@ -0,0 +1,123 @@
...
+            />
+          </FormGroup>
+        </Form>
+      </Tab>
+      <Tab
+        data-testid="BranchTagSelectionMenuBranchTab"
+        eventKey={1}
+        title={<TabTitleText>{__('Branches')}</TabTitleText>}
+        isDisabled={!props.branchTagsEnabled}
+      >
```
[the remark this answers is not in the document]
Thorben-D: This is sort-of happening. Since "main" is the default ref, the user would have to delete it from "manual input" to show the error message.

webpack/components/VcsCloneModalContent/components/GitLinkInputComponent.js
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      isFirsRender.current = false;
+    }
+  }, [textInput, props.smartProxySelection]);
+
+  return (
+    <React.Fragment>
+      <Form>
+        <FormGroup
+          label={__('Git-URL')}
```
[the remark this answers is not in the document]
Thorben-D: Missed this, will do.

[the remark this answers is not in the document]
Thorben-D: Do we have a fitting example for that link?

webpack/components/VcsCloneModalContent/components/SmartProxySelector.js
```diff
@@ -0,0 +1,152 @@
...
+      >
+        <Select
+          variant={SelectVariant.single} // change to typeaheadMulti to allow selection of multiple Smart Proxies
+          onToggle={() =>
+            setIsSmartProxyDropdownOpen(!isSmartProxyDropdownOpen)
+          }
+          onSelect={handleSmartProxySelect}
+          selections={props.smartProxySelection}
+          isOpen={isSmartProxyDropdownOpen}
+          placeholderText={__('Choose Smart Proxy')}
```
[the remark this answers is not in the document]
Thorben-D: Missed this, will do.

app/helpers/foreman_ansible/ansible_roles_helper.rb
Thorben-D: Missed this as well. Will add later.

app/views/ansible_roles/index.html.erb
```diff
@@ -1,7 +1,14 @@
+
+<%= webpacked_plugins_js_for :foreman_ansible %>
+<%= webpacked_plugins_css_for :foreman_ansible %>
```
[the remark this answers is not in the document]
Thorben-D: If i remove these two lines, the component cannot be mounted. Could you point me to a PR where this was deprecated?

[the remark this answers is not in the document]
MariaAga: selecting both was a misclick, only meant to select the css one. it was deprecated here :https://github.com/theforeman/foreman/commit/89c610451c9ea22eb426826e405518e04d68972f

webpack/components/VcsCloneModalContent/components/GitLinkInputComponent.js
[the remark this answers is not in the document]
nofaralfasi: Something like: `https://github.com/example/example.git`

app/views/ansible_roles/index.html.erb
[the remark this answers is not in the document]
Thorben-D: Yeah, that makes sense. I'll remove the one added by me and create a different PR for the rest.

webpack/components/VcsCloneModalContent/components/SmartProxySelector.js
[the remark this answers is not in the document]
Thorben-D: Actually, I am not sure how to handle this... Ideally, this would be handled on the React side.

app/helpers/foreman_ansible/ansible_roles_helper.rb
Thorben-D: Actually, I am not sure how to handle this... Ideally, this would be handled on the React side.

webpack/components/VcsCloneModalContent/components/BranchTagSelectionMenu.js
```diff
@@ -0,0 +1,123 @@
...
+            isRequired
+          >
+            <TextInput
+              id="branch_tag_commit_input"
+              value={props.gitRef}
+              onChange={value => props.setGitRef(value)}
+            />
+          </FormGroup>
+        </Form>
+      </Tab>
```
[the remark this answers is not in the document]
nofaralfasi: I see the UI after the changes now, and I think the previous version looked better.
The default selected tab should be the rightmost one.

nofaralfasi (changes requested):

nofaralfasi:
Since this PR has been inactive for a while, we’re closing it to keep the repository organized. If you’d like to continue working on it, please feel free to reopen or create a new PR.

PR #677, opened 2023-12-18
Fixes #36984 - Exclude all subdirectories for vendor in .rubocop.yaml

dosas:
default config: https://github.com/rubocop/rubocop/blob/master/config/default.yml#L68


stejskalleos:
**LGTM** :green_apple: Thanks @dosas

PR #678, opened 2024-01-01
Fixes #36469 - Fix Ansible Roles and Variables sub-tabs pagination

Ron-Lavi:
Since there is no easy way to do pagination with Graphql at the moment,
we are using `first` and `last` variables in the query.
to make the pagination work on tables where `page * per_page > totalCount`,
we needed to add the following calculation for the `last` variable

### DIFF
--- a/webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/index.js
+++ b/webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/index.js
@@ -17,6 +17,11 @@ const AnsibleVariableOverrides = ({ hostId, hostAttrs, history }) => {
   const hostGlobalId = encodeId('Host', hostId);
   const pagination = useCurrentPagination(history);
   const [totalItems, setTotalItems] = useState(0);
+  const setTotalCount = data => {
+    if (!data) return;
+    const { totalCount } = data.host.ansibleVariablesWithOverrides;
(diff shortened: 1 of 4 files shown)


nofaralfasi:
Clicking the "Next" icon on the Roles tab alters the displayed item count when moving between page.

Ron-Lavi:
@nofaralfasi Hey, can you check again?

nofaralfasi:
The `Variables` tab works well.
However, on the `Roles` tab, if I try to go to the last page, I get `No roles assigned directly to the host`, instead of the list of the remaining roles.

Ron-Lavi:
@nofaralfasi I am having a real difficulty to solve those tests that are failing..
I invested quite some time on trying to fix the tests with no success, I think this implementation relies on updating the URL with pagination search params and that doesn't work well in the test itself, also the mocks are very hardcoded :/
@mariaaga would appreciate some help if you got some time, thanks

MariaAga:
I have some tests failing also on master branch btw

Ron-Lavi:
for visibility, not sure I will have much time to continue work on this PR in the upcoming two weeks..
cc @nofaralfasi @stejskalleos @girijaasoni

nofaralfasi:
The issue can no longer be reproduced. Therefore, I am closing this PR.

PR #681, opened 2024-01-10
Test with Ruby 3

ekohl:
This tries out https://github.com/theforeman/foreman_ansible/pull/519 together with https://github.com/theforeman/foreman/pull/9871 to verify the test suite on Ruby 3.

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -23,3 +23,6 @@ jobs:
     uses: theforeman/actions/.github/workflows/foreman_plugin.yml@v0
     with:
       plugin: foreman_ansible
+      # https://github.com/theforeman/foreman/pull/9989
+      foreman_repository: ofedoren/foreman
+      foreman_version: foreman-on-ruby-3-letsgooo


.github/workflows/ruby_tests.yml
```diff
@@ -1,81 +1,23 @@
...
-        with:
-          name: logs
-          path: log/*.log
-          retention-days: 5
+    uses: theforeman/actions/.github/workflows/foreman_plugin.yml@v0
+    with:
+      plugin: foreman_ansible
+      # https://github.com/theforeman/foreman/pull/9871
+      foreman_repository: ofedoren/foreman
+      foreman_version: feat-36849-add-ruby-3-ga
```
ofedoren, line 23: This should be updated to https://github.com/theforeman/foreman/pull/9989

ekohl:
Rebased to be based on https://github.com/theforeman/foreman/pull/9989.

ekohl:
I wonder why this fails. In the console you see a deprecation warning, but that shoudn't be fatal. In the PostgreSQL log you do see errors about tables not existing so I wonder if that's the real reason.

ofedoren:
Hmm, not sure either. Maybe it's somehow related to webpack? I mean, some test failed for me locally until I run `bundle exec rake webpack:compile`. Otherwise it was green (on Ruby 3 with tasks/dynflow changes, didn't test otherwise).

PR #682, opened 2024-01-15
Fixes #36594 - Fix addition of '\' when editing ansible variables 

girijaasoni:


nofaralfasi (approved):
Fixed the issue for me.

PR #684, opened 2024-01-23
Fixes #37091 - Pretty print long YAML list

Et7f3:
Hello,

I come after https://github.com/Katello/katello/pull/10850. During testing a long list was printed for packages_name.

1) It was hard to read
2) Ruby print list in the same format than YAML compact form. It doesn't seems the original author was aware (I think it expected a String because it called it name and not names)

### DIFF
--- a/app/views/foreman_ansible/job_templates/package_action_-_ansible_default.erb
+++ b/app/views/foreman_ansible/job_templates/package_action_-_ansible_default.erb
@@ -30,8 +30,7 @@ template_inputs:
 provider_type: Ansible
 kind: job_template
 model: JobTemplate
-%>
-
+-%>
 # For Windows targets use the win_package module instead.
 ---
 - hosts: all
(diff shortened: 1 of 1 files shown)


app/views/foreman_ansible/job_templates/package_action_-_ansible_default.erb
```diff
@@ -41,7 +40,7 @@ model: JobTemplate
   <%- end -%>
   tasks:
     - package:
-        name: <%= input('name') %>
+<%= indent(8) { to_yaml({"name" => input('name')}).gsub(/---\n/, '') } -%>
```
Et7f3, line 43: I was able to remove my erb loop/if pasta code. `to_yaml` macro doesn't seems to support `:UseHeader => true` https://github.com/ruby/ruby/blob/4099b04c099899d92dab3f34f1152038ed718d79/test/psych/helper.rb#L51

nofaralfasi:
@Et7f3, I apologize for the delay.
Could you please rebase? Currently the tests are broken, but once they're fixed I will merge this PR.

Et7f3:
> for the delay

No issue it is OSS.

> Could you please rebase?

It was just to retrigger the CI ? I don't see conflict with it.

nofaralfasi:
Thanks @Et7f3!

PR #687, opened 2024-01-28
fix concurrency group

evgeni:
head_ref doesn't exist on push triggered workflows


ekohl (approved):
Failure looks like what we saw elsewhere and needs to use node_modules in the path, but that's unrelated to this change

PR #688, opened 2024-01-28
don't use tilde to import from node_modules, sass-loader will do

evgeni:


evgeni:
why does `webpack:compile` fail even tho webpack only generates warnings, no errors?

:eyes: @MariaAga

evgeni:
the interesting thing is: Debian packages build fine without it, need to figure out what they do differently

PR #691, opened 2024-02-02
Fixes #37136 - SSH User is ignored in advanced field of job invocation

OttaviaB:
**Relevant Bug**

This pull request fixes Bug #37136 - SSH User is ignored in advanced field of job invocation .

**Details**
The advanced field for the SSH user in the job invocation is ignored by `foreman_ansible`.
I fixed it so that now the priority for the SSH user is:
1. Advanced field in job invocation
2. Host parameter
3. Global setting

### DIFF
--- a/app/services/foreman_ansible/inventory_creator.rb
+++ b/app/services/foreman_ansible/inventory_creator.rb
@@ -97,7 +97,7 @@ module ForemanAnsible
 
     def remote_execution_options(host)
       params = {
-        'ansible_user' => host_setting(host, 'remote_execution_ssh_user'),
+        'ansible_user' => @template_invocation.job_invocation&.ssh_user || host_setting(host, 'remote_execution_ssh_user'),
         'ansible_become_method' => host_setting(host, 'remote_execution_effective_user_method'),
         'ansible_ssh_private_key_file' => ansible_ssh_private_key(host),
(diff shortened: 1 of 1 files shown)


app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -97,7 +97,7 @@ def ansible_extra_options(host)
 
     def remote_execution_options(host)
       params = {
-        'ansible_user' => host_setting(host, 'remote_execution_ssh_user'),
+        'ansible_user' => @template_invocation.job_invocation&.ssh_user || host_setting(host, 'remote_execution_ssh_user'),
```
adamruzicka, line 100: If I'm reading this right, then `@template_invocation` can be `nil` which could then raise. I know it is sort of guarded on L55, but it would be nice if we could stay on the safe side of things.

adamruzicka (commented):
Works well, left one nitpick to consider

app/services/foreman_ansible/inventory_creator.rb
OttaviaB: Should be better now!

adamruzicka (approved):

sbernhard (approved):

sbernhard:
Failing tests looks unrelated.

stejskalleos:
Thanks @OttaviaB @sbernhard @adamruzicka

PR #692, opened 2024-02-06
Register template dirs

stejskalleos:
See Foreman core PR for the usage: https://github.com/theforeman/foreman/pull/10038/files


stejskalleos:
(Doesn't work without merged Foreman PR first)

PR #694, opened 2024-03-01
Release gem GH action

stejskalleos:


.github/workflows/release_gem.yml
```diff
@@ -0,0 +1,14 @@
...
+    ref_type: tag
+
+jobs:
+  release:
+    name: Release gem
+    uses: theforeman/actions/.github/workflows/release-gem.yml@v0
+    with:
+      allowed_owner: theforeman
+    secrets:
+      api_key: ${{ secrets.RUBYGEM_API_KEY }}
```
evgeni, line 14: this needs to be generated and set in the repo settings.

@ekohl I recall you had created dedicated tokens for each gem. is it documented somewhere how to do that?

evgeni (approved):

.github/workflows/release_gem.yml
ekohl: No, but it's fairly straight forward. You go to https://rubygems.org/profile/api_keys and you'll see API keys named `release $gem` scoped to a single gem. https://rubygems.org/profile/api_keys/new should make that rather easy.

https://guides.rubygems.org/api-key-scopes/#creating-from-gem-cli says you can also create them from the CLI with `gem signin`, which is IMHO a non-obvious name.

There's also https://github.com/rubygems/release-gem which is supposed to be the proper thing but I found that to be hard to integrate. It's not possible to use from a reusable action, so you end up copying complex flows. There's also no API to set up each gem and you must use the bundler release tasks.

evgeni: created and added to the repo config

ekohl (approved):

PR #696, opened 2024-03-06
Fixes #37228 - Replace Ansible/Roles page with React component

Thorben-D:
[Feature #37228](https://projects.theforeman.org/issues/37228)

I am currently working on a different feature, for which I need to modify the roles table.
Since my new feature will also use React, I found it to be easier to just rewrite the whole table as a React component, than to shoehorn React into the existing one.

The new component acts as a drop-in replacement for the previously used erb-table.

centos8-katello-devel-stable example com_ansible_ansible_roles_order=name+asc


sbernhard:
What is blocking us from using this change? Possibly at the start of a new release cycle?

nofaralfasi:
The functionality works well. However, the page is loading much slower than before. I’m wondering if it’s an issue on my side or if the new code is causing the app to load slower.

Thorben-D:
@nofaralfasi 
Unfortunately, this is a side-effect of React rendering the component on the client-side. I am unsure when exactly react-rails renders components, but it feels pretty late (c.f. `/new/hosts`). Rendering server-side would help in this regard, but that does not work when `window` or `document` are used.

nofaralfasi (changes requested):

webpack/components/AnsibleRoles/components/AnsibleRolesTableRowActionButton.js
```diff
@@ -0,0 +1,65 @@
...
+  const dropdownItems = [
+    <DropdownItem
+      key="delete_action"
+      component="a"
+      data-method="delete"
+      href={`/ansible/ansible_roles/${props.ansibleRoleId}`}
+      data-confirm={sprintf(__('Delete %s?'), props.ansibleRoleName)}
+    >
+      {__('Delete')}
+    </DropdownItem>,
```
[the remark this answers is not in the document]
Thorben-D: AFAIK, the role that handles that permission is "Ansible Roles Manager", which is also required to view **Configure > Ansible > Roles**.
An unauthorized user is prevented from seeing the whole page.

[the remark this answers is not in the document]
nofaralfasi: That's correct. However, you can set your own specific permissions. We need to cover both cases.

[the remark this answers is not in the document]
Thorben-D: Ah, I see... In that case that is indeed an issue. The problem is that permission handling capabilities are severely lacking on the React side.
It seems to me that this may only be accomplished with workarounds for now. See these PRs by kmalyjur:
https://github.com/theforeman/foreman_remote_execution/pull/842/files
https://github.com/theforeman/foreman/pull/9902

While that functionality has been merged, I'd be inclined to wait a bit before taking that approach because of this pending RFC:
https://community.theforeman.org/t/handling-permissions-whats-our-approach/36858

sbernhard:
Of course, the site load would be pretty fast but rendering the page takes longer as its done in the browser itself. 
Should it have a spinner as this was done in Katello/Content-views for example?

ekohl (commented):
In Foreman we have the `TableIndexPage` component which is intended for these index pages. Have you considered using that and if so, why was it insufficient?

app/views/ansible_roles/index.html.erb
```diff
@@ -1,47 +1,19 @@
+<%= webpacked_plugins_js_for :foreman_ansible %>
+<%= webpacked_plugins_css_for :foreman_ansible %>
```
MariaAga, line 2: `webpacked_plugins_css_for` is deprecated, plugin css is already loaded.'

MariaAga (commented):
+1 to Ewouds question, the js loading is slower between ruby and js, but fater between full React pages.
Can this be a full react page? it will need to be added in `webpack/routes/routes.js`, `config/routes.rb`

app/views/ansible_roles/index.html.erb
Thorben-D: I think it would be best if #715 could be merged, so I can rebase this branch.

Thorben-D:
Thank you for your feedback. I would actually prefer a full React page as-well! I chose this approach to keep the changes as low-profile as possible, but I am happy to move it to a full page.
Regarding `TableIndexPage`, I'm afraid, I couldn't get it to work... Do you have an example where this was used outside of foreman? (The page not the hooks!)

MariaAga:
Webhooks use it here: https://github.com/theforeman/foreman_webhooks/blob/master/webpack/ForemanWebhooks/Routes/Webhooks/WebhooksIndexPage/WebhooksIndexPage.js

Thorben-D:
Thanks, that's what I was looking for.
I checked out the component and the biggest issue I see with it is the way the action buttons are handled.
The old page allows the selection of a smart proxy to import roles from through a dropdown button.
While `ActionButtons` does provide a dropdown, it is toggled by a "kebab" toggle and not a component with text on it.

Thorben-D:
^^Nevermind, I just noticed the `customToolbarItems` prop...

Thorben-D:
Sorry for the holdup, I was occupied otherwise.
I re-wrote the whole thing to use the `TableIndexPage` component. I decided to adapt the `ui_ansible_roles` endpoint for this, as I did not find any usages.
Tests are still missing, but I will of course add some once we agreed on the UI.

app/views/ui_ansible_roles/show.json.rabl
```diff
@@ -1,3 +1,18 @@
 object @ansible_role
 
-extends "api/v2/ansible_roles/show"
+attributes :id, :name
+code :hostgroupsCount do |role|
```
ekohl, line 4: I don't think we use camelCase elsewhere in our APIs. We do have at least one example of snake_case specifically for counts:
https://github.com/theforeman/foreman/blob/b9c2dd00bed537b73bbad5ba529130e8717dfabb/app/views/api/v2/smart_proxies/main.json.rabl#L5

app/views/ui_ansible_roles/show.json.rabl
```diff
@@ -1,3 +1,18 @@
...
+end
+code :variablesCount do |role|
+  role.ansible_variables.count
+end
+code :importTime do |role|
+  import_time(role)
+end
+code :can_delete do |role|
+  authorized_for(:controller => UiAnsibleRolesController, :action => :destroy)
+end
```
ekohl, line 18: Please ensure line endings in files

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,68 @@
...
+    dropdownItems = [
+      <DropdownItem key="loading_item">{__('Loading...')}</DropdownItem>,
+    ];
+  } else {
+    dropdownItems = results
+      .filter(result =>
+        result.features.some(feature =>
+          feature.capabilities.includes('ansible')
+        )
+      )
```
ekohl, line 38: This filters Smart Proxies with the Ansible feature on the client side. Can we please filter server side? The API should support `feature=Ansible`.

app/controllers/ui_ansible_roles_controller.rb
```diff
@@ -7,6 +10,12 @@ def index
     @ui_ansible_roles = resource_scope_for_index(:permission => :view_ansible_roles)
   end
 
+  api :DELETE, '/ui_ansible_roles/:id', N_('Deletes Ansible role')
```
ekohl, line 13: I don't see any permission filtering here. Am I missing something?

Also, couldn't this just use the regular REST API to delete roles?

ekohl (changes requested):

app/views/ui_ansible_roles/show.json.rabl
Thorben-D: First of all, thank you very much for the swift feedback!

Changed to snake_case. Force of JS habit...

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: Unfortunately, I couldn't get this to work...

app/controllers/ui_ansible_roles_controller.rb
Thorben-D: Indeed, I forgot to add the permission.

The table component uses the same endpoint for querying and deleting items, so it's unfortunately not possible to use the API just for deletion.

app/views/ansible_roles/index.html.erb
ekohl: That has been merged.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,66 @@
...
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import { importUrl } from '../AnsibleRolesPageHelpers';
+
+export const AnsibleRolesImportButton = () => {
+  let dropdownItems;
+
+  const {
+    response: { results },
+    status,
+  } = useAPI('get', '/api/v2/smart_proxies');
```
ekohl, line 13: Following up on https://github.com/theforeman/foreman_ansible/pull/696#discussion_r1636410033 I'd imagine this  should work:
```suggestion
  } = useAPI('get', '/api/v2/smart_proxies', { params: { feature: 'Ansible' }});
```

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,66 @@
...
+  const onFocus = () => {
+    const element = document.getElementById('toggle-primary');
+    element.focus();
+  };
+  const onSelect = () => {
+    setIsOpen(false);
+    onFocus();
+  };
+
+  if (status !== 'RESOLVED') {
```
ekohl, line 28: I'm a fan of explicit error handling instead. Otherwise it keeps "loading" forever.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,66 @@
...
+    dropdownItems = [
+      <DropdownItem key="loading_item">{__('Loading...')}</DropdownItem>,
```
ekohl, line 30: Should this also use `isDisabled`?

app/views/ansible_roles/index.html.erb
ekohl, line 1: This page still does things. Is that really still needed? Like `title_actions` or pagination. Or is this whole template even used at all?

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: Hmm, for me this still returns the `centos8-katello-devel-stable.example.com` Smart Proxy...
I don't want to rule out though that this is caused by environment being broken.

ekohl: Doesn't that proxy just have that feature? Could also be a bug in the API, in which case we'd need to fix that as well.

app/controllers/ui_ansible_roles_controller.rb
```diff
@@ -7,6 +10,12 @@ def index
...
+  param :id, :identifier, :required => true
+  def destroy
+    AnsibleRole.find_by(id: params[:id]).destroy
```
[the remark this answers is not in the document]
Thorben-D: Fixed, but the safe navigation operator could not be used, because then a 2XX code is returned, which leads to 
"\<role_name\> was successfully deleted" being shown in the UI.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: I am not sure. Infrastructure > Smart Proxies doesn't say that the Smart Proxy supports the Ansible feature and neither does the UI, but maybe I am missing something...
I have committed your suggestion and commented out the client-side filtering for now and will re-test this on a fresh machine.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: Applied

app/views/ansible_roles/index.html.erb
Thorben-D: I guess that would depend on whether or not we want to keep it for legacy reasons. 
If not, the view and routes may be removed.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
```diff
@@ -0,0 +1,34 @@
...
+import { getDocsURL } from 'foremanReact/common/helpers';
+import { AnsibleRolesImportButton } from './components/AnsibleRolesImportButton';
+import { getActions } from '../AnsibleRoles/AnsibleRolesPageHelpers';
+
+export const AnsibleRolesPage = () => (
+  <TableIndexPage
+    header={__('Ansible Roles')}
+    controller="ansible_roles"
+    apiUrl="/ansible/ui_ansible_roles"
+    apiOptions={{ key: 'WEBHOOKS_API_REQUEST_KEY' }}
```
MariaAga, line 13: Please change to an ansible key instead of webhooks

webpack/components/AnsibleRoles/AnsibleRolesPage.js
```diff
@@ -0,0 +1,34 @@
...
+    hasHelpPage
+    creatable={false}
+    isDeleteable
+    searchable
+    customToolbarItems={
+      <AnsibleRolesImportButton>Primary</AnsibleRolesImportButton>
```
MariaAga, line 19: Primary should be translated

webpack/components/AnsibleRoles/AnsibleRolesPage.js
```diff
@@ -0,0 +1,34 @@
...
+      'Managing_Configurations_Ansible',
+      'Importing_Ansible_Roles_and_Variables_ansible'
+    )}
+    rowKebabItems={getActions}
+    columns={{
+      import_time: { title: __('Imported at'), isSorted: true },
+      variables_count: { title: __('Variables') },
+      hosts_count: { title: __('Hosts') },
+      hostgroups_count: { title: __('Hostgroups') },
+      name: { title: __('Name'), isSorted: true },
```
MariaAga, line 31: Do we want to use a new order, or should this be re-ordered to fit the old page?

webpack/routes/AnsibleRoles/index.js
```diff
@@ -0,0 +1,16 @@
...
+import { AnsibleRolesPage } from '../../components/AnsibleRoles/AnsibleRolesPage';
+
+const AnsibleRoles = () => (
+  <React.Fragment>
+    <Helmet>
+      <title>{__('Ansible Roles')}</title>
+    </Helmet>
+    <AnsibleRolesPage />
+  </React.Fragment>
+);
```
MariaAga, line 14: No need to wrap it and add Helmet as its provided by TableIndexPage

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,80 @@
...
+  } else if (status === 'ERROR') {
+    showToast({
+      type: 'error',
+      message: __('There was an error requesting Smart Proxies'),
+    });
+  } else if (status === 'RESOLVED') {
+    dropdownItems = results
+      // .filter(result =>
+      //  result.features.some(feature => feature.name === 'Ansible')
+      // )
```
MariaAga, line 50: This seem to filter things correctly for me

MariaAga (changes requested):

app/controllers/ui_ansible_roles_controller.rb
```diff
@@ -7,6 +9,16 @@ def index
     @ui_ansible_roles = resource_scope_for_index(:permission => :view_ansible_roles)
   end
 
+  api :DELETE, '/ui_ansible_roles/:id', N_('Deletes Ansible role')
+  param :id, :identifier, :required => true
+  def destroy
+    role = AnsibleRole.find_by!(id: params[:id])
+    if role.nil?
```
ekohl, line 16: `find_by!` raises an exception, so this code is unreachable

app/views/ansible_roles/index.html.erb
ekohl: I don't see a benefit of keeping it if the table in it isn't rendered. Things like the pagination can only break things IMHO

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
```diff
@@ -0,0 +1,80 @@
...
+} from '../../AnsibleHostDetail/components/AnsibleVariableOverrides/EditableActionHelper';
+
+export const AnsibleRolesImportButton = () => {
+  let dropdownItems;
+
+  const {
+    response: { results },
+    status,
+  } = useAPI('get', '/api/v2/smart_proxies', {
+    params: { feature: 'Ansible' },
```
ekohl, line 19: Sorry, this should have been a search query
```suggestion
    params: { search: 'feature=Ansible' },
```

app/views/ansible_roles/index.html.erb
Thorben-D: I reorganized the routes and now the page loads much faster!
I guess the next step would be to adapt the RolesController to be client-side routing only. The only actions it is dealing with now are related to importing, which is already a React component. 
This is, IMO, out of scope for this PR though.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
Thorben-D: Whoops, that shouldn't be there...

webpack/components/AnsibleRoles/AnsibleRolesPage.js
Thorben-D: ...and neither should this.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
Thorben-D: For me, this arrangement orders the columns the same way they have been on the old version.

webpack/routes/AnsibleRoles/index.js
Thorben-D: Removed

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: It did, but as Ewoud suggested, this filtering can be done server-side.

app/controllers/ui_ansible_roles_controller.rb
Thorben-D: Ah, I was wondering about that. Apparently, the RecordNotFound-Exception still produces a 404 error, so this isn't needed anyways.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js
Thorben-D: Yeah, this works now.

ekohl:
Test failure looks related:
```
[MinitestRetry] retry 'test_: AnsibleRolesController::GET #index should render template index. ' count: 1,  msg: ActionController::MissingExactTemplate: AnsibleRolesController#index is missing a template for request formats: text/html
```

Other than that I think this looks pretty good, though I'll leave the JS review to @MariaAga.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
MariaAga: the old version is arranged opposite way, with name being the first/most left, or am I missing something?
 ```
 <th class="col-md-6"><%= sort :name, :as => s_("Role|Name") %></th>
      <th class="col-md-2"><%= _("Hostgroups") %></th>
      <th class="col-md-2"><%= _("Hosts") %></th>
      <th class="col-md-2"><%= _("Variables") %></th>
      <th class="col-md-2"><%= sort :updated_at, :as => _("Imported at") %></th>
      <th class="col-md-2"><%= _("Actions") %></th>
```
      
image
image

Comment (Thorben-D) [reply webpack/components/AnsibleRoles/AnsibleRolesPage.js:29]: Now this I don't understand...
image

Comment (MariaAga) [reply webpack/components/AnsibleRoles/AnsibleRolesPage.js:29]: This is fun, I'll try to see why it happens next week

**Review by ekohl — COMMENTED** [review]
I realized this in a different PR, but it also applies here. How do you deal with pagination in your API calls?

**Comment by Thorben-D** [comment]
@ekohl 
In this case, pagination is handled by `TableIndexPage`. It does this by using the `Pagination` component, which, through a couple of functions, adjusts the `page` parameter used during the API call.

File: webpack/components/AnsibleRoles/components/AnsibleRolesImportButton.js [hunk 18]
```diff
@@ -0,0 +1,72 @@
...
+
+export const AnsibleRolesImportButton = () => {
+  let dropdownItems;
+
+  const {
+    response: { results },
+    status,
+  } = useAPI('get', '/api/v2/smart_proxies', {
+    params: { search: 'feature=Ansible' },
+  });
```
ekohl, line 16: This is the regular API and returns paginated results. If there are more proxies than fits on a default page, some will be invisible to the user. The previous page used internal APIs that don't paginate, so it would work. It may not be great UX to have a huge list, but this would still be a regression.

Thorben-D: I added `per_page=all` to the API call.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
MariaAga: Looks like firefox and chrome implement `sort` differently, I'll find a fix in foreman core. so this is fine

MariaAga: Can you add weight key to each title?  just to ease the foreman core fix. Lower weights are displayed first.

test/functional/ansible_roles_controller_test.rb
```diff
@@ -2,19 +2,4 @@
...
-  test 'should destroy role' do
-    assert_difference('AnsibleRole.count', -1) do
-      delete :destroy,
-             :params => { :id => @role.id },
-             :session => set_session_user
-    end
-    assert_redirected_to ansible_roles_url
-  end
-end
+class AnsibleRolesControllerTest < ActionController::TestCase; end
```
ekohl, line 5: If it's an empty class, is there any point in keeping it at all? I'd also note that there are system tests (https://guides.rubyonrails.org/testing.html#system-testing) that can use a real browser. I'd love to still have some tests.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
```diff
@@ -0,0 +1,32 @@
...
+  <TableIndexPage
+    header={__('Ansible Roles')}
+    controller="ansible_roles"
+    apiUrl="/ansible/ui_ansible_roles"
+    apiOptions={{ key: 'ANSIBLE_ROLES_API_REQUEST_KEY' }}
+    hasHelpPage
+    creatable={false}
+    isDeleteable
+    searchable
+    customToolbarItems={<AnsibleRolesImportButton />}
```
MariaAga, line 18: Should only be visible if a user has a `import_ansible_roles` permission

Thorben-D: Good idea, fixed.

Thorben-D:
I fixed the issue with sorting by time. Unfortunately, this way, we lose the nicely formatted "<> ago" version, but I suppose this is fine for now.
It would not be unreasonably hard to add the ability to format a column as time to the `Table` component. This could then use a function similar to `import_time` to get the "ago" format.
Unless we already have such a function in core, I would propose [moment.js](https://momentjs.com/), which is also used by Patternfly.

test/functional/ansible_roles_controller_test.rb
Thorben-D: I suppose I could remove it, but I would like to have a test for the `format` action at some point.
I will definitely add tests for the code changed in this PR though, just wanted us to agree on the implementation details first.

webpack/components/AnsibleRoles/AnsibleRolesPage.js
Thorben-D: Ah yes, I looked into this a bit as well.
Apparently, ES2019 just mandates that the sorting algorithm must be stable, but the actual implementation of `sort` is up to the developers. 

When looking at the sorting function in [getColumnHelpers of helpers.js](https://github.com/theforeman/foreman/blob/8283db898b7e145e59dd376ac0b4cb2d30f7b436/webpack/assets/javascripts/react_app/components/PF4/TableIndexPage/Table/helpers.js#L26C1-L44C3) with a debugger, one can see that the order in which the items in the array are consumed differs between FireFox and Chromium based browsers:

FireFox:

1.  a: "import_time"
b: "variables_count"
2. a: "variables_count"
b: "hosts_count"
3. a: "hosts_count"
b: "hostgroups_count"
4. a: "hostgroups_count"
b: "name"

Chromium:

1.  a: "variables_count"
b: "import_time"
2. a: "hosts_count"
b: "variables_count"
3. a: "hostgroups_count"
b: "hosts_count"
4. a: "name"
b: "hostgroups_count"

This can be reproduced by this simple example:
```js
const test_array = ["a", "b", "c", "d", "e"];
console.log(test_array.sort((a, b) => 1));
```

A fix could be to have the sorting function return `0` if no weights are applied, as that produces the same order as the unsorted array.
Either way, adding weights to the columns solved this on both platforms.

MariaAga:
You can do:
```js

import RelativeDateTime from 'foremanReact/components/common/dates/RelativeDateTime';
...
      updated_at: {
        title: __('Imported at'),
        isSorted: true,
        weight: 4,
        wrapper: ({ updated_at: updatedAt }) => (
          <RelativeDateTime date={new Date(updatedAt)} />
        ),
      },
```

webpack/components/AnsibleRoles/components/AnsibleRolesImportButtonWrapper.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+    return null;
+  } else if (status === 'RESOLVED') {
+    if (
+      results.some(permission => permission.name === 'import_ansible_roles')
+    ) {
+      isButtonDisabled = false;
+    }
+  }
+  return <AnsibleRolesImportButton isDisabled={isButtonDisabled} />;
```
MariaAga, line 27: We usually hide buttons if the user has no permission, so returning null instead of a disabled button. Can you send the permission as part of the ui_ansible_roles api, to avoid extra api call

Thorben-D: The button is now hidden instead of just disabled.
I'm afraid, integrating the request for the permission into the request to fetch roles is not easily possible as it is made by TableIndexPage.

MariaAga (approved):
UI code looks good to me

webpack/components/AnsibleRoles/components/AnsibleRolesImportButtonWrapper.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { useAPI } from 'foremanReact/common/hooks/API/APIHooks';
+import { showToast } from '../../../toastHelper';
+import { AnsibleRolesImportButton } from './AnsibleRolesImportButton';
+
+export const AnsibleRolesImportButtonWrapper = () => {
```
ekohl, line 7: This is an interesting approach. We don't have this in Foreman, but now I'm wondering if this should be a common pattern. For example, provide a `hasPermission(permission)` function.

Linking back to my earlier approach, another one could be that you return the URL in the API response if the user has permissions. If they don't, then you can derive that in the frontend without checking permissions manually.

ekohl (commented):
These comments mostly look at the general API integration and don't need to be blockers to merging.

webpack/components/AnsibleRoles/components/AnsibleRolesImportButtonWrapper.js
MariaAga: Theres a discussion on permissions here: https://community.theforeman.org/t/handling-permissions-whats-our-approach/36858 BTW

webpack/components/AnsibleRoles/AnsibleRolesPageHelpers.js
```diff
@@ -0,0 +1,38 @@
...
+    ),
+    isDisabled: false,
+  },
+];
+
+const hostgroupsUrl = roleName => `/hostgroups${searchString(roleName)}`;
+const hostsUrl = roleName => `/hosts${searchString(roleName)}`;
+const variablesUrl = roleName => `ansible_variables${searchString(roleName)}`;
+const searchString = roleName =>
+  `?search=${encodeURIComponent(`ansible_role = ${roleName}`)}`;
```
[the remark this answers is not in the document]
Thorben-D: I would very much welcome such functionality in `foremanUrl`!
As you already said, using hard-coded URLs like I did here is not the most maintainable and the "JS way" is kind of an eye-sore if used many times...

webpack/components/AnsibleRoles/components/AnsibleRolesImportButtonWrapper.js
Thorben-D: Oh yes, I almost forgot about that post. I actually found a solution to the permissions problem that I think holds promise and am currently implementing a proof of concept.

Thorben-D:
I added some tests for the React components. 
I decided to omit a test for the `AnsibleRolesPage` component, as it is basically just a static wrapper for `TableIndexPage`, which is already extensively tested.

nofaralfasi (changes requested):
The page is significantly improved now. Thank you, @Thorben-D, for your hard work!

Here are a few notes from me:

- The "Import from" button is positioned on the same row as the top navigation, causing it to appear centered on the page.
- Table rows have a clickable cursor, but clicking them does not trigger any action.
- Clicking the Documentation button opens it in the current window.
- If the current user lacks permissions to view hosts or hostgroups, the corresponding options in the action menu should be disabled (same as we have now for the delete option).
- When the current user has permissions only to import roles (without viewing permissions), loading the Ansible Roles page results in the following error message: "There was an error requesting Smart Proxies," and the smart-proxy cannot be selected.

MariaAga:
> The "Import from" button is positioned on the same row as the top navigation, causing it to appear centered on the page.

This is fixed in foreman core

> Table rows have a clickable cursor, but clicking them does not trigger any action.
Clicking the Documentation button opens it in the current window

Those are issues from foreman core

sbernhard:
@nofaralfasi / @ekohl  This PR is now open since Mar 6 and we invested a lot to improve foreman_ansible here. @Thorben-D answered all the questions and worked hard to get it done. 
What is preventing us from merging this?

nofaralfasi:
@Thorben-D Could you please rebase the PR? I'll re-check the permission issue, and let's see if the tests pass after the rebase.

Thorben-D:
Thank you for having another look! 
Indeed, there was an issue with permissions again. Apparently, the `authorized_for` method I used before did not behave correctly. I replaced it with a different solution.
Test failure is also fixed.
Sorry about the linting, I am getting some psych error once again on my test-setup. It's nice though that linting issues are now shown in the diff!

app/controllers/ansible_roles_controller.rb
```diff
@@ -6,18 +6,9 @@ class AnsibleRolesController < ::ApplicationController
   include ForemanAnsible::Concerns::ImportControllerHelper
   include ::ForemanAnsible::AnsibleRolesDataPreparations
   def index
-    @ansible_roles = resource_base.search_for(params[:search],
-                                              :order => params[:order]).
-                     paginate(:page => params[:page],
-                              :per_page => params[:per_page])
   end
 
   def destroy
```
ekohl, line 11: If this is a dummy, should the whole method remain?

test/functional/ui_ansible_roles_controller_test.rb
```diff
@@ -11,4 +11,11 @@ class UiAnsibleRolesControllerTest < ActionController::TestCase
     res = JSON.parse @response.body
     assert_equal res['total'], res['results'].size
   end
+
+  test 'should destroy role' do
+    delete :destroy,
+           :params => { :id => @role.id },
+           :session => set_session_user
+    assert_response :success
+  end
```
nofaralfasi, line 20: Can we add a test to verify that a user without the `destroy_ansible_roles` permission cannot perform this action?

nofaralfasi:
Overall, it looks good to me. Once Ewoud's and my comments are addressed, I believe we should merge this

webpack/components/AnsibleRoles/AnsibleRolesPage.js
```diff
@@ -0,0 +1,40 @@
...
+export const AnsibleRolesPage = () => (
+  <TableIndexPage
+    header={__('Ansible Roles')}
+    controller="ansible_roles"
+    apiUrl="/ansible/ui_ansible_roles"
+    apiOptions={{ key: 'ANSIBLE_ROLES_API_REQUEST_KEY' }}
+    hasHelpPage
+    creatable={false}
+    isDeleteable
+    searchable
```
nofaralfasi, line 18: By default, the `searchable` property is already set to true. Do you think we should keep it here for readability?

Thorben-D: Removed

app/controllers/ansible_roles_controller.rb
Thorben-D: Removed the empty methods.

test/functional/ui_ansible_roles_controller_test.rb
Thorben-D: I am not sure if this is necessary. The permissions should already be tested by a test in core and I didn't really find an existing test that tested this.

app/controllers/ui_ansible_roles_controller.rb
Thorben-D: Unfortunately, this would require significant changes to TableIndexPage.js and Table.js in core, as the idea is that TableIndexPage represents an interface to a single controller that has the necessary CRUD methods.
One could add another prop, `deletionURL` to TableIndexPage and Table that would take precedence over the original URL for deletion but I am not sure that is a cleaner solution.

nofaralfasi: Can't we add the `Delete` option to the `rowKebabItems` action items instead of using the `isDeleteable` property? From there, we can use the URL of the existing API for the delete action.

Thorben-D: Unfortunately, we would lose the delete modal that way.

test/functional/ui_ansible_roles_controller_test.rb
sbernhard: @MariaAga what is your thought about this?

nofaralfasi:
I would expect to receive a proper error message when the user doesn't have the necessary permissions to view this page.
image

Thorben-D:
@nofaralfasi The server actually sends back the following message:
"Missing one of the required permissions: view_ansible_roles"
Unfortunately though, the useAPI hook, currently, does not store this information in the store. 
Given that a user does not normally get to this stage via the navigation, I think this is acceptable for now.

nofaralfasi:
@Thorben-D what about permissions to view hosts or hostgroups? I think we should disable the buttons that are not permitted in the action menu.

test/functional/ui_ansible_roles_controller_test.rb
MariaAga: Not my area, but I couldnt find a test in core for it in core

Thorben-D:
Apologies, I did not see the last comment. I updated the design to match the desired outcome. Links in the columns will only be given if a user has the authorization to view them.

PR #700, opened 2024-03-13
Fixes #37250 - Ensure js tests are executed on default branch

dosas:
I would like to address the situation that a change in foreman was breaking this plugin and it went unnoticed for nearly two months.

The best way to fix this integration problem would probably be nightly test runs.

This PR suggests to at least run the js tests when a branch is merged into (pushed to) master. This way possible errors are recognized by a failing default branch and development on PRs is not blocked by 'slow' js tests.

What do you think @nofaralfasi @evgeni ?

Obviously this requires https://github.com/theforeman/foreman_ansible/pull/698 to work.

### DIFF
--- a/.github/workflows/js_tests.yml
+++ b/.github/workflows/js_tests.yml
@@ -1,10 +1,9 @@
 name: JavaScript Testing
 on:
   pull_request:
-    paths:
-      - 'webpack/**'
-      - 'package.json'
-      - '.github/workflows/js_tests.yml'
+  push:
+    branches:
+      - "master"
 
 jobs:
   test_js:


evgeni:
I would prefer if we *also* would run JS tests on every PR, not only those touching webpack.

nofaralfasi:
@dosas can you please update and rebase?

dosas:
> @dosas can you please update and rebase?

done

evgeni (approved):

evgeni:
Node JS 12 tests fail as Core doesn't support Node 12 for a while now ;)

evgeni:
As an alternative, how about #701 that utilizes the shared workflow and uses the "right" node versions ;)

dosas:
> As an alternative, how about #701 that utilizes the shared workflow and uses the "right" node versions ;)

Fine with me but I think that this is mixing two topics into one commit: One is using shared GHA and the other is when to execute the tests

evgeni:
True. We can merge this one and then rebase #701 ontop.

@nofaralfasi ok?

nofaralfasi:
> True. We can merge this one and then rebase #701 ontop.
> 
> @nofaralfasi ok?

Yes please

dosas:
@evgeni since #701 was already merged with these changes we can close this, right?

evgeni:
I agree

PR #701, opened 2024-03-20
use shared GHA workflow for JS tests

evgeni:


nofaralfasi (approved):
Failing tests are not related. 
Thank you @evgeni!

PR #702, opened 2024-03-21
Send resource as keyword argument in resource_name

ekohl:
This was always a better way, but in Foreman 3.10 the args and kwargs handling was made more explicit.


adamruzicka:
This didn't seem to help for me, but changing `resource_scope` did

```
  def resource_scope(**kwargs)
    @resource_scope ||= scope_for(resource_class, **kwargs)
  end
```

ekohl:
Thanks! That's what I get for doing this in between meetings.

nofaralfasi:
Thanks @ekohl!

PR #705, opened 2024-03-29
Use theforeman-rubocop gem

archanaserver:
Inherited strictest.yml here, because this closely matches with what rules we already have in our .rubocop.yml config. Also dropped Rails, Style/HashSyntax, Style/Documentation and Style/SymbolArray from here because that was included in the inherited file.

This is part of RuboCop standardization, link for the reference:
https://community.theforeman.org/t/standardizing-rubocop-with-theforeman-rubocop/37239
https://github.com/theforeman/theforeman-rubocop/?tab=readme-ov-file#all-opinionated-cops-with-new-ones---strictest


.rubocop.yml
```diff
@@ -1,23 +1,18 @@
...
+  theforeman-rubocop:
+    - strictest.yml
 
 AllCops:
-  TargetRubyVersion: 2.7
-  TargetRailsVersion: 5.2
   Exclude:
-    - 'extras/**/*'
-    - 'locale/**/*'
-    - 'node_modules/**/*'
```
ekohl, line 15: Similar to the other place: you need to keep these locations. It's causing the current build issue.

ekohl (changes requested):

archanaserver:
The failure is due to the Ruby v2.2 but I'm not sure from where it is coming, can you help me here @ekohl?

.rubocop.yml
```diff
@@ -1,23 +1,21 @@
...
+    - strictest.yml
 
 AllCops:
   TargetRubyVersion: 2.7
   TargetRailsVersion: 5.2
   Exclude:
-    - 'extras/**/*'
-    - 'locale/**/*'
-    - 'node_modules/**/*'
-    - 'vendor/**/*'
```
evgeni, line 16: don't delete that line please, otherwise rubocop tries to scan all gems we install as dependencies which takes a lot of time and produces errors

.rubocop.yml
```diff
@@ -1,10 +1,9 @@
 inherit_from:
   - .rubocop_todo.yml
 
-require:
-  - rubocop-performance
-  - rubocop-rails
-  - rubocop-minitest
+inherit_gem:
+  theforeman-rubocop:
+    - strictest.yml
```
evgeni, line 6: Should we really use `strictest` given the amount of todo it generated?

archanaserver: @evgeni not sure why i used that earlier, but now with default we have less cops in todo as compared to what we have earlier.

archanaserver:
@evgeni i have made some changes, can you take a look?

evgeni:
@archanaserver looks reasonable to me

@nofaralfasi any last words on this? or shall I just hit he button?

PR #706, opened 2024-04-04
Fixes #37317 - Drop non-batch triggering

adamruzicka:


adamruzicka:
Turns out this is still occasionally useful, closing.

PR #712, opened 2024-04-17
SAT-22989 - Remove  'Run Insights maintenance plan'

stejskalleos:
> The Red Hat Access Insights plugin was deprecated in Satellite 6.9. It has been removed in this version of Satellite.

* https://access.redhat.com/documentation/en-us/red_hat_satellite/6.9/html-single/release_notes/index#deprecated_functionality
* https://access.redhat.com/documentation/en-us/red_hat_satellite/6.10/html-single/release_notes/index#deprecated_functionality

There is no need for changes in the [supported templates for Satellite](https://github.com/RedHatSatellite/foreman_theme_satellite/blob/develop/app/controllers/concerns/provisioning_templates_controller_branding.rb#L4)


jameerpathan111:
@stejskalleos what about all the other code related to it? won't that need to be removed too?
- https://github.com/search?q=repo%3Atheforeman%2Fforeman_ansible+Insights+maintenance+plan&type=code 
- https://github.com/theforeman/foreman_ansible/blob/master/app/services/foreman_ansible/insights_plan_runner.rb#L6
- https://github.com/search?q=repo%3Atheforeman%2Fforeman_ansible%20InsightsPlanRunner&type=code

stejskalleos:
I removed additional files, but I wonder if I'm not breaking anything.
@ShimShtein, can you please take a look? You know more about the Insights than I do.

@jameerpathan111 Thanks for pointing this out, I just want to let you know that the files in `locale/*` will be updated with the next translation sync.

jameerpathan111 (approved):

ShimShtein (approved):
LGTM, I don't see any problematic code here.

PR #717, opened 2024-05-08
Fixes #37418 - Fixes an issue that caused hidden Ansible variables to be shown in plain text on the Host-Details page

Thorben-D:
[Redmine Issue #37418 and reproducer](https://projects.theforeman.org/issues/37418)

Variables marked as __hidden__ were shown in plain text under **Variables** and **Inventory** on a host's details page.
This PR fixes that by masking the values in question in the UI.
Values are still shown in plain text when editing, as this requires the same permissions, _edit_ansible_variables_, as 

**Configure** > **Ansible** > **Variables**.

It should be noted, that _hidden_ variables are **NOT** considered secrets. The point of _hidden_ is to only hide the values of the respective variables in the UI. The [Foreman documentation](https://docs.theforeman.org/nightly/Managing_Configurations_Ansible/index-katello.html#Overriding_Ansible_Variables_in_foreman_ansible) clearly reflects this fact under point 6.

Changes:
- Add "hiddenValue" to GraphQL query hostVariableOverrides.gql
- Replace plain text secret with masked value
- Add a parameter "redact_secrets" to AnsibleInventoriesController#show_inventory
- Change frontend code to use newly added "redact_secrets" parameter
- Add a new "to_hash_with_secrets_redacted" method to InventoryCreator

Requires #716 

image
image


nofaralfasi:
I'm getting the following error when navigating to the `Variables` tab:
```
RuntimeError (Failed to implement AnsibleVariable.hiddenValue, tried:
- `Types::OverridenAnsibleVariable#hidden_value`, which did not exist
- `Presenters::OverridenAnsibleVariablePresenter#hidden_value`, which did not exist
- Looking up hash key `:hidden_value` or `"hidden_value"` on `#<Presenters::OverridenAnsibleVariablePresenter:0x00007f28c28ced48>`, but it wasn't a Hash
To implement this field, define one of the methods above (and check for typos)
```

Thorben-D:
@nofaralfasi I think that is because you still have the broken GQL scheme... Did you make sure the content of #716 is present on your branch?

nofaralfasi:
> @nofaralfasi I think that is because you still have the broken GQL scheme... Did you make sure the content of #716 is present on your branch?

You are right, I missed that part.
Now the `Variables` tab is shown properly. However, the default value of the hidden variable is still visible through the GraphQL response. 
It's a little more complicated to find it, but still very possible. 

Also, it's not possible to edit the variable value from the `Variables` tab. I see it's not related to the changes here, but we need to take care of it as well.

Thorben-D:
Glad you got it sorted.
Yes, the value may indeed still be extracted from the query response. I suppose we could check if the user has the `edit_ansible_variables` permission and require it for the value to be queried.

I tried to reproduce the issue you faced with editing the value, but without success. 
The default value is not changed by this action, but the value for this particular host is. For me, this is then shown under:
**Configure** > **Ansible** > **Variables** > <variable> > **Specify Matchers**

nofaralfasi:
> Glad you got it sorted. Yes, the value may indeed still be extracted from the query response. I suppose we could check if the user has the `edit_ansible_variables` permission and require it for the value to be queried.

Exactly. That should be the correct implementation. 


> I tried to reproduce the issue you faced with editing the value, but without success. The default value is not changed by this action, but the value for this particular host is. For me, this is then shown under: **Configure** > **Ansible** > **Variables** > _Variable_ > **Specify Matchers**

I apologize for the confusion, it was a problem on my setup. I'll be more careful next time.

Thorben-D:
Great, I'll implement that then!
No worries about the mix-up, thanks for having a look!

app/graphql/presenters/overriden_ansible_variable_presenter.rb
```diff
@@ -2,18 +2,32 @@ module Presenters
...
     def initialize(ansible_variable, override_resolver)
       @ansible_variable = ansible_variable
       @override_resolver = override_resolver
     end
 
     def current_value
-      @override_resolver.resolve @ansible_variable
+      resolved = @override_resolver.resolve @ansible_variable
+      resolved[:value] = '*****' unless ansible_variable.editable_by_user?
+      resolved
```
nofaralfasi, line 30: We need to check if the resolved value is not nil, otherwise we will get an error when trying to set the value.

app/services/foreman_ansible/inventory_creator.rb
```diff
@@ -19,28 +19,32 @@ def initialize(hosts, template_invocation = nil)
     # more advanced cases). Therefore we have only the 'all' group
     # with all hosts.
     def to_hash
+      to_hash_with_secrets_redacted(false)
+    end
+
+    def to_hash_with_secrets_redacted(redact_secrets = true)
```
nofaralfasi, line 25: Why do we always hide the value here? What if the user has permission to see the hidden values?

sbernhard: Which permission would this be? I don't think such a permission exists.

nofaralfasi: admin user/any user with `edit_ansible_variables` permission.

sbernhard: Ah. Thanks. I thought there is a permission whether to see hidden values or not

Thorben-D: I am hiding the value by default because like this, a conscious decision to show the values is required by the user. 
That's also how it works in **Configure**>**Ansible**>**Variables**>**<variable>**.

nofaralfasi: Correct. However, in our case, we are changing existing behavior. For example, if a user is using the API to read the hidden variable values, with our new changes, they won't be able to do that anymore, even with the right permissions.

Thorben-D: You are right. The hidden values were automatically hidden in the inventory by default. This is now fixed. 
As before, Inventories list hidden values by default. One may opt to include `redact_secrets=true` in the request to redact the secret values.
This is still done in the UI, as this, IMHO, is in line with the idea behind hidden values. Either way, a user needs the `view_hosts` permission to view a host's inventory, redacted or not.

nofaralfasi (changes requested):
Currently, any user with `Viewer` permissions can see the hidden values of the Ansible variables. 
The correct behavior should be as follows: if a user is not an admin and does not have the `edit_ansible_variables` permission, they should not be able to see the hidden values of the Ansible variables.

Thorben-D:
Indeed, I only hid the values for the overrides. Obviously the same needed to be done for the default value.
I also added a requirement for the "edit_ansible_variables" permission to view the value of hidden Ansible variables via the API and the Ansible inventory.
I am **very** unsure whether this is a good idea though, so I added those as a second commit for easy dropping. Also, I am sure that the filtering for the API call can be done much cleaner with ActiveSupport::ParameterFilter, I am just not sure how...

nixfu:
FYI, I think this same issue is affecting the ansible inventory report in Foreman that is used for ansible/AWX to use Foreman as an inventory source.    Previously any hidden params in Foreman, would show up as **** in the hostvars of the inventory json.  Now they are being shown as their full value and no longer hidden.

nofaralfasi:
> FYI, I think this same issue is affecting the ansible inventory report in Foreman that is used for ansible/AWX to use Foreman as an inventory source. Previously any hidden params in Foreman, would show up as **** in the hostvars of the inventory json. Now they are being shown as their full value and no longer hidden.

I've also noticed this behavior. However, could you clarify how it's related to the changes in the PR? The results haven't changed with this PR. It seems we may need to open a separate case to report this issue, unless I'm missing something.

app/services/foreman_ansible/ansible_info.rb
```diff
@@ -1,16 +1,18 @@
 module ForemanAnsible
   class AnsibleInfo < ::HostInfo::Provider
-    def host_info
-      { 'parameters' => ansible_params }
+    def host_info(redact_secrets = false)
+      { 'parameters' => ansible_params(redact_secrets) }
     end
 
-    def ansible_params
+    def ansible_params(redact_secrets = false)
```
nofaralfasi, line 7: If the value of `redact_secrets` is based on the user's permissions, why do we need to pass `redact_secrets` to this method? Can't we check if the user has the necessary permissions directly within the method instead?

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
```diff
@@ -22,6 +23,21 @@ export const formatValue = variable => {
     ? variable.currentValue.value
     : variable.defaultValue;
 
+  if (variable.hiddenValue) {
```
nofaralfasi, line 26: Is the `hiddenValue` necessary here, or could we rely on `variable.meta.canEdit` instead?

nofaralfasi (changes requested):
To summarize:
1. I believe we don’t need the new `to_hash_with_secrets_redacted` method or the `redact_secrets` parameter. Checking the user’s permissions directly on `ForemanAnsible::AnsibleInfo#ansible_params` should be sufficient.
2. We should simplify the GQL query by removing unnecessary fields, such as `override` and `hidden_value`.
3. Most of the functionality already exists; it's better to utilize it where possible. For example, use `variable.meta.canEdit` instead of `variable.hiddenValue`.

app/graphql/presenters/overriden_ansible_variable_presenter.rb
```diff
@@ -2,18 +2,33 @@ module Presenters
...
+    def default_value
+      ansible_variable.editable_by_user? ? ansible_variable.default_value : '*****'
+    end
+
     def current_value
-      @override_resolver.resolve @ansible_variable
+      resolved = @override_resolver.resolve @ansible_variable
+      unless resolved.nil?
+        resolved[:value] = '*****' unless ansible_variable.editable_by_user?
+      end
```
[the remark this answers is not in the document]
Thorben-D: Done

app/graphql/presenters/overriden_ansible_variable_presenter.rb
```diff
@@ -2,18 +2,33 @@ module Presenters
...
+      ansible_variable.override?
+    end
 
     def initialize(ansible_variable, override_resolver)
       @ansible_variable = ansible_variable
       @override_resolver = override_resolver
     end
 
+    def default_value
+      ansible_variable.editable_by_user? ? ansible_variable.default_value : '*****'
```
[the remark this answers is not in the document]
Thorben-D: Good idea, done

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
```diff
@@ -9,7 +9,10 @@ import AnsibleHostInventory from './AnsibleHostInventory';
 import ErrorState from '../../../ErrorState';
 
 const WrappedAnsibleHostInventory = ({ hostId }) => {
-  const params = useMemo(() => ({ params: { host_ids: [hostId] } }), [hostId]);
+  const params = useMemo(
+    () => ({ params: { host_ids: [hostId], redact_secrets: true } }),
```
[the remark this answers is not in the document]
Thorben-D: For the inventory, I am unsure what's best. As opposed to the variables, there is no action to toggle the visibility of hidden variables. I chose to hide hidden variables in the inventory for everyone, as permitted users are able to view those values under the "Variables" tab. I also feel like showing the values by default sort of defeats the purpose of hiding them. 
Though, I can change this if you disagree.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
Thorben-D: Unfortunately, the `hiddenValue` field is needed to decide whether the "fake" variable should be shown instead of the real one. Relying on `canEdit` here would not work, as a permitted user can edit hidden and "not-hidden" variables.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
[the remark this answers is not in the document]
nofaralfasi: As I mentioned, I prefer not to change the current behavior where authorized users can view hidden values in the inventory. Consider a scenario where someone is using Foreman with a script that makes API calls to retrieve the inventory and relies on those variable values. If we change this behavior, it could break their setup. However, if you believe your approach is better, feel free to initiate a discussion in the community to gather more opinions.

webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
nofaralfasi: On second thought, this block isn't necessary. If the user is authorized to view hidden values, we can simply display them as is. If they aren’t authorized, the value should already be hidden by the backend before it's sent.

webpack/components/AnsibleHostDetail/components/AnsibleHostInventory/index.js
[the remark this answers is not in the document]
Thorben-D: I should note that the variables are only hidden if `redact_secrets=true` is passed during the API call. That is also the reason I have to pass this value through a couple of functions. This only happens in the UI.

nofaralfasi:
> Is this going to be fixed? Its a pretty big security issue and makes hidden params useless in Foreman.

The PR is still under review and requires some changes before it can move forward.

PR #722, opened 2024-06-20
Add a template that enables download and execution of a script

ShimShtein:
The main idea is to execute scripts from the Foreman server, similar to theforeman/foreman#10208.

### DIFF
--- a/app/views/foreman_ansible/job_templates/download_and_execute_script.erb
+++ b/app/views/foreman_ansible/job_templates/download_and_execute_script.erb
@@ -1,0 +1,28 @@
+<%#
+name: Download and execute a script
+job_category: Ansible Commands
+description_format: Download script from %{url} and execute it
+snippet: false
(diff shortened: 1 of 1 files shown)


app/views/foreman_ansible/job_templates/download_and_execute_script.erb
```diff
@@ -0,0 +1,24 @@
...
+  advanced: false
+provider_type: Ansible
+kind: job_template
+model: JobTemplate
+%>
+
+---
+- hosts: all
+  tasks:
+    - shell:
```
ekohl, line 20: Would it be better to use the [url](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/url_lookup.html) module to retrieve the content and then execute the retrieved content?

ShimShtein:
@ekohl Now I am using uri module to download the script. How does this one look?

PR #723, opened 2024-06-30
Fixes #37612 - Fix deletion of Ansible variables from Host page

nofaralfasi:
- Use `isEqual` for deep comparison when matching variable values,
  ensuring correct behavior during variable deletion.
- Add `defaultValue` field to GraphQL mutation, for proper value display
  after Ansible variable deletion.

### DIFF
--- a/webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
+++ b/webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { isEqual } from 'lodash';
 import { TimesIcon, CheckIcon } from '@patternfly/react-icons';
 import { sprintf, translate as __ } from 'foremanReact/common/I18n';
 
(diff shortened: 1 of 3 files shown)


Thorben-D:
Looks like a sensible change but I can't test it unfortunately, as I am not sure how to reproduce the issue.

nofaralfasi:
> Looks like a sensible change but I can't test it unfortunately, as I am not sure how to reproduce the issue.

I added the steps to reproduce this bug - https://projects.theforeman.org/issues/37612#note-2

Thorben-D:
Thank you for adding the reproducer!
I tested it and can confirm that the deletion of overrides is now working.
I did however notice a different issue:
`TypeError: Cannot read properties of null (reading 'value')
    at AnsibleVariableOverridesTableHelper.js:52:59`
    
This happens because the [OverridenAnsibleVariablePresenter](https://github.com/theforeman/foreman_ansible/blob/88ed8568c56d6ab989766c1520ffde5f398f49df/app/graphql/presenters/overriden_ansible_variable_presenter.rb#L16) returns `nil` for the **current_value** field, as the override has just been deleted. 
This then breaks [this](https://github.com/theforeman/foreman_ansible/blob/88ed8568c56d6ab989766c1520ffde5f398f49df/webpack/components/AnsibleHostDetail/components/AnsibleVariableOverrides/AnsibleVariableOverridesTableHelper.js#L51C4-L51C66) function, which is used to set the text of the "Value" and "Edit" field. Right now, the edit field still retains the override-value after deletion.

stejskalleos:
@MariaAga review please

MariaAga (approved):
lgtm

stejskalleos:
Thanks @nofaralfasi @Thorben-D @MariaAga

PR #727, opened 2024-08-06
Update capsule upgrade

ehelms:
The foreman-maintain upgrade command has changed to remove target-version and we have introduced a dedicated update command for .Z updates. This updates the upgrade playbook and adds a dedicated update playbook to reflect this split in foreman-maintain.

### DIFF
--- a/app/views/foreman_ansible/job_templates/capsule_update_-_ansible_default.erb
+++ b/app/views/foreman_ansible/job_templates/capsule_update_-_ansible_default.erb
@@ -1,0 +1,79 @@
+<%#
+name: Capsule Update Playbook
+snippet: false
+template_inputs:
+- name: target_version
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  hidden_value: false
+- name: whitelist_options
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  hidden_value: false
+model: JobTemplate
+job_category: Maintenance Operations
+description_format: "%{template_name}"
+provider_type: Ansible
+kind: job_template
+feature: ansible_run_capsule_update
+%>
+
+---
+- hosts: all
+  tasks:
+<% if plugin_present?('foreman_theme_satellite') -%>
+    - name: Gather the rpm package facts
+      package_facts:
+        manager: auto
+
+    - name: Fail if the target server is a Satellite server
+      fail:
+        msg: "This playbook cannot be executed on a Satellite server. Use only on a Capsule server."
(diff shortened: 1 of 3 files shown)


pondrejk (approved):
ack from the functional side (based on the prt pass within https://github.com/SatelliteQE/robottelo/pull/15863)

app/views/foreman_ansible/job_templates/capsule_upgrade_-_ansible_default.erb
```diff
@@ -47,7 +45,7 @@ feature: ansible_run_capsule_upgrade
                                "--whitelist=#{input('whitelist_options')}"
                              end -%>
       - name: Upgrade Capsule server using satellite-maintain
-        shell: satellite-maintain upgrade run --assumeyes --target-version={{ target_version }} <%= whitelist_option %>
+        shell: satellite-maintain upgrade run --assumeyes <%= whitelist_option %>
```
adamruzicka, line 48: What does it do now? Does it go up a version or all the way to latest available?

adamruzicka (commented):
Looks reasonable

app/views/foreman_ansible/job_templates/capsule_update_-_ansible_default.erb
```diff
@@ -0,0 +1,79 @@
+<%#
+name: Capsule Update Playbook
+snippet: false
+template_inputs:
+- name: target_version
+  required: false
+  input_type: user
+  advanced: false
+  value_type: plain
+  hidden_value: false
```
adamruzicka, line 10: Where is this input used?

adamruzicka (commented):
One last question

app/views/foreman_ansible/job_templates/capsule_update_-_ansible_default.erb
ehelms: I don't understand the question. I just copy/pasted the upgrade playbook and updated it to the best of my knowledge.

adamruzicka: The old template took `target_version` as an input and then passed it as a parameter to the installer. Now this input was dropped from the upgrade template (and its use from the corresponding installer invocation), but it is still kept here without, as far as I can see, being used anywhere. Either it should be passed to the installer or dropped from here.

ehelms: :facepalm: Now I get it! I have deleted it.

adamruzicka (approved):
LGTM

PR #729, opened 2024-08-14
Fixes #37733 - Add audits of ansible role changes on hosts

adamlazik1:
This records audits when ansible role is added to a host or removed from a host.

Defining the `to_label` method is necessary so the name of host ansible
role is shown in the format `role / host` rather than `id / host`.

### DIFF
--- a/test/unit/host_ansible_role_test.rb
+++ b/test/unit/host_ansible_role_test.rb
@@ -17,4 +17,34 @@ class HostAnsibleRoleTest < ActiveSupport::TestCase
     end
     should validate_uniqueness_of(:ansible_role_id).scoped_to(:host_id)
   end
+
+  describe 'auditing' do
+    setup do
+      @host = FactoryBot.create(:host)
+      @ansible_role = FactoryBot.create(:ansible_role)
+      @host_ansible_role = FactoryBot.create(:host_ansible_role, :with_auditing, host_id: @host.id, ansible_role_id: @ansible_role.id)
+      @audits = @host_ansible_role.audits
+    end
+
+    test 'should audit creation of a host ansible role' do
+      assert_equal 1, @audits.size
+      assert_associated_ids
+      assert_equal 'create', @audits.last.action
+    end
+
+    test 'should audit deletion of a host ansible role' do
+      @audits.clear
+      @host_ansible_role.destroy
+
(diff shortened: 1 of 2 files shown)


adamlazik1:
The look of the audits

image

nofaralfasi:
@adamlazik1 Could you add tests to this PR to help validate the changes?

adamlazik1:
Added the test, I hope it's in the correct place.

test/unit/host_ansible_role_test.rb
```diff
@@ -17,4 +17,25 @@ class HostAnsibleRoleTest < ActiveSupport::TestCase
     end
     should validate_uniqueness_of(:ansible_role_id).scoped_to(:host_id)
   end
+
+  describe 'auditing' do
+    test 'audit creation and deletion of host ansible roles' do
```
nofaralfasi, line 22: I recommend splitting this into separate tests for each action. One test should focus on verifying the audit record for creation, and another for deletion.

test/unit/host_ansible_role_test.rb
```diff
@@ -17,4 +17,25 @@ class HostAnsibleRoleTest < ActiveSupport::TestCase
...
+      host = FactoryBot.create(:host)
+      ansible_role = FactoryBot.create(:ansible_role)
+      host_ansible_role = FactoryBot.create(:host_ansible_role, :with_auditing, host_id: host.id, ansible_role_id: ansible_role.id)
+
+      assert_not host_ansible_role.audits.blank?
```
nofaralfasi, line 27: IMO, it would be clearer to ensure that each test verifies only one audit record is created. This helps confirm that the auditing is working as expected.

nofaralfasi (changes requested):

adamlazik1:
Alright, I split the test and tried to extract some common parts.

nofaralfasi (changes requested):
Could you please update the `assert_equal` statements so that the expected value is the first argument, in line with common [practice](https://ruby-doc.org/stdlib-3.0.2/libdoc/test-unit/rdoc/Test/Unit/Assertions.html#method-i-assert_equal)?

adamlazik1:
Done.

nofaralfasi (approved):

PR #731, opened 2024-08-19
i18n - pulling from tx

stejskalleos:


locale/foreman_ansible.pot
```diff
@@ -895,37 +877,37 @@ msgid "Runs an Ansible playbook which contains all the roles defined for a host"
...
+#: ../lib/foreman_ansible/remote_execution.rb:40
 msgid "Upgrade Capsules on given Capsule server hosts"
 msgstr ""
 
+#: ../lib/foreman_ansible/remote_execution.rb:45
+msgid "Update Capsules on given hosts"
+msgstr ""
+
+#: ../lib/foreman_ansible/remote_execution.rb:46
+msgid "Update Capsules on given Capsule server hosts"
```
ekohl, line 908: I think it's really poor practice to include downstream terminology into upstream, but it was already there so not blocking for this PR.

@adamruzicka I see you added it in c1ebb2c78233302573abd2c2cc79da0bbafa2063 but should this have been added to foreman_theme_satellite?

locale/ja/foreman_ansible.po
ekohl, line 1: This whole file is messed up in terms of encoding. Looks like we missed that in 3dace0ad2cb3fe0c6812ff1d7b30dd49d036257d.

Right now it's needed to update the `.pot` file, but something we absolutely need to fix before the next release.

ekohl:
Test failures are probably unrelated

locale/foreman_ansible.pot
adamruzicka: It's not only terminology, the whole thing is downstream specific without any sensible upstream equivalent.

While it would make sense putting downstream things into the downstream plugin, including random pieces of functionality in a branding plugin doesn't feel exactly great either.

PR #734, opened 2024-09-05
i18n - pulling from tx

ekohl:
This primarily solves multiple encoding issues in the Japanese translation. It also removes the empty English "translation".


nofaralfasi:
Failing tests seem related.

PR #738, opened 2024-09-11
Bump version to 15.0.0

nofaralfasi:


evgeni:
```
Gem loading error: Could not find 'foreman_remote_execution' (>= 14.0, < 15) - did find: [foreman_remote_execution-13.2.5]
```
I guess REX 14 is not packaged yet.

PR #739, opened 2024-09-20
Fixes #37848 - Support Rails 7.0

ofedoren:


config/routes.rb
```diff
@@ -7,30 +7,28 @@
           :defaults => { apiv: 'v2' },
           :apiv => /v2/,
           :constraints => ApiConstraints.new(version: 2, default: true) do
-      constraints(:id => %r{[^\/]+}) do
```
ekohl, line 10: Was this never needed? I can't find anything in the Rails upgrade guide relating to this.

nofaralfasi:
> @nofaralfasi I recall we talked about the CI failures of foreman_ansible and my impression is that this is the same failure we see in master. Have you had any chance to look closer at it?

Yes, but unfortunately I couldn't figure out how to solve it.

ofedoren:
Dropped commit for testing with Foreman PR.

ekohl (approved):

stejskalleos (approved):

PR #743, opened 2024-11-21
Fixes #38031 - Use the same routes constraints as Foreman

ofedoren:
#### Overview of Changes
- Somewhat reverts and updates https://github.com/theforeman/foreman_ansible/commit/73bb74a0d2f2c06048a5a0f86cfe9091c81a4a76.

- Without this change `GET /api/v2/hosts/<fqdn_or_id>/power` is not recognized in API calls, even though it's listed internally (e.g. rails routes command) on Rails 7.0.

#### Testing Steps
- I've tried to write a test, but locally it didn't discover this issue (test passed), I've tested it by running the app with/out this patch checked on hosts index page the power state icons (they should not be grey for real machines) or just check in the logs for failed requests. 

#### Checklists

* [?] I have added relevant tests for my changes.
* [?] I have updated the documentation accordingly.

#### Additional Notes
- Doesn't depend on any other PR, but to observe the issue, check with https://github.com/theforeman/foreman/pull/10299.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,14 +7,14 @@ Rails.application.routes.draw do
           :defaults => { apiv: 'v2' },
           :apiv => /v2/,
           :constraints => ApiConstraints.new(version: 2, default: true) do
-      constraints(:id => %r{[^\/]+}) do
-        resources :hosts, :except => [:new, :edit] do
+      constraints(:id => /[^\/]+/) do
+        resources :hosts, :only => [] do
           member do
             post :play_roles
             get :ansible_roles
             post :assign_ansible_roles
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -7,14 +7,14 @@
           :defaults => { apiv: 'v2' },
           :apiv => /v2/,
           :constraints => ApiConstraints.new(version: 2, default: true) do
-      constraints(:id => %r{[^\/]+}) do
```
ekohl, line 10: I'm fine merging this because it exactly matches what's in Foreman, but I wonder if `%r{[^/]+}` would also work?

ofedoren: It should, since it's still accurate than the previous one, but I'd rather sync between core and plugins.

ekohl (approved):

nofaralfasi:
Thanks @ofedoren!

PR #744, opened 2024-11-26
Fixes #38045 - Fix differing hidden_value variable name

dosas:
#### Overview of Changes
the value for hidden_value should be the same for create and query

#### Testing Steps
Component tests within this repository

#### Checklists

#### Additional Notes
https://github.com/SatelliteQE/nailgun/pull/1220#discussion_r1840863525

### DIFF
--- a/app/views/api/v2/ansible_variables/show.json.rabl
+++ b/app/views/api/v2/ansible_variables/show.json.rabl
@@ -2,7 +2,7 @@ object @ansible_variable
 
 attribute :parameter
 attributes :id, :variable, :ansible_role, :ansible_role_id, :description, :override,
-           :variable_type, :hidden_value?, :validator_type,
+           :variable_type, :validator_type,
            :validator_rule, :merge_overrides, :merge_default,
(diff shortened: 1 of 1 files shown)


Thorben-D (approved):

app/views/api/v2/ansible_variables/show.json.rabl
```diff
@@ -19,3 +19,5 @@ end
 node :override_values_count do |lk|
   lk.lookup_values.count
 end
+
+node :hidden_value, &:hidden_value?
```
nofaralfasi, line 23: As I mentioned [here](https://github.com/theforeman/foreman_ansible/pull/716#issuecomment-2314902463), the `hidden_value` does not appear to be used in this query. Could you clarify how this change impacts the functionality? I might be missing the reason why it’s necessary.

nofaralfasi: After looking into this further, I noticed that this change only affects the API call and does not impact the GraphQL query.

PR #753, opened 2025-02-11
Use trusted publishers to release the gem

ekohl:
#### Overview of Changes

To convert from the legacy API-key based publishing to using Trusted Publishers. See https://community.theforeman.org/t/use-trusted-publishers-to-release-gems/42009 for the details.

See the RFC. Some admin (likely me) will also need to set up GitHub and Rubygems.

#### Testing Steps

Push a tag and see that it gets published.

#### Checklists

* [ ] I have added relevant tests for my changes.
* [x] I have updated the documentation accordingly.

#### Additional Notes

### DIFF
--- a/.github/workflows/release_gem.yml
+++ b/.github/workflows/release_gem.yml
@@ -1,6 +1,6 @@
 name: Release
 
-on:
+on:  # yamllint disable-line rule:truthy
   push:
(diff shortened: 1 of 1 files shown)


nofaralfasi:
Thanks @ekohl!

PR #754, opened 2025-02-19
i18n - pulling from tx

ekohl:


ekohl:
76a53e3563df223e82076bc917a42836d9387819 added a new string and this updates the .pot file for that. For 100% translation it will need another round.

stejskalleos (approved):

PR #758, opened 2025-03-05
update acts_as_list requirement

evgeni:
#### Overview of Changes

Update acts_as_list to (officially) support latest Ruby and Rails

- 1.0.4 had official support for Ruby 3 (but only rails 6.1)
- 1.1.0 added support for Rails 7.0
- 1.2.0 added support for Rails 7.1

#### Testing Steps
- What steps should reviewers follow to test this PR? 

#### Checklists

* [ ] I have added relevant tests for my changes.
* [ ] I have updated the documentation accordingly.

#### Additional Notes
- Please include any additional information that might be helpful for reviewers.
- If this PR is dependent on any other PRs, please list them here.

### DIFF
--- a/foreman_ansible.gemspec
+++ b/foreman_ansible.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
             ['package.json']
   s.test_files = Dir['test/**/*']
 
-  s.add_dependency 'acts_as_list', '~> 1.0.3'
+  s.add_dependency 'acts_as_list', '~> 1.2'
   s.add_dependency 'deface', '< 2.0'
   s.add_dependency 'foreman_remote_execution', '>= 14.0', '< 16'
   s.add_dependency 'foreman-tasks', '>= 10.0', '< 11'


foreman_ansible.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
             ['package.json']
   s.test_files = Dir['test/**/*']
 
-  s.add_dependency 'acts_as_list', '~> 1.0.3'
+  s.add_dependency 'acts_as_list', '~> 1.2'
```
evgeni, line 21: Technically we could also do `~> 1.0, >= 1.0.3` if you prefer.

nofaralfasi (approved):

foreman_ansible.gemspec
nofaralfasi: Do you see any valid reason for using the wider range?

evgeni: The only I can think of is packaging.
Today we package 1.0.3, so setting 1.2 here would mean we need immediately also update packaging. (Which we can do! Just spelling it out loud here)

nofaralfasi: Thanks Evgeni, so I'm good with merging it as is.

PR #772, opened 2025-09-25
Add breaks-robottelo workflow as per theforeman/foreman PR 10568

lhellebr:
Before merging, `breaks-robottelo` label needs to be added to the repo manually by a maintainer


ofedoren (approved):
Thanks, @lhellebr, I've added the label.

PR #773, opened 2025-10-01
Rename release workflow file

adamruzicka:
to adhere to the release process RFC


adamruzicka:
We are, I'll deal with rubygems once this gets in

PR #774, opened 2025-10-02
Test dynflow fix

ofedoren:
Just to run few times.


adamruzicka:
Fix for the redirect is https://github.com/theforeman/foreman_ansible/pull/773/commits/c1c581983bd4e962ad0859674cb8859f0f48c102

ofedoren:
Should we rather open a PR with that? :)

I'll just include it here for the time being.

PR #775, opened 2025-10-07
i18n - pulling from tx

adamlazik1:


ofedoren (commented):
This is weird... Wasn't there at least 1 new string to translate?.. The changes are quite empty :/

adamlazik1:
@ofedoren I wasn't able to get anything more out of transifex. Would you mind testing on your environment whether you'll achieve the same result?

ofedoren (approved):
Kinda weird, but I've got the same results, nevermind then :) Thanks, @adamlazik1 !

PR #786, opened 2026-03-04
Fixes #39129 - Reload host's roles after role assignment

adamruzicka:
This was previously done implicitly by apollo (as of 3.3.7), it seems to be no longer happening with apollo 3.14.0.


adamruzicka:
Robottelo approves https://github.com/SatelliteQE/robottelo/pull/20937#issuecomment-3997763328

pondrejk (approved):

adamlazik1 (approved):

adamlazik1:
Thanks @adamruzicka @pondrejk !

ofedoren:
Just for my sanity and future reference on why we changed that. It seems the default refetching was disabled in 3.4.0: https://github.com/apollographql/apollo-client/compare/v3.3.0...v3.4.0#diff-5fb8ad16cfbcb51bc035d32b5963734561c440b48417a1c2dbf80a16098be67bR460-R461

And it seems the other `useMutation`s don't need such change, the cache seems to be updated correctly.

PR #788, opened 2026-03-28
Fixes #29583 - Hide taxonomy options from Ansible API docs

jakduch:
AnsibleRole and AnsibleVariable models do not support taxonomies,
but the API documentation inherited organization_id and location_id
parameters from BaseController. Passing these parameters resulted
in a 500 error (Association not found for organization).

Add `hide_taxonomy_options` to both `AnsibleRolesController` and
`AnsibleVariablesController` to remove these parameters from the
API documentation and silently drop them from requests.

### DIFF
--- a/app/controllers/api/v2/ansible_roles_controller.rb
+++ b/app/controllers/api/v2/ansible_roles_controller.rb
@@ -6,6 +6,7 @@ module Api
     class AnsibleRolesController < ::Api::V2::BaseController
       include ::Api::Version2
       include ::ForemanAnsible::AnsibleRolesDataPreparations
+      hide_taxonomy_options
 
       resource_description do
         api_version 'v2'
--- a/app/controllers/api/v2/ansible_variables_controller.rb
+++ b/app/controllers/api/v2/ansible_variables_controller.rb
@@ -6,6 +6,7 @@ module Api
     class AnsibleVariablesController < ::Api::V2::BaseController
       include ::Api::Version2
       include Foreman::Controller::Parameters::AnsibleVariable
+      hide_taxonomy_options
 
       resource_description do
         api_version 'v2'


ofedoren (commented):
Thanks, @jakduch, but unfortunately it doesn't work.

The issue is that `hide_taxonomy_options` internally uses `resource_description` which gets overridden by `resource_description` below.

It will also won't work if we move `hide_taxonomy_options` below `resource_description` here, since it will override `api_base_url` :/

The solution which would work is something like https://github.com/Katello/katello/pull/11644/changes. But I'd like to not make that a pattern, I'd rather adjust `hide_taxonomy_options` in Foreman, so all the places get treated automatically and there is less clean-up in the future.

What about adjusting https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/base_controller.rb#L165 with

```ruby
      def self.hide_taxonomy_options
        prepend_before_action :drop_taxonomy_id_from_params

        original_method = method(:resource_description)
        define_singleton_method(:resource_description) do |options = {}, &block|
          original_method.call(options) do
            instance_exec(&block) if block
            param :location_id, Integer, :show => false
            param :organization_id, Integer, :show => false
          end
        end

        resource_description
      end
```

It'll require some testing and ensuring it won't mess up with current state of things, but I believe it will be a better solution.

jakduch:
Thanks for the review @ofedoren. I've implemented your suggestion as a separate core Foreman PR:

**Core fix:** [theforeman/foreman#10953](https://github.com/theforeman/foreman/pull/10953) - adjusts `hide_taxonomy_options` to wrap `resource_description` via `define_singleton_method`, so subsequent `resource_description` calls don't overwrite the hidden params. Includes tests for both scenarios (with and without subsequent `resource_description`).

This PR (foreman_ansible) remains unchanged - it just adds `hide_taxonomy_options` to both controllers, which will work correctly once the core fix is merged.

PR #789, opened 2026-04-27
Fixes #39260 - Move theforeman-rubocop dependency to Gemfile

ofedoren:
#### Overview of Changes

Move theforeman-rubocop dependency to Gemfile instead of specifying it in gemspec.

With newer bundler and the way we setup the Foreman with plugins, there are resolution failures due to duplicated dependencies. 

#### Testing Steps

Run `bundle install` with any other plugin that has the same dependency with a different version.

Before: fails to resolve (or there are warning on older bundler version)
After: passes without any issue, rubocop still works locally and in CI

#### Checklists

* [ ] ~I have added relevant tests for my changes.~
* [ ] ~I have updated the documentation accordingly.~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,5 @@
 source 'http://rubygems.org'
 
 gemspec
+
+gem 'theforeman-rubocop', '~> 0.1.2', require: false, groups: [:development, :rubocop]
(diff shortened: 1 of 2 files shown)


adamruzicka (approved):

adamruzicka:
Thank you @ofedoren !

PR #794, opened 2026-06-24
Drop foreman-tasks dependency upper bound

ogajduse:
The upper bound on foreman-tasks (currently `< 13`) causes repoclosure failures whenever foreman-tasks releases a new major version, requiring cascading unpins across plugins.

Since the Endeavour team controls both foreman_ansible and foreman-tasks, drop the upper bound entirely — other plugins (`foreman_remote_execution`, `katello`, `foreman_rh_cloud`) already use only a lower bound.

See: https://github.com/theforeman/foreman-packaging/pull/13661


adamruzicka (approved):

ofedoren:
Since we're at it, should we do the same for REX?

adamruzicka:
Thank you @ogajduse !

PR #803, opened 2026-08-19
Fixes #36359 - Run roles on nested host groups

jakduch:
## Summary

Fixes [#36359](https://projects.theforeman.org/issues/36359).

Running Ansible roles from a host group selected only hosts assigned directly to that group. Hosts in nested groups were omitted even though they inherit the parent's Ansible roles.

Add a host-group subtree relation and use it for both the HTML and API actions. Bulk runs combine the selected subtrees in one host query, so selecting both a parent and its child does not duplicate targets.

Add model and API regression coverage for nested groups and overlapping bulk selections.

## Testing

- Ruby syntax checks pass for all changed files
- Added model and controller regression tests; CI runs them in the supported Foreman test environment

## AI usage disclosure

Per the [community discussion on AI policy](https://community.theforeman.org/t/could-foreman-benefit-from-an-ai-usage-policy/44638), the issue was investigated and the changes, tests, and PR wording were prepared with the assistance of Codex 5.6 Sol High. The resulting changes were reviewed before submitting. The commit also includes an `Assisted-By` trailer.

### DIFF
--- a/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
+++ b/app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
@@ -23,7 +23,7 @@ module ForemanAnsible
 
           def play_roles
             find_resource
-            composer = job_composer(:ansible_run_host, @hostgroup.hosts)
+            composer = job_composer(:ansible_run_host, "parent_hostgroup = \"#{@hostgroup.title}\"")
             process_response composer.trigger!, composer.job_invocation
           end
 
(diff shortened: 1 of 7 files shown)


app/controllers/foreman_ansible/api/v2/hostgroups_controller_extensions.rb
```diff
@@ -23,7 +23,7 @@ module HostgroupsControllerExtensions
 
           def play_roles
             find_resource
-            composer = job_composer(:ansible_run_host, @hostgroup.hosts)
+            composer = job_composer(:ansible_run_host, @hostgroup.hosts_in_subtree)
```
adamruzicka: It should also be possible to pass a string as the second parameter, which will then be interpreted as a search query over hosts. Would this work?
```suggestion
            composer = job_composer(:ansible_run_host, "parent_hostgroup = \"#{@hostgroup.title}\"")
```

jakduch: Good point, this works and avoids materializing the host relation. Updated, thanks!

adamruzicka: Similar thing could be done in the other call sites, right?

jakduch: Legit, I’ll use the same parent_hostgroup search for the HTML action and a combined search for the bulk action, then remove the extra hosts_in_subtree relation.

